PR #8, opened 2017-01-11
change URLs in README to use https

hex2a:


snaow:
Hi @hex2a,

Thanks so much for your contribution.
You are right is way better to display all the links using HTTPS.

Let me take this opportunity to announce that Wazuh App will be hosted at Amazon S3 + CloudFront.
Using the URL https://packages.wazuh.com.

In the meantime, thanks again for your feedback. We will be here in case you need anything.

PR #28, opened 2017-09-05
add warning about how to install plugin

jsoref:


faust64:
just noticing this PR, as I broke my 3 kibanas post-install, earlier today...
I tried fixing some chmod/chown in plugins and optimized, yet I'am still fighting with broken display, ... no error in chrome console, 200 codes everywhere wiping cache, ... kibana logs look fine, even on restart. I've been looking at this without understanding wtf for a few hours ...
please merge!

Also consider updating docs (https://documentation.wazuh.com/current/installation-guide/installing-elastic-stack/elastic_server_deb.html#kibana). Suggesting to execute `apt-get install` without sudo would indicate we're using a root shell. The very next command installs plugin, while it is not made clear that one should be run by kibana.
Not being familiar with kibana plugins, I can confirm their doc does state plugins should be installed using some sudo (https://www.elastic.co/guide/en/kibana/current/_installing_plugins.html#_installing_plugins_with_linux_packages). I could have known, ... still, your doc is misleading.

havidarou:
Hi @jsoref @faust64, thanks for your input!

jsoref:
Yeah, as it happens, my kibana no longer has a wazuh app thing. I've tried to fix it but I've given up.

rossengeorgiev:
It would be simpler to just change the command use sudo:

```bash
$ sudo -u kibana /usr/share/kibana/bin/kibana-plugin install https://packages.wazuh.com/wazuhapp/wazuhapp-2.0_5.2.2.zip

jsoref:
@rossengeorgiev: I don't have a strong opinion on the matter.
The command appears a whole bunch of times on the page. `sudo` doesn't seem to be present in the repo. And I suppose it's possible for a system not to use `sudo` at all.

Personally, I generally do use `sudo`.

If a maintainer asks for `sudo` w/ instructions, I can certainly apply it. But such a change is a commitment to add more text w/ each additional row that needs to be added to the table...

havidarou:
We've updated our documentation accordingly to reflect the fact that the plugin installation command should be sudo.

PR #50, opened 2017-11-06
Fix call when filename is ar.conf

jesusgn90:


havidarou:
LGTM! (Waiting for API fix)

PR #52, opened 2017-11-07
Added "Unknown" platform to the agents tab. 

jesusgn90:
Now you can filter agents without os and having no conflict with the API call. The trick was check if is an "Unknown" case, then: 

```js
if(isUnknown){
       this.items = this.items.filter(item => typeof item.os === 'undefined');
}
```

### DIFF
--- a/public/controllers/agentsPreview.js
+++ b/public/controllers/agentsPreview.js
@@ -40,6 +40,11 @@ app.controller('agentsPreviewController', function ($scope, Notifier, genericReq
                     }
                 }
             }
+            $scope.osPlatforms.push({
+                name:     'Unknown',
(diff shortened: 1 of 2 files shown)


havidarou:
LGTM!

jesusgn90:
Ok now the "Unknown" osPlatform is only added if there is at least 1 agent without known OS. You can see bellow:

```js
            return apiReq.request('GET', '/agents/summary', { });
        })
        .then(data => {
            if(parseInt(data.data.data['Never connected']) > 0){
                $scope.osPlatforms.push({
                    name:     'Unknown',
                    platform: 'Unknown',
                    version:  'Unknown'
                });
            }
        });
```

PR #54, opened 2017-11-07
Agents and files sortable in groups tab

jesusgn90:
• Also now we can click in any field of the agent to go to the main agent view.
• Fix sorting by column in agents tab


jesusgn90:
Conflict is now resolved :smile:

havidarou:
LGTM!

PR #60, opened 2017-11-10
3.0-dev-ruleset-filters

jesusgn90:
Summary:
- Common menu items are now a directive
- Removed some console.log for better performance
- New function in DataHandler to get a filter value from a filter name
  - Ruleset template - rules uses getFilter function from DataHandler
- Fixed last agent in agents tab
- Fixed when-scroll directive in multiple tabs
- Fixed "Last registered agent" in Agents tab


havidarou:
LGTM!

PR #61, opened 2017-11-13
Removed unused appState $cookies

jesusgn90:
Summary:
- Removed unused methods from appState
- Fix the last agent information on manager - status tab
- In manager - status, don't show keep alive date in last registered agent if it is the manager


havidarou:
LGTM!

PR #79, opened 2017-11-23
First approach for CSS reordering

JuanjiJG:
This PR reorders all the CSS from the app, removing the hard-coded one from the templates, and making a first approach of reordering in the `less` folder.

The next step is to start removing and clearing the existing CSS in order to reduce the amount of repeated and unused rules.


havidarou:
LGTM!

PR #80, opened 2017-11-23
"Dashboard" tab deleted from app

JuanjiJG:
There was some old files of a "Dashboard" tab. They were deleted from the app and also every other remaining references and links to the Dashboard templates.


havidarou:
LGTM!

PR #81, opened 2017-11-23
Added cluster configuration to the manager configuration tab

jesusgn90:
In the "manager status" tab, the cluster daemon __will appear__ when it is available, in the sense that we make the call to the api `/manager/status` and then iterated on the array obtained.

The configuration of the cluster has been added to the "manager configuration" tab.


havidarou:
LGTM!

PR #86, opened 2017-11-24
Parallellized manager status

jesusgn90:
We can make use of Promise.all to handle parallel request at same time in order to optimize view load time.


jesusgn90:
If you don't ever seen Promise.all read bellow:

```js
foo()
.then(data => {
   $scope.a = data;
   return bar();
})
.then(data => $scope.b = data)
.catch(console.error)
```

Is equivalent to:

```js
Promise.all([foo(),bar()])
.then(data => {
   $scope.a = data[0];
   $scope.b = data[1]
})
.catch(console.error)
```

If have any question, just ask me.

Best regards, Jesús.

havidarou:
LGTM!

PR #87, opened 2017-11-24
Enhacements in agents loading. Removed unneccessary extra api call

jesusgn90:
Agents preview is now faster. Also I noticed a duplicated api call that was inneccessary. Improved way to request data.


havidarou:
LGTM!

PR #92, opened 2017-11-27
3.0 several CSS fixes

JuanjiJG:
* The index selector from our Discover tab doesn't resize oddly.
selector-fixed
* The Manager/Ruleset buttons are clearing showing what the current sub-tab is opened.
buttons
* The pointer is now default when hovering a metric visualization
untitled
* The Panels/Discover buttons now change color correctly according to the hover and current section.
* Fixed the Discover icon not appearing on the Agents/Overview section.

(Sorry about the huge cursor, it's a problem with my 4k display).


havidarou:
LGTM!

PR #93, opened 2017-11-27
Some loops enhacements

jesusgn90:
- Replaced some angular.forEach loops by native loops
- Better and shorter removeFilter function on datahandler


havidarou:
LGTM!

PR #98, opened 2017-11-28
Ensure syscheck and rootcheck are showing properly

jesusgn90:
- It's only a `$scope.$digest()` (that only applies current $scope, not the whole app)
- Just noticed that `syscheck` and `rootcheck` minutes, sometimes they are showing and sometimes they aren't so I force to ensure update them when finished.
- Also adds an extra condition in order to prevent wrong offset with datahandler 
  - `if(this.offset !== 0 && this.offset >= this.totalItems){ ... }`
- Fixed wrong `hover` in groups tab and agents preview tab,


havidarou:
LGTM!

PR #102, opened 2017-11-28
Fix groups and agents preview

jesusgn90:
- New controller, new view and new route
- Agents preview controller now has a new function named `showAgent` that works same as the groups `showAgent` function.
- Also that function is now called from the html view of "agents preview".
- That function put on the `$rootScope` a variable named `globalAgent` with the desired agent and a variable named `comeFrom` with the values `groups` or `agentsPreview` depends on the controller, and then `$location.path('/agents')`

Hope it helps!


havidarou:
LGTM!

PR #105, opened 2017-11-28
Settings fix

jesusgn90:
- Change top menu on settings tab
- Sending only agent id to rootScope from agents and groups tabs


havidarou:
LGTM!

PR #107, opened 2017-11-29
Fix app typography to gain consistency

JuanjiJG:
This PR changes the Wazuh app typography in order to gain more consistency between the app and the rest of Kibana sections.
font

This is more noticeable if you look at the Kibana sidebar when on Discover/Visualize and Wazuh.


havidarou:
LGTM!

PR #111, opened 2017-11-30
Fix API is down situation

jesusgn90:
- Check if the API is down when starting the app, if the API is down we force the user stay on settings tab and show help messages to handle this situation.
  - The error is also catched from back-end to front-end
- This PR also fix the decoders and the rules tables on manager/ruleset
- Added rule level filter on ruleset tab


havidarou:
LGTM!

PR #112, opened 2017-11-30
Firefox 57 bugfixes

JuanjiJG:
* Fixed width and height of some visualizations
* More width fixes in Manager/FIM
* Fixed Manager/Logs input filter box


havidarou:
LGTM!

PR #116, opened 2017-12-01
Minor enhacements dev

jesusgn90:
This PR fix the bellow list:
- Logs table are now sortable by all fields except description
- Removed unneccessaries tooltips (redundant information)
- Added `.catch` statement to catch errors in multiple places where were not being catched.
```js
foo()
.then(() => { ... })
.catch(error => notify.error(error.message));
```


havidarou:
LGTM!

PR #124, opened 2017-12-11
New logic on groups controller

jesusgn90:
- There is a new card on the agent view to show its group
- Now it's possible to click on the group card from an agent and go to the groups section
- The boolean to check if groups section is loading is now in the right place
- If you come from an agent to groups section, the opened group is the agent group

Hope it helps :+1:


havidarou:
LGTM!

PR #129, opened 2017-12-12
Removed inline CSS from every template

JuanjiJG:
This PR removes all the inline `style` tags from the templates and puts classes instead. The `.less` files have not been moved or deleted.


havidarou:
LGTM!

PR #130, opened 2017-12-12
Fix autocomplete

jesusgn90:
Affected "autocompletes":
- Ruleset/Rules
- Ruleset/Decoders
- Agents

Fixed features:
- Search by any term
- Next page of data when scroll
- Search with lucene syntax (eg. group: syslog)
  - Can press "Enter" key to set a filter


havidarou:
LGTM!

PR #137, opened 2017-12-15
Minor fixes

jesusgn90:
- Menu top directive now is not hardcoded
- Fix a wrong condition when formatting agent status
- Moved each template to its own folder
- Fix template route on routes
- Also removed an unused template


havidarou:
LGTM!

PR #138, opened 2017-12-18
3.0 Regex validations

jesusgn90:
- Validating form fields when adding a new API (**frontend**)
- Validating form fields when adding a new API (**backend**)
- Fix wrong line getting manager info


havidarou:
LGTM!

PR #148, opened 2017-12-21
Fixed extensions not working

JuanjiJG:
This PR makes extensions working again.

There's a known bug in the frontend when you enable/disable extensions and just after that you change to another API from the list. The workaround is to reload the page after selecting the new API.


havidarou:
LGTM!

PR #163, opened 2017-12-27
Fixes for the logo and visualizations

JuanjiJG:
This PR solves two things:

* Adjust the logo from the app to have less space between the icon and the name.
* Reduces the heigh from some visualizations in Agent/General.


havidarou:
LGTM!

PR #164, opened 2017-12-27
Multiple fixes

jesusgn90:
- Fixes blank fields when "check manager" is clicked after add an API
- Prevents any trigger when go from a tab to the same tab
- Fixes agents status visualization on Overview


havidarou:
LGTM!

PR #165, opened 2017-12-28
Some API fixes

jesusgn90:
Summary:

- Solved an error ocurred when you try to connect to the API through `https` and it's not under https
- Removed wrong code and rewritten `checkManager` function code
- More error situations are now properly handled
- Minor JS optimizations

### DIFF
--- a/public/controllers/settings.js
+++ b/public/controllers/settings.js
@@ -174,7 +174,6 @@ let app = require('ui/modules').get('app/wazuh', []).controller('settingsControl
             cluster_info: {},
             insecure:     'true',
             id:           (Array.isArray($scope.apiEntries)) ? $scope.apiEntries.length : 0,
-            cluster_info: null,
             extensions:   { oscap: true, audit: true, pci: true }
         };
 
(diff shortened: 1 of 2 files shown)


public/controllers/settings.js
```diff
@@ -291,43 +288,30 @@ let app = require('ui/modules').get('app/wazuh', []).controller('settingsControl
...
-            'id':           $scope.apiEntries[index]._id
+        const tmpData = {
+            user:         $scope.apiEntries[index]._source.api_user,
+            password:     $scope.apiEntries[index]._source.api_password,
+            url:          $scope.apiEntries[index]._source.url,
+            port:         $scope.apiEntries[index]._source.api_port,
+            cluster_info: {},
+            insecure:     'true',
+            id:           $scope.apiEntries[index]._id,
+            cluster_info: null
```
snaow: cluster_info was previously defined. Keep just one, isn't it?

jesusgn90: That line of code initializes `tmpData` in a const object, after that we give a value to `cluster_info`, before this PR it was cluster_info: {}, but is indiferent cos we assign a direct value after that.

PR #166, opened 2018-01-05
Now checks if .data exists before accessing it

jesusgn90:
`if(data.data.data.apiIsDown)` could crash if the last `.data` doesn't exists, now we check if it exists before accessing `apiIsDown`

`if(data.data.data.apiIsDown)`  ==> `if(data.data.data && data.data.data.apiIsDown)`


havidarou:
LGTM!

PR #173, opened 2018-01-12
Dev new agent config tab

JuanjiJG:
This PR adds a brand-new tab for the Wazuh App, called **Agent/Configuration**. This tab is located on the detail page for an Agent, next to the PCI DSS tab.

Example screenshot of the tab:
192 168 1 63_5601_prc_app_wazuh

* This tab shows the current [Centralized configuration](https://documentation.wazuh.com/current/user-manual/reference/centralized-configuration.html). Currently it has sub-tabs for FIM, PM and Log Collection, more tabs will be added on the future.

* The tab shows the current group of the agent, and also the current **sync status** (that means, if the configuration you're seeing is actually loaded by the Agent, comparing the `merged_sum` hashes).

* This tab will disappear if you go to a Manager, since they cannot belong to a group.

* If you click on the Group name, it will redirect you to Manager/Agents, and select automatically the group of that Agent.

* The tab has a JSON viewer, you can enable it with the switch control, located at the upper left corner of the tab. The viewer shows the `agent.conf` content in `JSON` format.


havidarou:
LGTM!

PR #185, opened 2018-01-17
Small bugfixes

JuanjiJG:
This PR does the following:
* Fix the height of some metric visualizations where the numbers were cut.
arreglo 1
* Now all the data table visualizations should not have their pagination cut and the title has been changed to **Alerts summary**.


havidarou:
LGTM!

PR #194, opened 2018-01-18
Regular user expression is now less restrictive

jesusgn90:
- `/^[a-zA-Z0-9]{3,100}$/` is now `/^.{3,100}$/` (any character, min length: 3)


havidarou:
LGTM!

PR #196, opened 2018-01-18
(Fixed) More CSS bugfixes

JuanjiJG:
This pull request fixes some bugs **related to the CSS and the visualizations**, such as:

* Unified colours across the whole app.
ruleset

* Fix the readability issues with some text.
agents

* Fix the width of some visualizations who doesn't show properly on small screens.
width


havidarou:
LGTM!

PR #210, opened 2018-01-25
Removed old Wazuh App blue color from more app components

JuanjiJG:
Hi everyone,

This PR removes the old blue colour from some components of the app and uses a new, darker one which is more consistent with the new style of the app for Kibana 6.

On the attached screenshots, we can see an example of components with the modified colour.

navbar

md-switch

captura


havidarou:
LGTM!

PR #218, opened 2018-01-29
New index pattern selector on the navbar

JuanjiJG:
Hello,

This PR adds **an index-pattern selector** on the upper right corner of the app, next to the `Settings` button.

It works the same as the current selector on the `Settings` page. The app implements now a new Service called `patternHandler`, which will provide the functionality for changing the index-pattern.

In addition, this PR adds `logall_json` to Manager/Configuration.


havidarou:
LGTM!

PR #224, opened 2018-01-30
New Vulnerabilities tab

JuanjiJG:
Hello everyone,

This PR adds a new tab for the Wazuh App: **Vulnerabilities**

192 168 1 56_5601_ogv_app_wazuh

This is a first prototype for this tab. In the future, more visualizations and data will be added.

The tab is not an extension, so it will be always present on `Overview` and `Agents`.


havidarou:
LGTM!

PR #226, opened 2018-01-31
Patch PCI DSS tab

jesusgn90:
This PR fixes some PCI DSS tab mistakes, now we have a more in deep list of PCI requirements obtained from [Prioritized-Approach-for-PCI_DSS-v3_2.pdf](https://www.pcisecuritystandards.org/documents/Prioritized-Approach-for-PCI_DSS-v3_2.pdf). Also modifies the PCI backend route as follow:

- Read our requirements list(composed by key + description items)
- Request the api call `/rules/pci`
- Checks which keys returned from api match with our requirements list and compose the final response.


havidarou:
LGTM!

PR #227, opened 2018-01-31
Patch index pattern selector

jesusgn90:
This pull request fixes the following mistakes:
- Pattern handler service because it was not implemented fine due to a wrong use of the promises. 
- Typo on settings.js 
- The menu-top directive is reimplemented in order to work properly with the pattern selector.


havidarou:
LGTM!

PR #228, opened 2018-01-31
Prevents overwrite of settings on the scope variable

jesusgn90:
This pull request ensures to only overwrite scope.extensions on settings if there is no current pattern (there is no api)


havidarou:
LGTM!

PR #249, opened 2018-02-08
Better Elasticsearch error handling

jesusgn90:
Hi App team! after a first approach of our error handler module, I've been researching more complicated scenarios, after some simulations I catch some situations where a better error handling is needed, with this PR we have a better error hanling module. 

It's mostly focused on Elasticsearch related errors.

Best regards,
Jesús


havidarou:
LGTM!

PR #256, opened 2018-02-12
Minor fixes (6.1.x)

jesusgn90:
- Replaces the login route name from /login by /wlogin to avoid conflict with xpack own login route.
- Shows index pattern selector only if the list length is greater than 1, if it's exactly 1 shows the index pattern without a selector.


havidarou:
LGTM!

PR #257, opened 2018-02-12
Rules patch (6.1.x)

jesusgn90:
- Fixes a wrong situation with the visualizations when we refresh on rules/decoders tab


havidarou:
LGTM!

PR #261, opened 2018-02-13
Fixes api list wrong sorting

jesusgn90:
This PR fixes the order we are showing the list of stored APIs, now they are sorted by timestamp, the newest is the first now.


havidarou:
LGTM!

PR #262, opened 2018-02-13
Removed kbn-dis from ruleset and added proper logic on vis directive

jesusgn90:
- This PR removes the discover directive from Ruleset tab and modifies a bit the visualization directive to properly loading these visualizations.


havidarou:
LGTM!

PR #266, opened 2018-02-19
Minor bugs fixes

jesusgn90:
Hi team, this PR fixes the following mistakes:

- Fix a trouble related to _hardcoded_ extensions and minor problem when is the very first time you open the app
- Fix minor design issue with the rules/decoders tables
- The linear progress bar is now within the proper directive
- Prevents a dangerous access to an object when it hasn't got the desired member on the tables directive


havidarou:
LGTM!

PR #273, opened 2018-02-22
Fix wrong agent id parsing when come from groups twice

jesusgn90:
This PR fixes a wrong data flow when go from groups tab to a specific agent and the come back to groups tab. Once you are again on groups tab and go to a different agent it remains the last agent. Now it's working as expected.


havidarou:
LGTM!

PR #274, opened 2018-02-23
3.2 minor bugs patch

jesusgn90:
Hello team, this PR fix the following bugs:

- Minor issue with some table fields, now are adjusted
- Click on groups should reload groups view
- Wrong sorting on all tables using wz-table, this was an error since long time ago, but now it works as expected.


havidarou:
LGTM!

PR #275, opened 2018-02-23
No id on kbn-vis

jesusgn90:
- Hi team this PR makes a little fix on the visualization directive and now it's no longer neccessary the `id` field when using a `kbn-vis` element.
- Also all `kbn-vis` elements have no `id` field now.


havidarou:
LGTM!

PR #276, opened 2018-02-26
Added complete group config to JSON viewer

JuanjiJG:
Hello all,

This PR will add the **complete JSON output** from the API when requesting the group configuration data. Previously, it only showed the first one, when no filters were applied.

This will help to see if more configuration is available through the JSON viewer until we finally decide the functionality of the `Agent configuration` tab.

agents

Regards,
Juanjo


havidarou:
LGTM!

PR #283, opened 2018-03-01
Hotfix to current visualization directive

jesusgn90:
Hello team, this pull request fixes a bug with the kibana visualization directive itself. We were parsing on a wrong way some filters, now the function behind this situation is better and safer and fixes the bug. We need to test it better but it fixes two wrong scenarios:

- Duplicates inside the implicit filter
- False `false` values in some conditional statements

Best,
Jesús


havidarou:
LGTM!

PR #293, opened 2018-03-05
Prevents accesing undefined array position

jesusgn90:
Hello team, this pull request only fixes a little bug related to the visualization directive, there is a unsafe array access that could cause troubles if you go directly to ruleset/decoders leaving the visualizations on a weird white box.


havidarou:
LGTM!

PR #296, opened 2018-03-06
Query filter patch

jesusgn90:
Hello team, this pull request prevents double agent.id filter on agents tab. If you go to the agents tab and go into an agent and then change the agent, the agents are accumulated, it's wrong, now it's fixed. Read the code comments bellow for a better understanding.


havidarou:
LGTM!

PR #298, opened 2018-03-07
Hotfix minor bugs

jesusgn90:
Hello team, this pull requests fixes two bugs:

- If you are on Overview and/or Agents and open Discover and then go to another sub tab, the watchers for the metrics were being lost, now it's fixed.
- The rule level on the ruleset tab was wrong whenever it was 0, now it's fixed.


havidarou:
LGTM!

PR #304, opened 2018-03-08
Manager Configuration readjustments

JuanjiJG:
Hello all,

This PR brings two new changes to the app:
* Recovered the previous state of the redesigned *Manager/Configuration* tab, adapting the HTML code to use the refactored CSS.
* Fixed a misplaced tooltip in the *Settings* tab.

This PR will not be added to the CHANGELOG, since it's part of the refactoring process and there's already a log of the tab redesign itself.

Regards,
Juanjo


havidarou:
LGTM!

PR #307, opened 2018-03-12
Logs

jesusgn90:
Hello team, this pull request adds a new feature for the Wazuh App:

- New module named `logger` to write logs
- Currently there are two modules using the logger: initialize.js and monitoring.js,
this way we can get the whole trace whenever the server is started/restarted.
- Also makes a minor code refactor

_Note: the logs will live under `/var/log/wazuh/wazuhapp.log` on Linux systems, on Windows systems they will live under `kibana/plugins/`. It rotates the log whenever it reaches 100MB._

Best regards,
Jesús


havidarou:
LGTM!

PR #308, opened 2018-03-12
Better checks

jesusgn90:
Hello team, this pull request add some improvments to the Wazuh App:

- Now we have a new field on the `.wazuh-version` index named `lastRestart`, it keeps the last kibana restart date.
- The citated field is used to check if the cookies have consistency with the current sever status. 
- Now the app is clever and take decissions depending on new consistency checks.

It will help to prevent weird situations on our users machines whenever they reinstall the app, reload Kibana, etc. It has been tested on several ways, hope it help us.

Best regards,
Jesús


havidarou:
LGTM!

PR #309, opened 2018-03-12
Agents configuration tab redesign

JuanjiJG:
Hello all,

This PR brings a new redesign to the app. This time is for the **Agents configuration** tab.

### Description
This redesign is based on the same made for the **Manager configuration tab**. With this UI revamp, some modifications were made to the side nav cards and were applied to both tabs, providing consistency between them.

For example, the `fa-eye` icon was removed, and replaced by a bold-text active element indicator. When the user hovers the title, the font will turn bold, and also a tooltip will appear, indicating the possible actions.

The current group and sync status indicators were replaced by the new information ribbon, the same that we can see on other app sections.

In addition to this, this PR fixes some minor bugs on the UI, and also removes old CSS classes, continuing with the huge work done with the HTML & CSS removal and cleanup process.

### Sample screenshot
rediseno

nogroup

This redesign is one more step into a better app design, achieving consistency between pages and a better overall look.

Best regards,
Juanjo


JuanjiJG:
Don't worry about the huge amount of commits. With this branch, I tried to solve a weird situation where the `3.2` branch was 63 commits behind the `master` branch. This doesn't make any kind of conflicts.

PR #311, opened 2018-03-13
Fix API version check, dismissing patch number

jesusgn90:
Hello team, this pull request allows us to have Wazuh App compatible with Wazuh APIs who has different patch number.


havidarou:
LGTM!

PR #312, opened 2018-03-13
Groups on agents

jesusgn90:
Hello team, this pull request adds the following enhacements:

- Added missing group column on agents preview
- Implements filter by group on agents preview
- Added the option to go directly to the agent group on agents preview

_Note: since the API still has no way to filter by group or sort by group, that code is commented up to the API is ready._

Best regards,
Jesús


havidarou:
LGTM!

PR #313, opened 2018-03-13
Optional index pattern selector

jesusgn90:
Hello team, this pull request add the next new feature:

- New option on the config.yml file to set if we should show the index pattern selector on the Wazuh App top menu or not.

_Note: the new option is `ip.selector`, if not present or if it has a `true` value, it will be `true`._

Best regards,
Jesús


havidarou:
LGTM!

PR #326, opened 2018-03-20
Refactored and redesigned top navbar

JuanjiJG:
Hello all,

This PR brings a new refactor and redesign for the app. This time is for the top nav bar!
prototipofinal

This refactoring process cleaned the HTML/CSS code, and reduce the number of needed classes to give style to the bar. Now it uses a `md-toolbar` component instead of a `md-navbar` component, which also fixes several visual bugs.

Also, with this PR, we recover the ability to open new tabs with the mouse wheel when clicking the different navbar buttons. This allows reloading the list of agents on the `Agents` tab, instead of using <kbd>F5</kbd>.

I hope you like this new design, every feedback is really appreciated.

Regards,
Juanjo


havidarou:
LGTM!

PR #331, opened 2018-03-22
Better API error message on settings

jesusgn90:
Hello team, this small pull request only give more information to the user who is facing problems whenever adds a new API entry, thanks.

Best,
Jesús


havidarou:
LGTM!

PR #332, opened 2018-03-22
Truncated text for tables and metric ribbons

JuanjiJG:
Greetings,

This pull request adds truncatable text to the app. Specifically, to the metric ribbons and the tables. We've implemented a new CSS class, `wz-text-truncatable`, used for text containers (not the text element itself):
```CSS
.wz-text-truncatable {
    overflow: hidden;
    white-space: nowrap;
    text-overflow: ellipsis;
}
```
In addition, the table from *Manager Logs* is configured to not truncate the **Description** column, so the user won't miss the complete log.

### Example screenshots
overflow1
overflow2

Regards,
Juanjo


havidarou:
LGTM!

PR #333, opened 2018-03-23
Better index pattern list handling

jesusgn90:
Hello team, this pull request fixes mainly 3 things:

- Redirects to blank screen with the proper message whenever index pattern list length is 0
- Overwrite current pattern cookie if it is no longer avalaible
- Ignore index patterns with custom id on the initialization process, instead it uses config.yml default pattern

Best,
Jesús


havidarou:
LGTM!

PR #342, opened 2018-03-26
Fixed multiple agent.id issue

jesusgn90:
Hello team, this _one line pull request_ only fixes the issue related to multiple agent.id filters whenever go from agents to groups and come back with different agent.

Best,
Jesús


havidarou:
LGTM!

PR #348, opened 2018-04-03
Fixed flickering on Settings tab

JuanjiJG:
Hello team!

This pull request fixes a weird flickering on the *Settings* tab, due to a situation where the template was loaded before knowing if there were API entries loaded, and after that, hiding again a card.

Best regards,
Juanjo

### DIFF
--- a/public/templates/settings/settings.html
+++ b/public/templates/settings/settings.html
@@ -15,7 +15,7 @@
     <div layout="column" layout-align="start stretch">
 
         <!-- First-time welcome card -->
-        <md-card class="wz-md-card" layout-padding layout-align="start stretch" ng-if="submenuNavItem == 'api' && apiEntriesLoaded && apiEntries.length == 0">
+        <md-card class="wz-md-card" layout-padding layout-align="start stretch" ng-if="submenuNavItem == 'api' && !load && apiEntries.length == 0">
             <md-card-content>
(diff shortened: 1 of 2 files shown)


public/templates/settings/settings.html
```diff
@@ -15,7 +15,7 @@
     <div layout="column" layout-align="start stretch">
 
         <!-- First-time welcome card -->
-        <md-card class="wz-md-card" layout-padding layout-align="start stretch" ng-if="submenuNavItem == 'api' && apiEntries.length == 0">
+        <md-card class="wz-md-card" layout-padding layout-align="start stretch" ng-if="submenuNavItem == 'api' && apiEntriesLoaded && apiEntries.length == 0">
```
jesusgn90: @JuanjiJG  we could use `load` variable who is already defined, replace `&& apiEntriesLoaded` by `&& !load`.
Since we can use it, you don't need any change con `settings.js` file, so fix it please. You should try it before any change because I didn't try it, but it should work, thanks.

JuanjiJG: You're right, colleague. Now it's fixed, and the new variable was removed from the code. Now there are only CLRF fixes on the `settings.js` file. Thanks for your valuable feedback, it's really appreciated.

Regards,
Juanjo

jesusgn90:
LGTM

PR #351, opened 2018-04-04
Dynamic visualizations seed

jesusgn90:
Hello team this PR is only to merge two development branchs into single branch.


jesusgn90:
OK!

PR #358, opened 2018-04-11
Fix shards zero-issue

jesusgn90:
Hello team, this PR fixes wrong situation whenever shards value is zero and/or replicas value is zero.

Best,
Jesús


havidarou:
LGTM!

PR #363, opened 2018-04-16
Initialize fix

jesusgn90:
Hello team, reviewing the 2.x to 3.x upgrading process I've catched two minor bugs. This pull request fix two minor bugs related to the recently merged branch, which changes all the server-side. 

- The first fix is related to this line ` const id = config._id;` before the fix, we were missing the `_id` value.
- The second fix is related to ignore the old `wazuh-monitoring` pattern to be refreshed.
- Additionally I've changed the error message `checkWazuhVersionIndex` because the .wazuh-version throws an error whenever it doesn't exists and its right, but not enough to generate an error.

Best,
Jesús

### DIFF
--- a/server/initialize.js
+++ b/server/initialize.js
@@ -273,7 +273,7 @@ module.exports = (server, options) => {
             try{                
                 await wzWrapper.getWazuhVersionIndex();
             } catch (error) {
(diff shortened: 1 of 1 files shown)


jesusgn90:
LGTM

PR #368, opened 2018-04-16
Fix visualizations cron task log message

jesusgn90:
Hello team, this pull request fixes the next issue:

- https://github.com/wazuh/wazuh-kibana-app/issues/361

```js
const output = await wzWrapper.deleteVisualizationByDescription(date.getTime(),true);
if(output && output.deleted > 0) log('vis-deletion-cron.js',`Deleted ${output.deleted} visualizations with a timestamp lower than ${date.getTime()}`,'info');
```
Also it logs the number of deleted visualizations when we delete one or more visualizations.

Best,
Jesús


jesusgn90:
LGTM

PR #377, opened 2018-04-17
Added cluster.name to monitoring index

jesusgn90:
Hello team, this pull request ables the App to store the `cluster.name` property on the monitoring daily index. Assuming the agent from the `/agents` request comes with the field  `{ cluster: { name: 'wazuh-cluster-01' } }` we'll store it as it. If the agent doesn't come with that field, we'll store it as `{ cluster: { name: 'disabled' } }`. On the other hand, whenever we are on a non-cluster environment we'll still able to know it from the `.wazuh` index like on the other app packages.

This PR solves https://github.com/wazuh/wazuh-kibana-app/issues/375

Best,
Jesús


havidarou:
LGTM! Now we will properly filter clusters in our monitoring visualization!

PR #391, opened 2018-04-19
Removed implicit filter from PCI 

jesusgn90:
Hello team, this pull request removes the implicit filter from PCI to avoid a multi-search which crashes the very first time you enter Overview-PCI or Agent-PCI.

Since the visualizations from PCI has their own filter to check PCI related alerts, we don't need that filter which only causes overload to Elasticsearch.

Also adds this condition to not push empty arrays (it's not harmful but not needed to do it):

```js
if(cleaned.length) queryFilter.addFilters(cleaned);
``` 

Best,
Jesús


JuanjiJG:
I'll proceed to close this issue, since it needs more development and further research.

Regards,
Juanjo

PR #394, opened 2018-04-20
Fix agents autocomplete

jesusgn90:
Hello team, this pull request replaces an if statement on the discover directive by a trace deletion on the agents controller. This way we are reaching the same goal in a cleaner mode, also reduces overload for the discover directive.

Best,
Jesús


havidarou:
LGTM!

PR #413, opened 2018-04-25
Added errorHandler service to bs controller, better error parsing

jesusgn90:
Hello team, this pull request parses raw error objects from some scenarios on the blank screen controller, this way we can show the user the exactly message not the raw output. I'm using a try/catch block because this controller needs to be always avalaible, so if the parsing fails we can show the raw any way.

Regards,
Jesús


havidarou:
LGTM!

PR #415, opened 2018-04-25
Fix false "not enough privileges" scenario

jesusgn90:
Hello team, this pull request fixes https://github.com/wazuh/wazuh-kibana-app/issues/414, read it to have a more in deep view about the problem.

Brief summary of the change:

Whenever `allow.hits.total >= 1` it means the user has enough privilege to use that index pattern. This is always true unless there is no indices matching that pattern. If `allow.hits.total === 0` but `!forbiden` the user still has enough privileges to use that pattern but no indices are being matched --> visualizations without results.

Regards,
Jesús


havidarou:
LGTM!

PR #420, opened 2018-04-26
New Manager/Ruleset redesign

JuanjiJG:
Hello all,

This PR bring a new redesign for the Wazuh app: the **Manager/Ruleset** tab:

ruleset

### Main description
This new design divides the Ruleset tab into two different sections: ***List view*** and ***Detail view***. The old expandable tables have been replaced with a new section that will help us for improving the information that we show from Rules and Decoders.

The filter functionality is still present. You can apply search filters and use a Lucene-like query syntax.

On the ***Detail view*** you can see all the information about the Rule or Decoder, and also you can see a list of related Rules/Decoders.

If a rule has groups or PCI requirements, you can click on them to go back to the list with that filter applied.

### Other changes
This redesign comes with a bunch of small improvements to the overall user interface. For example, the font size on some elements has been reviewed and it's a bit smaller.

This new tab will improve a lot in the future, because now we have more space to implement more rich information for every rule or decoder.

Regards,
Juanjo


havidarou:
LGTM!

Great work!!

PR #428, opened 2018-04-30
Added message for successful API removal

### LINKED ISSUE #424: Add success/error messages when deleting an API entry
Hello team,

Currently, when the user deletes an API entry from the *Settings* tab, there's no message indicating if the process finished well or not.

We should add this so the user could have a better user experience. Every process made on the aff should prompt a confirmation message.

Regards,
Juanjo

JuanjiJG:
Hello team,

This pull request adds a new message when the user deletes an API entry on the Settings tab:
message

Now not only the user will know if something went wrong, he'll also confirm that the deletion process was successful.

This PR closes #424.

Regards,
Juanjo


havidarou:
LGTM!

PR #445, opened 2018-05-10
Fix requests

jesusgn90:
Hello team, this pull request improves the way we handle multi-requests, timeout, etc.

- Added a 200ms delay when fetching iterables
- Prevents from searching on rules/decoders with a string length <= 1
- Prevents from search if the factory is busy or it's ended
- Properly handling cancelled requests

Best regards,
Jesús


havidarou:
LGTM!

PR #452, opened 2018-05-15
Fix Kibana filters bug

jesusgn90:
Hello team, this pull request fixes https://github.com/wazuh/wazuh-kibana-app/issues/451

Brief summary:

- Added new class to build queries from base query
- Removed `rison` dependency since it's no longer needed
- The filter management is being done on controllers instead the `Discover` directive
- Now we are emitting specific events whenever we are fetching data or communicating to the `Discover`  directive
- The number of useless requests to fetch data has been reduced
- The synchronization actions are working as expected regardless the amount of data and/or the amount of the machine resources
- Added factories to store shared information between services and/or controllers, leaving the root scope ligther
- Removing useless listeners whose could cause wrong situations

Best regards,
Jesús


havidarou:
LGTM! Faster!

PR #456, opened 2018-05-16
Added check for new extensions

jesusgn90:
Hello team, this pull request fix an inconsistency situation whenever we add a new extension to the App which is enabled by default.

The problem comes for old App packages, whose already have `.wazuh` documents created without the new extension. If the new extension is enabled on the `config.yml` file the App thinks it is disabled. 

With this pull request we are ensuring we have all the .wazuh documents with the proper extensions updated. If the document already had a extension, it is ignored, otherwise we update it adding the new extension using the default value.

To check the extensions the function reads the `config.yml` and compares it with `default-ext.js` extensions. If it's not found con `config.yml`, looks for it on `default-ext.js`. Finally looks if the current `.wazuh` document being analyzed has that extension, if it hasn't got it, the extension is added to its extensions with the default value, otherwise that extension is skipped for that `.wazuh` document.

From now and onwards we need to explicitly add any new extension to `default-ext.js` with its default value.

This PR fixes https://github.com/wazuh/wazuh-kibana-app/issues/455

Regards,
Jesús

### DIFF
--- a/server/initialize.js
+++ b/server/initialize.js
@@ -17,6 +17,7 @@ import ElasticWrapper     from './lib/elastic-wrapper'
 import packageJSON        from '../package.json'
 import kibana_template    from './integration-files/kibana-template'
 import getConfiguration   from './lib/get-configuration'
+import defaultExt         from './lib/default-ext'
 
 export default (server, options) => {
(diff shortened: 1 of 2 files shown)


havidarou:
LGTM!

PR #459, opened 2018-05-16
Fix discover to panels preserve extra filters on filter-bar

jesusgn90:
Hello team, this pull request fix a little non-desired behaviour whenever a user has a custom non-pinned filter on any overview/agent tab and goes to Discover sub tab then come back to Panels sub tab.

The trick here was to add a boolean variable to check if I'm on the same tab but come from Discover or if indeed I'm new for this tab. If `localChange` is true it means we are on that situation, otherwise it'll be false.

Finally, and using `localChange`, we can decide what to do on the right way:

```js
$state.filters = localChange ? $state.filters : [];
```

Best regards,
Jesús

### DIFF
--- a/public/controllers/agents.js
+++ b/public/controllers/agents.js
@@ -102,7 +102,7 @@ app.controller('agentsController', function ($timeout, $scope, $location, $rootS
     };
 
     let filters = []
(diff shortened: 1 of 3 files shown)


havidarou:
LGTM!

PR #461, opened 2018-05-16
Fix changing agent not render

jesusgn90:
Hello team this pull request prevents from removing visHandlers whenever we only change the current agent on Agents section.

Regards,
Jesús


havidarou:
LGTM!

PR #465, opened 2018-05-17
Fix extensions wrong situation

jesusgn90:
Hello team, this pull request fix a bug related to Settings and the Wazuh API extensions.

- Keeps the current API entry index updated any time you toggle an extension
- Extensions tab is hidden if there are no API entries
- Any case if you are on extensions tab but there are no API entries on the controller, the function returns empty

Regards,
Jesús


havidarou:
LGTM!

PR #486, opened 2018-05-21
Prevents stop searching if "this.end" is true

### LINKED ISSUE #472: Minor bug while searching using the data-handler factory
Whenever the data-handler is on an "end" status, it doesn't search any more. This bug has a simple fix, but we need to test it, so this fix will be done on further packages. 

Regards,
Jesús

jesusgn90:
Hello team, this pull request fixes #472 however we shouldn't merge it before the new package is released. It has been tested only under a development environment and the Agents list. It would be better if we await for next test session to verify it.

Regards,
Jesús


havidarou:
LGTM!

PR #488, opened 2018-05-23
Fix ignoring enter on main Discover

jesusgn90:
Hello team, this pull request fixes https://github.com/wazuh/wazuh-kibana-app/issues/487, the solution is to ignore our filter restrictions if we are on the main Discover cause we don't need it and they won't be satisfied never.

Regards,
Jesús


havidarou:
LGTM!

PR #493, opened 2018-05-25
Prevents "000" agents in md-autocomplete

jesusgn90:
Hello team, this PR fixes https://github.com/wazuh/wazuh-kibana-app/issues/492. The only one modification is in the data-handler factory. 

> If we are fetching agents, exclude any "000" agent before assign the array

It's useful for the whole app, and after think about it I concluded it's the best place to do the filtering.

Regards,
Jesús


havidarou:
LGTM!

PR #504, opened 2018-05-28
Fix check connection feature

### LINKED ISSUE #503: Check API connection not working on Settings
Since the last changes we've lost that feature cause we removed clear password from Settings controller. Now we need to accept the API id on server side as parameter.

Regards,
Jesús

jesusgn90:
Hello team, this pull request fixes #503 

Brief summary:

- The route checkAPI now accepts a Wazuh API id, this way we can use it to check already stored APIs or new APIs


havidarou:
LGTM!

PR #531, opened 2018-06-06
Fix agents issue

jesusgn90:
Hello team, this PR adds a pretty simple control logic to prevent problems regarding to no visualization tabs vs visualization tabs on agents. Same logic will be applied to Overview once the redesign finishes.

- Now we are storing 3 items array which includes 3 last tabs. If A is a visualization tab and B is a no visualization tab and we have A,B,A it means Kibana won't repeat the same search so we can manage that situation like going from discover to panel (local change)

Regards,
Jesús


havidarou:
LGTM!

PR #533, opened 2018-06-07
Reduced vis tab checking to 2 elements

jesusgn90:
Since we'll have more than one agents tab without visualizations in further packages we've fixed the solution previously merged to be a bit more generic.

Now it only stores last two tabs only if their content includes visualizations. This way we can simply check `arr[0] === arr[1]` to detect if it's a local change.

Regards,
Jesús


havidarou:
LGTM!

PR #548, opened 2018-06-08
Fix agent changed condition

jesusgn90:
The team just found a little bug when changing agent and going to the same tab regarding to the agent.id filter. This pull request prevents that situation.

Regards,
Jesús


havidarou:
LGTM!

PR #556, opened 2018-06-08
Fix set default manager resets extensions

jesusgn90:
Hello team this PR prevents from change the extensions from the user if click on set default manager.


havidarou:
LGTM!

PR #586, opened 2018-06-19
Agents preview improvement

### LINKED ISSUE #578: Move some "agents" logic to server-side
Currently we are doing some tasks whose could be heavy for certain browsers and we can achieve them obtaining same result but from our server-side.

Regards,
Jesús

jesusgn90:
Hello team, this PR fixes #578 .

Since this view could has a huge amount of data we concluded to move some of its controller logic to our server side. Once the Wazuh API can provide us a list of unique fields from the data sets we can replace this logic.

Regards,
Jesús


havidarou:
LGTM!

PR #601, opened 2018-06-20
New extension - CIS-CAT alerts

JuanjiJG:
Hello team

This PR adds a new extension for the app: **CIS-CAT** alerts.

ciscat

This is a new extension, disabled by default, for the **Overview** and **Agents** tabs. If you're using the CIS-CAT wodle, now you can see the alerts on the app, with useful information about scans, and their results.

As the rest of tabs, you can generate a report of this tab, and as every extension, you can enable or disable it.

I hope you like this new tab.

Regards,
Juanjo


havidarou:
LGTM! Great work!

PR #610, opened 2018-06-25
Fixed flex layout on Ruleset tabs

### LINKED ISSUE #609: Fix button overflow on Management/Ruleset
Hello team,

We have to fix this overflow on the Ruleset buttons when displaying the app on small screens. The search bar should be the flexible component and adjust its width to the remaining space, and the buttons should always use the required space to not overflow the label.

Example on Rules:
ruleset
Example on Decoders:
ruleset2

Regards,
Juanjo

JuanjiJG:
Hello team,

This pull request fixes #609.

Now the Ruleset buttons always take the needed space to not overflow the labels. The search bar and Decoders filter now are also taking the space properly, according to its component type.

Some example screenshots:
### Before:
ruleset
ruleset2

### After:
rules
decoders

Regards,
Juanjo


havidarou:
LGTM!

PR #642, opened 2018-06-29
Reporting minor improvements

jesusgn90:
Hello team, this PR adds some reporting minor improvements

- Some kbn-vis from html that had the id located in the wrong place. 
- Adds a validation for the Elastic parser, each row must has same length as columns length
- Adds agent id for reporting
- Fixes some fonts size and margin from generated PDF

Regards,
Jesús


havidarou:
LGTM!

PR #644, opened 2018-06-29
Review for Kibana tables

jesusgn90:
Hello team, this PR modifies some Kibana tables due to performance reasons. Also fixes a minor bug from the Discover search bar(https://github.com/wazuh/wazuh-kibana-app/commit/0a53d4087802a8d256d26aba336c0e3d222d606b) and makes the reporting tables take full width when rendering (https://github.com/wazuh/wazuh-kibana-app/commit/548960c06088976e1c30820648b8c6d399742208).

On the other hand, the Elastic parser itself has been modified to accept more kind of data tables (https://github.com/wazuh/wazuh-kibana-app/pull/644/commits/b7474043084b8990758b418a64923dc50796694f). Now we can handle different bucket aggregation structures and in a safer way.

This doesn't mean these data tables will be added to the reporting module cause they also need to match our current parser expected structure.

Regards,
Jesús


havidarou:
LGTM!

PR #647, opened 2018-07-02
Updated health check behaviour

### LINKED ISSUE #638: Change health check
Currently we have a blocking health check and it's fine but we can do it better. This task is not defined at all, but the desired goal is to have a health check like the health check from "Kibana" and to have more in deep information about the environment where is running Kibana plus the Wazuh App.

Regards,
Jesús

jesusgn90:
Hello team, this PR fixes #638 

Regards,
Jesús


havidarou:
LGTM!

PR #670, opened 2018-07-09
Configuration factory

jesusgn90:
Hello team, this PR fixes https://github.com/wazuh/wazuh-kibana-app/issues/656

- Now we have a new **factory** which stores all the configuration values
- Added a new **resolve** function before each app route, this way we have the configuration always up to date
- Also we are showing the current configuration under settings tab, this was added in https://github.com/wazuh/wazuh-kibana-app/commit/0f8ce068a8f63ccbce6a29e27a93c78a5e0c5330 but could be removed

Regards,
Jesús


havidarou:
LGTM!

PR #696, opened 2018-07-10
Checking if filter is negated

### LINKED ISSUE #695: [Reporting] Negated filters not working as expected
Hello team, working into https://github.com/wazuh/wazuh-kibana-app/tree/issue-651 I've found a little bug in our reporting module. E.g: `agent.id: 000 AND NOT rule.level: 7` is being parsed as `agent.id: 000 AND 
 rule.level: 7` which is wrong. Fortunately our current reports are only showing wrong the filters but applying them properly cause they are applied under the hood before generate the report.

On the other hand we need to fix this situation for the PDF itself and for further Elasticsearch queries that we are going to do soon in https://github.com/wazuh/wazuh-kibana-app/tree/issue-651

Regards,
Jesús

jesusgn90:
Hello team, this PR fixes #695 

Regards,
Jesús

### DIFF
--- a/server/controllers/wazuh-reporting.js
+++ b/server/controllers/wazuh-reporting.js
@@ -139,8 +139,8 @@ export default class WazuhReportingCtrl {
         for (let i = 0; i < len; i++) {
             const filter = filters[i];
             str += i === len - 1 ?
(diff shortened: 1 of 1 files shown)


havidarou:
LGTM!

PR #700, opened 2018-07-12
Fix searching index pattern

jesusgn90:
Hello team, this PR fixes https://github.com/wazuh/wazuh-kibana-app/issues/699. I've found a wrong function in our server side. That function is currently being used to get the top active agent in Elasticsearch and it had the _wazuh-alerts-3.x-*_ pattern hardcoded. 

The bug was not so harmful but since we are going to use that function with the reporting module in further packages, it must accept dynamic index patterns. 

Regards,
Jesús


havidarou:
LGTM!

PR #725, opened 2018-07-17
WIP: Replace use of mapping types with "type" field in document

### LINKED ISSUE #724: Deprecate use of mapping types
ElasticSearch is in the process of [deprecating mapping types](https://www.elastic.co/guide/en/elasticsearch/reference/6.3/removal-of-types.html). Our own production infrastructure is future-proofing this by enforcing a mapping type of `_doc` on all new documents. In our case, this means we cannot save the configuration because the wazuh plugin attempts to add the type "wazuh-configuration" to the configuration documents; ES6 removes support for multiple mapping types on a single doc.

The suggested workaround in Elastic's documentation is to add your own `type` field and filter on that, with the understanding that this really only fits at the application level.

I'm working on a PR to take a stab at implementing this.

Disclaimer: I am fairly new to working with ES at this level, so please correct me if my vocabulary or understanding is wrong.

Off topic: is there an IRC channel or other public realtime chat facility where y'all hang out? `#wazuh` on Freenode seems dead, there's only one other person in there besides me.

fuhry:
Closes #724

WIP. Needs:

* Review
* Testing
* Code to migrate existing documents by adding the `type` field, or removal of types entirely


jesusgn90:
Closing this PR for now, once Kibana 7.0.0 is released we probably have ton of new code in our App, then is the time to review this problem. In any case, thanks for your advice and you are always welcome @fuhry ! 

Regards,
Jesús

PR #758, opened 2018-07-30
Fix specific node vis (cluster.node missing)

### LINKED ISSUE #757: Missing filter for specific node vis
Hello team, our section named Management > Monitoring is missing the `cluster.node` filter when looking for specific node because we are not providing it to the directive. The solutions is as simple as:

```js
const monitoringFilter = discoverList[1].filter(item => item && item.meta && item.meta.key && (item.meta.key.includes('cluster.name') || item.meta.key.includes('cluster.node')));
                    
// Applying specific filter to cluster monitoring vis
if (Array.isArray(monitoringFilter) && monitoringFilter.length) {
   visualization.searchSource.set('filter',monitoringFilter);
}
```

Regards,
Jesús

jesusgn90:
Closes #757


havidarou:
LGTM!

PR #782, opened 2018-08-10
[3.5-6.3] Fixed Decoders detail view when only parents filter enabled

### LINKED ISSUE #781: Decoders details view can't be opened when "Parents only" filter is enabled
The click event on the decoders doesn't work when the "Parents only" filter is enabled.

JuanjiJG:
Hello team,

This pull request fixes #781. The bug was caused by a different path on Decoders when the `Only parents` filter was enabled, and it was not triggering the event to open the Detail view.

Regards,
Juanjo


havidarou:
LGTM!

PR #790, opened 2018-08-16
[3.5-6.3] Fix error when first line is a comment or empty

### LINKED ISSUE #776: Dev-tools throws error when the first line is a comment
This error happens when the user leaves the Dev tools tab with a comment on the first line.

I think this is caused by the fact that the tab will always execute the first line when the user opens the app, and if it's a comment, it won't work.

devtools

Regards,
Juanjo

jesusgn90:
Hello team, this PR fixes #776 

Tested using empty lines, comment lines, mixed lines (comments and empty lines) and valid query line.

Regards,
Jesús


havidarou:
LGTM!

PR #1036, opened 2018-11-20
Added ports to syscollector network ports table

juankaromo:
Hi team,

This PR adds the missing 'Ports' columns to syscollector network ports table.

Regards


juankaromo:
Ports are marked as non-sortable because the API doesn't allow it.

PR #1078, opened 2018-12-11
Fix step in manual

Rohlik:
Because we stop Kibana in first step, final step should be `start` instead of `restart`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,13 +98,13 @@ Restart Kibana
 - Systemd:
 
 ```
-systemctl restart kibana
+systemctl start kibana
 ```
 
 - SysV Init:
 
 ```
-service kibana restart
+service kibana start
 ```
 
 ## Older packages


jesusgn90:
Hi @Rohlik ,

Thanks a lot, but I have some questions:

- Is there a problem using `restart`?
- What about people that skipped the first step (`stop`), is it still valid to use `start` instead of `restart`?

Regards!

Rohlik:
If you thinking that way then yes, there is no problem using `restart` instead of just `start`.
But there is theoretically no reason for doing first step (`stop`) if we will use `restart`.

jesusgn90:
Hello again @Rohlik ,

You are right and I'm with you, it's recommended doing the first step because all modifications to any service should stop the service before doing it but as I said we want to ensure Kibana is re/started properly and since `restart` makes a start when the service is stopped I prefer to leave the README as it.

Is it okay for you @Rohlik ?

Regards

Rohlik:
Hi @jesusgn90 ,
I am very grateful for your explanation. Thank you. I will try find another way how can I help to contribute to Wazuh.
Have a nice day ;-)

PR #1088, opened 2018-12-13
Fix typo new filters bar

juankaromo:
Hi team,

This PR fix an unhandled casuistic with blankspaces and case sensitive tags in new filters bar.

Regards


jesusgn90:
Thanks @juankaromo !

Issue #1080, opened 2018-12-12
Edit our config.yml file from the app


jesusgn90:
- [x] The section Settings > Configuration view must be modified to add "edit" options.
- [x] The section Settings > Configuration must always show the applied settings.
- [x] The app must advise the user if a Kibana restart is needed in order to apply the new settings.
- [x] We always must have consistency between the applied settings and the uncommented settings under `config.yml`

juankaromo:
Working on this branch https://github.com/wazuh/wazuh-kibana-app/tree/issue-1080

PR #1105, opened 2018-12-20
Edit config.yml file from the app

juankaromo:
Hi team,

This PR implements a new feature to be able to edit config.yml file from the Settings > Configuration section view in the app.

This PR solves #1080 

Regards!


public/controllers/settings/settings.js
```diff
@@ -75,6 +76,10 @@ export class SettingsController {
     // Tab names
     this.tabNames = TabNames;
     this.configuration = wazuhConfig.getConfig();
+    this.configurationTypes = [];
+    for (var key in this.configuration) {
```
jesusgn90: Use `const` instead of `var`

public/controllers/settings/settings.js
```diff
@@ -207,6 +212,20 @@ export class SettingsController {
     return;
   }
 
+  // Get configuration file
+  async setValueConfigurationFile(key, value) {
+    const data = {
+      key: key,
```
jesusgn90: Use _short_ syntax `key: key` should be `key`

public/controllers/settings/settings.js
```diff
@@ -703,4 +722,45 @@ export class SettingsController {
...
+    this.editingNewValue = value;
+  }
+
+  /**
+   * Change value for a given configuration key
+   * @param {String} key Configuration key
+   * @param {String} newValue new configuration value for key
+   */
+  editKey(key, newValue) {
+    this.setValueConfigurationFile(key, newValue);
```
jesusgn90: This function returns a promise, you are not handling it.

public/controllers/settings/settings.js
```diff
@@ -703,4 +722,45 @@ export class SettingsController {
...
+    this.errorHandler.handle(
+      'You must restart kibana for the changes to take effect', '', true
```
jesusgn90: kibana -> Kibana

server/controllers/wazuh-utils.js
```diff
@@ -43,6 +44,25 @@ export class WazuhUtilsCtrl {
...
 
+    /**
+   * Returns the config.yml file in raw
+   * @param {Object} req
+   * @param {Object} reply
+   * @returns {Object} Configuration File or ErrorResponse
+   */
+  updateConfigurationFile(req, reply) {
+    try {
+      const configFile = updateConfigurationFile(req);
```
jesusgn90: We don't want two functions with the same name, please.

server/lib/update-configuration.js
```diff
@@ -0,0 +1,63 @@
...
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Find more information about this on the LICENSE file.
+ */
+import fs from 'fs';
+import readline from 'readline';
+import path from 'path';
+
+export function updateConfigurationFile(req) {
```
jesusgn90: This function must be async, we don't want read lines using events. Even if we do that, we want promises always, so you may want to wrap a callback function into promise function.

server/lib/update-configuration.js
```diff
@@ -0,0 +1,63 @@
...
+        || trimLine.indexOf('#' + req.payload.key + ':') >= 0)
+        && trimLine.indexOf('.' + req.payload.key + ':') === -1) {
+        notFound = false;
+        findedLine = line.replace(/(\r\n\t|\n|\r\t)/gm, '').trim();
+      }
+    }).on('close', () => {
+      if (notFound) {
+        fs.appendFile(customPath, '\r\n' + req.payload.key + ':' + req.payload.value, function (err) {
+          if (err) {
+            return false;
```
jesusgn90: This return has no effect

server/lib/update-configuration.js
```diff
@@ -0,0 +1,63 @@
...
+        })
+      } else {
+        let currentValue = findedLine.split(':');
+        if (currentValue[1]) {
+          currentValue = currentValue[1].trim();
+          if (currentValue === 'true' || currentValue === 'false') {
+            currentValue = currentValue === 'true';
+          }
+          if (typeof req.payload.value != typeof currentValue) {
+            return false;
```
jesusgn90: This return has no effect

server/lib/update-configuration.js
```diff
@@ -0,0 +1,63 @@
...
+          }
+        }
+        const result = raw.replace(findedLine, req.payload.key + ' : ' + req.payload.value);
+        fs.writeFile(customPath, result, 'utf8', function (err) {
+          if (err) {
+            return false;
```
jesusgn90: This return has no effect

server/lib/update-configuration.js
```diff
@@ -0,0 +1,63 @@
...
+          }
+        });
+      }
+    });
+    return true;
```
jesusgn90: This function is weird, and it's not fine. If the instance is a bit slow all the function is broken. This function must be re-written, thanks for the effort but we must review it together is not the idea behind this task.

server/routes/wazuh-utils.js
```diff
@@ -23,6 +23,15 @@ export function WazuhUtilsRoutes(server) {
     }
   });
 
+  // Returns the config.yml file in raw
+  server.route({
+    method: 'PUT',
+    path: '/utils/updateconfiguration',
```
jesusgn90: The path must be `/utils/configuration` because we have a GET route for the same file and this is a PUT route.

public/controllers/settings/settings.js
```diff
@@ -703,4 +722,45 @@ export class SettingsController {
...
+  }
+
+  /**
+   * Enable edition for a given key
+   * @param {String} key Configuration key
+   */
+  setEditingKey(key, value) {
+    if (typeof value === 'object') {
+      try {
+        value = JSON.stringify(value);
```
jesusgn90: Why we need a string instead of an object?? We should try to avoid any stringify/parse always if it's possible

Issue #1162, opened 2019-01-21
Restart agents from the app


jesusgn90:
Tasks:

- [x] Add button to restart it (API call)

juankaromo:
Working on this branch https://github.com/wazuh/wazuh-kibana-app/tree/issue-1162

pablotr9:
Today's work:

- [ ] UI improvement in agent
- [ ] Show feedback when an agent is upgrading
- [ ] Test that an agent upgrade is successful and update feedback.
Currently, the agent is giving a time out when upgraded that has to be fixed

pablotr9:
Today's work:

- [ ] Fix the feedback shown when an agent is upgrading.
- [ ] Fix timeout using "wait_for_complete" param when upgrading an agent.
When testing upgrading an agent, the API call replies an error 1715-"Error sending WPK file."

juankaromo:
Today work
- [x] Making fixes and merging 3.9 to do PR

PR #1229, opened 2019-02-11
Restart agents

juankaromo:
Hi team,

This PR adds new functionality to be able to upgrade the version of an agent. It's added new buttons in the agent's preview table and in an agent's main section to perform this action. When clicking, the upgrade process is started.

Unfortunately, we can't have feedback about the process at the moment. So, every time we fetch the agent, it will be checked if it is outdated and if it is in an update process. If so, a message will appear indicating it. Otherwise, if the process is over, neither the update button nor the message indicating the process will appear.

On the other hand, it has been added a new button that allows restarting the agent.

This PR solves #1162 

Regards.


juankaromo:
Today's work

- [x] Refactor agent actions adding new block
- [x] Delete upgrade agent from agents table
- [x] Added restart agent option

PR #1406, opened 2019-04-24
Add timezone offset to dates

juankaromo:
Hi team,

This PR adapts the dates shown throughout the whole app, which come in UTC, adapting them to the browser's time zone. For this purpose, a date transformation service has been created, which currently contains the function to add the offset of the browser to the date that arrives from the server.

The dates of the scans, the dates included in the wz-table, in reports table, and the dates of the logs have been adapted.

This PR solves https://github.com/wazuh/wazuh-kibana-app/issues/1380

Regards.


public/controllers/agent/agents-preview.js
```diff
@@ -92,6 +94,14 @@ export class AgentsPreviewController {
         !parameters.items.length > 0 && !parameters.filters.length;
     });
 
+    this.$scope.offsetTimestamp = (text, time) => {
```
jesusgn90: Move this method to an instance method.

```js
offSetTimestamp(text,time) {
 ...
}
```

Then, in the view, you can access it using `ctrl.offsetTimestamp`

public/controllers/agent/agents-preview.js
```diff
@@ -92,6 +94,14 @@ export class AgentsPreviewController {
...
+      try {
+        return text + this.timeService.offset(time);
+      } catch (error) {
+        return '';
```
jesusgn90: What the view looks when returning `''`? What happens?

public/controllers/agent/agents.js
```diff
@@ -309,6 +311,19 @@ export class AgentsController {
       );
     };
 
+    /**
+     * This adds timezone offset to a given date
+     * @param {String} binding_text
+     * @param {String} date
+     */
+    this.$scope.offsetTimestamp = (text, time) => {
```
jesusgn90: Very similar to the last comment, now, you still need to wrap to the scope, it's okay but this time move the logic to an instance method, then just wrap it using `this.$scope.offsetTimestamp = (text, time) => this.offsetTimestamp(text, time);`

public/controllers/misc/reporting.js
```diff
@@ -46,6 +47,14 @@ export class ReportingController {
     this.groupToPages();
   }
 
+  offsetTimestamp = time => {
+    try {
+      return this.timeService.offset(time);
+    } catch (error) {
+      return '';
+    }
+  };
```
jesusgn90: Remove `;`

public/controllers/misc/reporting.js
```diff
@@ -46,6 +47,14 @@ export class ReportingController {
     this.groupToPages();
   }
 
+  offsetTimestamp = time => {
```
jesusgn90: `offsetTimestamp = time => {` should be `offsetTimestamp (time) {`

public/controllers/misc/reporting.js
```diff
@@ -46,6 +47,14 @@ export class ReportingController {
...
+    try {
+      return this.timeService.offset(time);
```
jesusgn90: Here is not needed to add text as a parameter like in the previous methods of this PR?

public/directives/wz-table/lib/parse-value.js
```diff
@@ -23,7 +23,13 @@ const checkIfArray = item => {
   return typeof item === 'object' ? splitArray(item) : item === 0 ? '0' : item;
 };
 
-export function parseValue(key, item, instancePath, $sce = null) {
+export function parseValue(
+  key,
+  item,
+  instancePath,
+  $sce = null,
```
jesusgn90: `null` --> `false`

public/directives/wz-table/lib/parse-value.js
```diff
@@ -23,7 +23,13 @@ const checkIfArray = item => {
...
+  timeService = null
```
jesusgn90: `null` --> `false`

public/services/time-service.js
```diff
@@ -0,0 +1,28 @@
...
+  /**
+   * Returns given date adding the timezone offset
+   * @param {string} date Date
+   */
+  offset(d) {
+    try {
+      const date = new Date(d);
+      const offset = new Date().getTimezoneOffset();
+      const offsetTime = new Date(date.getTime() - offset * 60000);
+      return offsetTime.toLocaleString('en-ZA').replace(',', '');
```
jesusgn90: Why `en-ZA`?

public/services/time-service.js
```diff
@@ -0,0 +1,28 @@
...
+    } catch (error) {
+      return Promise.reject(error);
```
jesusgn90: Why are you returning a rejected promise in a non-async method???

public/templates/agents/agents-welcome.html
```diff
@@ -124,13 +125,13 @@
...
-                                || '-'}}
+                            <p class="euiTitle euiTitle--small euiStat__title" ng-if="!agent.rootcheck.inProgress"
+                                    ng-bind="agent.rootcheck.end ? offsetTimestamp('', agent.rootcheck.end) : '-'">
                                 <md-tooltip
-                                    ng-if="!agent.rootcheck.inProgress && agent.rootcheck.start && agent.rootcheck.end"
+                                    ng-show="!agent.rootcheck.inProgress && agent.rootcheck.start && agent.rootcheck.end"
                                     md-direction="bottom" class="wz-tooltip">
-                                    Start time: {{ agent.rootcheck.start || '-'}} <br>
-                                    End time: {{ agent.rootcheck.end || '-'}} <br>
+                                    Start time: {{ agent.rootcheck.start ? offsetTimestamp('', agent.rootcheck.start) : '-'}}&nbsp;&nbsp;|&nbsp;&nbsp;
```
jesusgn90: Can we avoid the usage of `&nbsp`?

public/controllers/agent/agents-preview.js
juankaromo: I removed the function from this controller because it wasn't being used.

public/services/time-service.js
juankaromo: BEcause according to this: http://www.localeplanet.com/icu/en-ZA/index.html, this country (South-Africa) uses the exact format we need (yyyy/MM/dd hh:mm:ss).

public/controllers/misc/reporting.js
juankaromo: No, because in the reporting section is not needed to append a label before the date (e.g: Last scan: )

public/controllers/agent/agents-preview.js
juankaromo: Nothing was shown. Now I have changed it so that it returns the date and the base timezone. `return '${time} (UTC)';`

Issue #1436, opened 2019-05-09
Fix configuration validation error management


juankaromo:
Hi team,

We need to fix how we are handling the error output of the `configuration/validation` API call. Currently, we are filtering by path (in order to only show an error if we are editing a file of this path), but some error replies don't contain the path, so we are not showering any error.

To solve this we are going to show the error messages regardless of where we are.

Regards.

juankaromo:
**Update**
In this PR #1437 a temporary fix has been implemented, displaying the error messages in any path, because at the moment not all messages contain the path of the file it belongs to. 

This is being fixed in @wazuh/core , and until it is implemented, I mark this as `pending-core`.

PR #1437, opened 2019-05-09
No filter by path in config validation

juankaromo:
Hi team,

This PR stops filtering by the current path in files configuration validations.
This PR solves #1436 

Regards.


jesusgn90:
Please, find a proper way to do not show different section errors. For example, we don't want to show decoder errors in rules. On the other hand, show all once a restart is fired.

Issue #1379, opened 2019-04-15
[visualize] Fix top line of "Agents status" visualization


jesusgn90:
Hi team, we need to found a way to make the top line be fully displayed and not just half of the line as the screenshot below:

![image](https://user-images.githubusercontent.com/8860227/56146473-90334600-5fa6-11e9-8395-d65ceccc047b.png)

Regards

juankaromo:
**Update**
Fixed the problem of 'covered dots' with these patch styles:

```
.visChart__container .chart > svg{
  position: absolute;
}

.visChart__container svg g.points.line, 
.visChart__container svg g.points.area {
  display: contents;
}
```

jesusgn90:
The next line is unsafe:

```js
if ((visParams || {}).valueAxes && (this.visData.series || [])[0]) {
```
Replace it with the next ones:

```js
const valueAxes = (visParams || {}).valueAxes || false;
const hasSeries = ((this.visData || {}).series || []).length;
if (valueAxes && hasSeries) {
```

PR #1463, opened 2019-05-23
Fix covered points in visualizations

jesusgn90:
Hi team, 

This PR manages to show the top line of visualizations and not just half of the line as the screenshot below:

image

This PR solves #1379 

Regards

### DIFF
--- a/public/kibana-integrations/loader/visualize_data_loader.ts
+++ b/public/kibana-integrations/loader/visualize_data_loader.ts
@@ -111,6 +111,14 @@ export class VisualizeDataLoader {
       );
     }
 
+    if (((this.vis || {}).params || {}).valueAxes && (this.visData.series || [])[0]) {
+      const maxValue = Math.max.apply(Math, this.visData.series[0].values.map(x => { return x.y; }));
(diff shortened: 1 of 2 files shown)


jesusgn90:
Hi @juankaromo, this is not the solution we are looking for. 

image

We don't want to put the circle over the visualization, we need to increase the max value, for example, total hits + 1 should be enough for most cases.

Regards

juankaromo:
**Update**

I have left behind the first approach of the solution by CSS and I have chosen to implement a method that obtains the value of the highest data and from adding its unit (1, 10, 100, 1000...) to the maximum value of the `y axis`.

https://github.com/wazuh/wazuh-kibana-app/blob/7653165fbd82d83f5f3f615c69b595aa858f8768/public/kibana-integrations/loader/visualize_data_loader.ts#L114-L120

juankaromo:
I did some changes in the method in order to not apply it in area-based visualizations. Now it looks like this:

https://github.com/wazuh/wazuh-kibana-app/blob/1ddf8b4ec7eade908d55bb380e546b2d4682d440/public/kibana-integrations/loader/visualize_data_loader.ts#L114-L123

PR #1566, opened 2019-07-01
Fix hanging up when editing cdb list key

adri9valle:
Hi team,

This PR solves: https://github.com/wazuh/wazuh-kibana-app/issues/1485

Regards,


adri9valle:
This PR is closed because the error that it solves will be solved in other PR which goes to pointing to another App version.

Issue #1610, opened 2019-07-17
Defects reported by the DevOps team for v3.9.3


jesusgn90:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 3.9.3   | 7.2     | 0523 |

- [x] Reporting for Overview > Security Events doesn't work if _wazuh.monitoring_ is disabled due to the Agent Status visualization is missing.
- [x] Console error in PCI-DSS
![image](https://user-images.githubusercontent.com/8860227/61357561-cd92c280-a878-11e9-9da5-8e6be9d5d6e2.png)
- [x] Policy monitoring has unrelated visualizations
  - Top 5 PCI DSS
  - Top 5 CIS
- [x] Path filter doesn't work for custom rules. Using the dropdown or writing it by hand, doesn't work anyway.
- [x] Inconsistent data in VirusTotal dashboard. The number of files is not represented properly in the table
- [x] Different size between created and implicit filters. Add a filter manually. Notice the size difference.
- [x] Problem with SCA filter. <del>Navigate to Agents > SCA. Check that there are alerts. Then go to Overview > SCA. The agent.id filter is kept and it should be removed. It doesn't happen when going to other dashboards so it's fully related to SCA</del>.
- [x] Syscheck visualizations inconsistencies. <del>The number of top agents and top users doesn't match</del>
- [x] Rules and decoders tables don't show the path column. But the path filter is allowed. Check the available columns to select, the path column is not included.
- [x] Grammar typo in VirusTotal dashboard. <del> "No results displayed because all values equal 0." The rest of the empty visualizations displays different messages. </del>
- [x] In Docker section. Empty visualizations when there are no related data. There are related data, but not being used by the visualizations.
- [x] In Osquery section. Osquery alert doesn't resolve variable content | Rule.id: 24052
- [x] Duplicated failing requests (error code 500) are being launched when entering the app.
- [x] Remove Overview > SCA dashboard
- [x] <del>Alerts table is unordered. Review all table visualizations.</del> Ordered by Alerts count
- [x] Weird color selection in several visualizations.
- [x] Ruleset 
  - [x] Make columns _level_, _group_, _pci_, _gdpr_ clickable columns, then it adds a filter for the table. 
  - [x] Make the file column clickable, it should show the file content. 
  - [x] Must be changed in the rule list and the related rules list inside a rule. 
  - [x] Apply the same for decoders. 
[...]

PR #1615, opened 2019-07-18
Defects in v3.9.3

juankaromo:
#1610 and prepares branch to Wazuh 3.9.4


public/controllers/management/components/reporting-table.js
```diff
@@ -32,7 +32,7 @@ export class ReportingTable extends Component {
   filterItems(pageIndex, pageSize) {
     const { items } = this.state;
     const pages = [];
-    const len = items.length;
+    const len = (items || []).length;
     for (let i = 0, j = len; i < j; i += pageSize) {
```
jesusgn90: `let i = 0, j = len; i < j; i += pageSize` -> `let i = 0; i < len; i += pageSize`

public/controllers/management/decoders.js
```diff
@@ -335,4 +343,6 @@ export class DecodersController {
   restart = () => {
     this.$scope.$emit('performRestart', {});
   };
+  
+
```
jesusgn90: Remove this empty line

public/controllers/management/management.js
```diff
@@ -92,6 +92,9 @@ export class ManagementController {
     this.$scope.$on('removeCurrentConfiguration', () => {
       this.currentConfiguration = false;
     });
+    this.$scope.$on('viewFileOnly', (ev, params) => {
+      $scope.$broadcast('viewFileOnly', { file:params.item, path:params.path });
+    });
```
jesusgn90: Add an empty line on top and bottom, please

public/controllers/management/management.js
```diff
@@ -272,8 +275,11 @@ export class ManagementController {
     this.breadCrumbBack();
   }
 
-  switchFilesSubTab(flag) {
+  switchFilesSubTab(flag,showFile) {
     this.managingFiles = flag || true;
+    if(showFile){
```
jesusgn90: Why we need this `if`? why not directly `this.showFile = showFile` ?

public/controllers/management/rules.js
```diff
@@ -381,4 +381,13 @@ export function RulesController(
     $scope.overwriteError = true;
     $scope.$applyAsync();
   });
+
+
```
jesusgn90: Remove this empty line

public/directives/wz-table/wz-table-directive.js
```diff
@@ -270,6 +271,10 @@ app.directive('wzTable', function() {
       $scope.parseValue = (key, item) =>
         parseValue(key, item, instance.path, $sce, timeService);
 
+      $scope.parseKey = (key) => {
+        return typeof key === 'string' ? key : key.value;
```
jesusgn90: `return key ? key.value || key : key;` I think it's better

public/directives/wz-table/wz-table-directive.js
```diff
@@ -299,11 +304,11 @@ app.directive('wzTable', function() {
       $scope.prevPage = () => pagination.prevPage($scope);
       $scope.nextPage = async (currentPage, last = false) =>
         pagination.nextPage(currentPage, $scope, errorHandler, fetch, last);
-      $scope.firstPage = function() {
+      $scope.firstPage = function () {
```
jesusgn90: Add an empty line on top and on bottom

public/directives/wz-table/wz-table-directive.js
```diff
@@ -487,23 +492,48 @@ app.directive('wzTable', function() {
...
           item.showTooltip[id2] = true;
         }
-        $c.remove();
       };
 
+      const filterableColumns = () => {
+        $scope.filterableColumns = [];
+        $scope.keys.forEach(k => {
+          const key = $scope.parseKey(k);
+          const canFilter = (($scope.path === '/rules' && (key === 'level' || key === 'pci' || key === 'gdpr' || key === 'groups' || key === 'file' || key === 'path')) ||
```
jesusgn90: Split this into different variables

public/directives/wz-table/wz-table-directive.js
```diff
@@ -487,23 +492,48 @@ app.directive('wzTable', function() {
...
+            ($scope.path === '/decoders' && (key === 'path' || key === 'file')))
+          if (canFilter) {
+            $scope.filterableColumns[key] = true;
```
jesusgn90: Remove the `if`, and use the above variables:
`$scope.filterableColumns[key] = !!(var1 || var2 || var3..)`

public/directives/wz-table/wz-table-directive.js
```diff
@@ -487,23 +492,48 @@ app.directive('wzTable', function() {
...
+      $scope.handleClick = (key, item, ev) => {
+        const value = $scope.parseValue(key, item)
+        let keyTmp = $scope.parseKey(key);
+        const valueTmp = typeof value !== 'string' ? value.toString() : value;
+        if ($scope.filterableColumns[keyTmp]) {
+          if (value !== '-' && keyTmp !== 'file') {
+            // only 'group' is accepted as a filter so it has to be overwritten
+            if (keyTmp === 'groups')
+              keyTmp = 'group';
+            // PCI, GDPR and GROUPS can contains multiple values and the api only supports one at a time, just the first value is sent
```
jesusgn90: This doesn't make sense, instead, we should get the exactly clicked filter. If not applicable, we need a ticket opened in the Wazuh API and remove this code from here.

public/controllers/management/management.js
juankaromo: Because `showFilesSubtab` can be called from different locations, so, sometimes the `showFile` param could arrive empty. It's the expected behavior because we only want to set up the variable if the param is different from undefined.

public/directives/wz-table/wz-table-directive.js
juankaromo: It can make sense. When we ask for the rules, they can belong to one or several groups, so the object we receive is an array called groups:

```js
            "groups": [ 
               "syslog 
            ],
```
On the other hand, when we filter and do a search for rules, for groups it is only allowed to filter by some, so the filter is called `group`.

Then, if we click on the Key groups, we take the first group and we filter through it with the `group` parameter.

PR #1637, opened 2019-07-29
Fix pinned agent filters

juankaromo:
Hi team,

This PR solves the casuistic of having pinned agent filters across sections explained here https://github.com/wazuh/wazuh-kibana-app/issues/1626

Regards.

### DIFF
--- a/public/kibana-integrations/kibana-discover.js
+++ b/public/kibana-integrations/kibana-discover.js
@@ -15,7 +15,7 @@ import { uiModules } from 'ui/modules';
 import discoverTemplate from '../templates/kibana-template/kibana-discover-template.html';
 
 uiModules.get('app/wazuh', ['kibana/courier']).directive('kbnDis', [
-  function() {
+  function () {
     return {
       restrict: 'E',
       scope: {},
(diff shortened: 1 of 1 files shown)


jesusgn90:
We should not call `addFilters` inside an `addFilters` callback. 

Here is an idea that should be refined and tested:

```js
   ...
      const currentFilters = queryFilter.getFilters();
      const pinnedAgentIDs = currentFilters.filter(
        item =>
          ((item || {}).meta || {}).key === 'agent.id' &&
          ((item || {}).$state || {}).store === 'globalState'
      );

      const implicitAgentIDs = wzCurrentFilters.filter(
        item =>
          ((typeof item || {}).meta || {}).removable !== 'undefined' &&
          !((item || {}).meta || {}).removable &&
          ((item || {}).meta || {}).key === 'agent.id'
      );

      if(pinnedAgentIDs.length && implicitAgentIDs.length) {
        for(const filter of pinnedAgentIDs) {
          queryFilter.removeFilter(filter);
        }
      }
      
      queryFilter
        .addFilters(wzCurrentFilters)
        .then(() => {})
        .catch(error => console.log(error.message || error)); // eslint-disable-line
    }
  };
```

I can't merge this PR with your current solution @juankaromo

juankaromo:
It works perfectly and also improves the performance of my approach.

Thanks @jesusgn90 !

PR #1641, opened 2019-07-30
Fix delete api calls

adri9valle:
Hi team,

This PR solves https://github.com/wazuh/wazuh-kibana-app/issues/1555.

All `PUT` and `DELETE` methods were checked and fixed if it needed in order to make it compatible with Wazuh API 3.10.

Regards,


adri9valle:
This PR has been closed because we are checking in the `wazuh-api.js` if there is a DELETE request and transforms the body in a query string.

PR #1865, opened 2019-10-25
Remove format for URL Fields

jsanchez91:
In the Discover view, fields known as URLs are displayed with
the html tag <a>

We set them back to blank (default value),
if you remove the property `fieldFormatMap` the format does not change.

In future versions if Kibana solves this problem we will revert the
change, otherwise we remove the `fieldFormatMap` property.


juankaromo:
Thx @jsanchez91 , but it is fixed in KIbana 7.4.2

Issue #1947, opened 2019-12-04
Typo and inconvenient filesize limit


jctello:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 3.11.0   | 7.4.3     | 0561 |

**Description**
The list upload feature only accepts files up to 300KB. 
This value is insufficient to import the AlienVault IP reputation list which is mentioned in [this blog post](https://wazuh.com/blog/cdb-lists/) and is 1.23MB in size. 

Is there a reason for this limit, or could it be raised to 5MB?

Also the error message provided has a typo: "Allowd"
**Steps to reproduce**

1. Go to Management > Ruleset > Lists
2. Click on "Import Files"
3. Attempt to upload a file over 300KB
4. See error message with typo

**Screenshots**
![image](https://user-images.githubusercontent.com/36071202/70183765-a84b0780-16e6-11ea-8b86-43557b6cbe29.png)

juankaromo:
This fix will be included in Wazuh release 3.11.2

PR #1993, opened 2020-01-09
Increased list filesize limit and fixed typo

jctello:
As mentioned in #1947 the limit for uploading a list for the CDB-list capability is too low and the resulting error message contains a typo.

This PR addresses that by increasing the limit to 5MB and fixing the typo.


juankaromo:
Thanks @jctello

PR #2016, opened 2020-01-22
Fix configuration error

pablotr9:
Hi team,

When the configuration is edited multiple times in 10 seconds, the configuration is being duplicated as explained here: https://github.com/wazuh/wazuh-kibana-app/issues/2010


juankaromo:
thanks @pablotr9

Issue #2087, opened 2020-02-20
Move Wazuh app configuration file to optimize folder


juankaromo:
Hi team,

After the migration of the APIs configured in the app from the `.wazuh` index to the app configuration file `wazuh.yml` there was a problem when updating the app. 

Because of how Kibana manages the plugins, to make an upgrade it is necessary to delete the plugin and reinstall the new version. This causes that the configuration file (that is inside the path of the plugin) is lost and is restored by the file that comes included in the new version. So far this does not seem to have been a problem, but with the new API management, this problem becomes more important as the custom configured APIs are lost when upgrading.

**Upgrade Scenario from 3.10.**
Here there is no problem because the .wazuh index still existed and there is a procedure when initializing the app where the migration of the data from the index to the configuration file is done.

**Upgrade Scenario from 3.11.X to 3.11.y**
This scenario is the most conflictive because there is no migration and that is where the problem occurs. The solution is to specify in the upgrade guide of the Wazuh documentation that a manual backup of the configuration file `wazuh.yml` must be done, update the plugin, and then restore the saved file. Within versions 3.11.X no changes have been made to the file, so no settings have been added or removed that could break this backup.

**Future scenarios and proposed solution**
It is proposed to move the wazuh.yml file to the `/usr/share/kibana/optimize` directory, which is not removed when the plugin is updated. Before a new installation, the existence of this file will be checked. If it does not exist, it will be created with the current logic. If it already exists, no changes will be made.

In addition, a modification of the structure of this directory will be made, in order to have all the resources of Wazuh in the same path.

That's why the directory is going to take the following distribution:

```
/usr/share/kibana/optimize
└── wazuh/
    ├── logs/
    |    ├── wazuhapp.log
    |    └── wazuhapp-plain.log
    ├── downloads/
    |    └── reports/
    └── config/
         └── wazuh.yml
```

Regards,

pablotr9:
**Update**
Working on this branch: https://github.com/wazuh/wazuh-kibana-app/tree/3.12-7.6-optimize-wazuh-directory

As mentioned in the issue, the ../optimize/ folder has been restructured with this distribution:
```
/usr/share/kibana/optimize
└── wazuh/
    ├── logs/
    |    ├── wazuhapp.log
    |    └── wazuhapp-plain.log
    ├── downloads/
    |    └── reports/
    └── config/
         └── wazuh.yml
```

- [ ] Fresh installation 
[...]

PR #2116, opened 2020-03-03
Restructure optimize/wazuh folder

juankaromo:
Hi team,

This PR moves the wazuh.yml file to the /usr/share/kibana/optimize directory. 

In addition, it did a modification of the structure of this directory, in order to have all the resources of Wazuh in the same path.

This PR solves #2087 

The directory has now the following distribution:

```
/usr/share/kibana/optimize
└── wazuh/
    ├── logs/
    |    ├── wazuhapp.log
    |    └── wazuhapp-plain.log
    ├── downloads/
    |    └── reports/
    └── config/
         └── wazuh.yml
```

Regards,


server/lib/initial-wazuh-config.js
```diff
@@ -0,0 +1,140 @@
...
+#hosts:
+#  - <id>:
+#     url: http(s)://<url>
+#     port: <port>
+#     user: <user>
+#     password: <password> 
+
+hosts:
+  - default:
+     url: http://localhost
```
jsanchez91: Set the protocol as `https`

server/logger.js
```diff
@@ -37,7 +37,7 @@ const initLogger = () => {
       new winston.transports.File({
         filename: path.join(
           __dirname,
-          '../../../optimize/wazuh-logs/wazuhapp.log'
+          '../../../optimize/wazuh/logs/wazuhapp.log'
```
jsanchez91: The log folder path is repeated a lot, maybe you can save it to a constant. 
```js
const logsPath = '../../../optimize/wazuh/logs';

path.join(__dirname, `${logsPath}/wazuhapp.log`);
```

public/controllers/settings/components/logs.js
```diff
@@ -98,7 +98,7 @@ export default class SettingsLogs extends Component {
           <EuiFlexGroup>
             <EuiFlexItem>
               <EuiText color="subdued" style={{ paddingBottom: '15px' }}>
-                Log file located at /usr/share/kibana/optimize/wazuh-logs/wazuhapp.log
+                Log file located at /usr/share/kibana/optimize/wazuh/logs/wazuhapp.log
```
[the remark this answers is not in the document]
juankaromo: This is not within the scope of this issue. This was before this restructuring, so we should open issue to study it. But as far as this issue is concerned, everything is correct.

server/logger.js
juankaromo: Here if you are right, please change `../../optimize/wazuh/logs`  with a constant.

juankaromo:
LGTM, but we need to take into account that the removal plugin step in the upgrade guide (README.md) needs to be reviewed.

PR #2202, opened 2020-04-22
Improve index management and performance

pablotr9:
Hi team,

This PR solves:
- **Improve navigation performance** https://github.com/wazuh/wazuh-kibana-app/issues/2200. We were doing some API requests on every navigation, this PR solves these and allows the app to load new pages faster.
- **Migrate IndexPattern management to use the SavedObject APIs** https://github.com/wazuh/wazuh-kibana-app/issues/2197 The `wazuh-alerts-3.x-*` index patter was being created by directly writing into the .kibana index which is not recommended, the app now creates this index pattern in the health check using the Saved Object API (without restarting Kibana) it solves some problems when using Opendistro tenants or Kibana spaces

### DIFF
--- a/public/controllers/misc/health-check.js
+++ b/public/controllers/misc/health-check.js
@@ -97,6 +97,8 @@ export class HealthCheck {
         if (!patternData.status) {
           const patternList = await PatternHandler.getPatternList();
           if(patternList.length){
+            const currentPattern = patternList[0].id;
+            AppState.setCurrentPattern(currentPattern);
             return this.checkPatterns();
           }else{
             this.errors.push('The selected index-pattern is not present.');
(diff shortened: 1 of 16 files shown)


public/controllers/misc/health-check.js
```diff
@@ -283,6 +329,7 @@ export class HealthCheck {
       this.$scope.$applyAsync();
       return;
     } catch (error) {
+      console.log("getlist", error)
```
juankaromo, line 332: is it necessary?

public/services/resolves/get-ip.js
```diff
@@ -76,6 +85,7 @@ export function getIp(
         stateValFound: false
       });
     } catch (error) {
+      console.log("12", error)
```
juankaromo, line 88: Please remove the console logs and comments from this file

PR #2401, opened 2020-07-24
Avoid the blanc page in the agent detail view

jsanchez91:
Hi team!

This PR solved this Issue: https://github.com/wazuh/wazuh-kibana-app/issues/2398

For the first case, I modified the "VisRequest" component to show an error when the request to Elasticsearch does not return the object of the aggregation.

The second case occurs because the platform name is not defined in the "UnsupportedComponents" object. In this case, we return the list of the "other" key by default.

Regards, 
Jose


0xtaf:
Hi, how can we apply this change in kibana docker container?

PR #2472, opened 2020-09-21
Fix filters index pattern

### LINKED ISSUE #2471: [7.9] Error when adding filters to dashboards
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 3.13   | 7.9     |   |

Some filters that added by our App  are forced to use the index pattern with id: `wazuh-alerts-*`, if the index pattern is manually created using Kibana, the index pattern won't have that id `wazuh-alerts-*`, instead a random id will be generated.
Since 7.9, if the current index pattern ID doesn't match the filters index pattern, an error will be thrown:
image

We need to adapt all of our filters to add the current selected index pattern id, which is already stored in a cookie.

pablotr9:
closes #2471


jsanchez91:
LGTM

PR #2486, opened 2020-09-23
Add "sudo " prefixes in several locations to README.md

CODE-REaD:
In several locations, the instructions prefixed with "sudo " commands that must run as specific users (root, and kibana in one instance).
However, in many other locations "sudo" was not prefixed.  I have simply corrected this by prefixing "sudo " in those locations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,18 +119,18 @@ sudo chown kibana:kibana /usr/share/kibana/optimize/wazuh/config/wazuh.yml
 sudo chmod 600 /usr/share/kibana/optimize/wazuh/config/wazuh.yml
 ```
 
-Restart Kibana
+Start Kibana
 
 - Systemd:
 
 ```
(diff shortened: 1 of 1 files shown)


jsanchez91:
Some environments do not have the `sudo` command, for this reason, we do not include this in our guide.

CODE-REaD:
Dear @jsanchez91, just to clarify: the guide currently _does_ include sudo in several locations.  My changes are merely to make the use of sudo consistent throughout the guide, because it is confusing to instruct users to use sudo in some of the places it is needed in while not in others.

While I personally favor the use of sudo, the other option would be to eliminate sudo entirely and instead instruct users what account to run each command under.  For example,

As root, run:
`systemctl restart kibana`

As kibana, run:
`bin/kibana-plugin remove wazuh`

Thank you for your consideration.

PR #2543, opened 2020-10-15
[4.0-7.9] Fix app configuration cache

### LINKED ISSUE #2534: The configuration is applied after a period of time
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.x   | 7.x     | 444 |

When modifying the configuration from the `settings > configuration` view, the changes take some time to be applied.

configuration_delay

## Context:
To solve this error we must take into account the following elements:

**WazuhConfig service**: This object implements a singleton pattern to have a single instance with the configurations of the application. When changing views a request is made again to the backend to update the object containing the configuration.
https://github.com/wazuh/wazuh-kibana-app/blob/master/public/react-services/wazuh-config.js#L15

**getConfiguration method**: Due to the time it takes to obtain the configuration of the wazuh.yml file, the last reading of the file is stored in memory for 10 seconds.
https://github.com/wazuh/wazuh-kibana-app/blob/master/server/lib/get-configuration.js#L21

Desvelao:
Hello team,

this PR resolves:
-  Fix app configuration cache after updating it in WUI in:
  - Frontend-side
  - Server-side

close #2534

### DIFF
--- a/public/components/settings/configuration/components/bottom-bar.tsx
+++ b/public/components/settings/configuration/components/bottom-bar.tsx
@@ -24,7 +24,7 @@ import {
   EuiButtonEmpty,
   EuiButton
 } from '@elastic/eui';
-
+import { WazuhConfig } from '../../../../react-services/wazuh-config';
 
 
 interface IBottomBarProps {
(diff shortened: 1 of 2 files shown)


jsanchez91 (approved):
LGTM!

PR #2561, opened 2020-10-23
Add tool to check plugin version and clean browser data

jsanchez91:


public/app.js
```diff
@@ -9,6 +9,9 @@
  *
  * Find more information about this on the LICENSE file.
  */
+import { checkPluginVersion } from "./utils";
+checkPluginVersion();
```
Desvelao, line 13: Move the `checkPluginVersion()`'s function execution to `app.run` callback

public/utils/check-plugin-version.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  !_.isEqual(appInfo, JSON.parse(storeAppInfo)) && clearBrowserInfo(appInfo);
+}
+
+const deleteWazuhCookies = ([name, path]: cookie) => {
+  document.cookie = `${name}=; expires=Thu, 01 Jan 1970 00:00:00 UTC; path=${path};`;
+}
+
+function clearBrowserInfo(appInfo: TAppInfo) {
+  wazuhCookies.map(deleteWazuhCookies);
```
Desvelao, line 58: `Array.prototype.map` creates an array. You could use `Array.prototype.forEach` if you don't need the result array.

`deleteWazuhCookies` always sets the `document.cookie` to a string. Are you trying to concatenate for each element of `wazuhCookies`?

jsanchez91: The `map` has been changed to `forEach`

public/app.js
jsanchez91: I added a "try/catch" block to avoid errors in the application.

It is a pre-check to clean the data browser before the application starts, I think it should be executed as soon as possible to avoid the "health-check" to be executed in parallel.

jsanchez91: I moved the `checkPluginVersion` method into the `app.run` method and it works fine.

Desvelao (approved):

PR #2590, opened 2020-11-04
feat(frontend): added API user allow_run_as

### LINKED ISSUE #2576: [4.0-7.9] Run as with no wazuh-wui API user
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.x   | 7.x     | 400x |

**Description**
Requirement: allow no `wazuh-wui` API users can use `run_as`.
- [x] Check and cache in memory when Kibana starts the `run_as` property for each API
- [x] Add the check of `run_as` in Settings/API configuration
- [x] Add a column/icon with the `run_as` in Settings/API configuration

Desvelao:
- Created a cache in memory service to save the allow_run_as values
- Refactorized /hosts/apis endpoint to return the allow_run_as value for API
- Removed token from /host/apis response
- Refactorized /api/check-api to refresh and return the allow_run_as value
- Added a column in Settings/Api configuration to show the allow_run_as value

Updated:
- Added `run_as` setting in the host configuration
- The use of `run_as` depends on the run_as setting and if the Wazuh API user is allowed to use it

Updated:
- Show a warning icon when there is a misconfiguration with Wazuh API user and `run_as` setting
- Launch a toast when there is an error in the login

closes #2576


server/controllers/wazuh-api.js
```diff
@@ -406,6 +427,21 @@ export class WazuhApiCtrl {
...
+              'GET',
+              `${apiAvailable.url}:${apiAvailable.port}/security/users/me`,
+              {},
+              { idHost: req.payload.id }
+            );
+            if(responseApiUserAllowRunAs.status === 200) {
+              apiUserAllowRunAs = responseApiUserAllowRunAs.data.data.affected_items[0].allow_run_as;
+            }
+          }
+          CacheInMemoryAPIUserAllowRunAs.set(req.payload.id, apiAvailable.username, apiUserAllowRunAs);
```
frankeros, line 443: maybe both algorithms to check the apiUserAllowRunAs can be merged into a generic function, to avoid duplicate code.

server/controllers/wazuh-hosts.js
```diff
@@ -47,21 +50,30 @@ export class WazuhHostsCtrl {
...
-        if (removePassword) delete host.password;
-        joined.push(host);
-      });
-      return joined;
+        // Check if api.run_as is false or undefined, then it set to false in cache
+        if(!api.run_as){
+          CacheInMemoryAPIUserAllowRunAs.set(id, api.username, false);
+        };
+        // Add to run_as from API user. Use the cached value or get it doing a request
+        host.allow_run_as = CacheInMemoryAPIUserAllowRunAs.has(id, api.username)
```
frankeros, line 68: isn't the same validation into the function checkAPIUserAllowRunAS?

server/controllers/wazuh-hosts.js
```diff
@@ -117,4 +129,32 @@ export class WazuhHostsCtrl {
       );
     }
   }
+
+  async checkAPIUserAllowRunAs(apiId){
```
frankeros, line 133: this looks the same as the function into wazuh-api.js, if this is true you could put it into a separate file and use it from both classes.

frankeros:
I was testing the branch and I found thath if you add a new host with the property `run_as: true` in the wazuh.yml and then you enter with the user admin you have no the Wazuh rol `administrator` assigned. 
image
With default host selected logged in as admin user
image
With default2 host selected logged in as admin 
image

frankeros:
Another issue is happening when you try to generate sample data using `run_as:  false` (with true I didn't able to see this section, even as admin user)
image
the kibana server logs 
image
from https://github.com/wazuh/wazuh-kibana-app/blob/4.0-7.9-add-api-user-run-as/server/controllers/wazuh-elastic.js#L998

Desvelao:
I couldn't reproduce the problem you commented:
> I was testing the branch and I found thath if you add a new host with the property `run_as: true` in the wazuh.yml and then you enter with the user admin you have no the Wazuh rol `administrator` assigned.

The authentication error in Sample Data could be caused, when add/remove sample data indices because this action uses the internal user configurated in Kibana to do the request to Elasticsearch.

Maybe, it would have to use, the current user with his Elasticsearch permissions instead.

server/controllers/wazuh-api.js
Desvelao: I extracted the code to a function to be reusable

server/controllers/wazuh-hosts.js
Desvelao: I removed the validation, because, the refactorized function is doing the same that here.

server/controllers/wazuh-hosts.js
Desvelao: Now, it is using the refactorized function instead repeat the same code.

frankeros (approved):
LGTM!

jsanchez91 (changes requested):
I have tested the next use case:

- Wazuh API user: `allow_run_as: false`
- Wazuh.yml setting: `run_as: true`

Pude acceder a la interfaz porque recibí el token del usuario configurado. El run_as no se ejecuta en este caso.

El comportamiento deseado es: Indicar al usuario que está intentando usar "run_as" con un usuario sin privilegios y dirigirlo a la vista de "Configuración de la API".

jsanchez91 (approved):
LGTM!

PR #2631, opened 2020-11-17
Removed agent substraction from code

### LINKED ISSUE #2579: [RBAC] Issues detected when using RBAC
Hi team,

Recently we have had three problems using the APP with Wazuh version 4.0. The first of these problems is related to the dev-tools. The endpoint `/security/users/me` is not working properly. It always returns all system permissions. For example, in this screenshot we only have permission to see the agents that belong to the group US_WEST:

We see that the permissions have been correctly applied as only agent 002 belongs to this group. However, when we make a request with the endpoint `/security/users/me` all the permissions are returned to us:

Note: This error may occur with more security endpoints.

The second problem is related to the overview of the Wazuh APP. Despite having the permission to see the agents of the group US_WEST, we see that we have 0 agents. Investigating the requests that are made to obtain this information, we see that it is called the endpoint `/agents/summary/status` which returns a correct output taking into account our permissions: `{active: 1, disconnected: 0, never_connected: 0, pending: 0, total: 1}`

The third problem is related to the 'Agents' tab, when we click on it no agent appears. In order to obtain this information, the endpoint `/agents/000/config/request/remote` is used. However, our user does not have permissions on agent 000 as it does not belong to the US_WEST group. This causes users to always have to add agent 000 to the permissions.

These are the permissions we have used to perform these tests:

Regards

frankeros:
Hi team! This resolves:

- Remove all default agent subtraction from total and active 

Closes #2579

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ All notable changes to the Wazuh app project will be documented in this file.
 
 - [4.0-7.9] Run as with no wazuh-wui API user [#2576](https://github.com/wazuh/wazuh-kibana-app/issues/2576)
 
+### Fixed
+
+- [RBAC] Issues detected when using RBAC [#2579](https://github.com/wazuh/wazuh-kibana-app/issues/2579)
+
 ## Wazuh v4.0.1 - Kibana v7.9.1, v7.9.3 - Revision 4009
 
 ### Changed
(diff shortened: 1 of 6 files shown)


Desvelao (approved):
LGTM!

jsanchez91 (approved):
LGTM!

PR #2633, opened 2020-11-17
[BUGFIX] Default user field for current platform

frankeros:
Hi team! This improves

-  Rule editor on Security/Role mapping should add internal users and custom rules with the `user_field` according to the current platform 
Opendistro => `user_name`
X-Pack or defautl => `username`
- Removed unnecessary prompt for security plugin 
Now if there is no security plugin the internal users should show an empty list.

image


public/components/security/roles-mapping/roles-mapping.tsx
```diff
@@ -18,8 +18,14 @@ import { Rule } from '../rules/types/rule.type';
 import { Role } from '../roles/types/role.type';
 import RolesServices from '../roles/services';
 import RulesServices from '../rules/services';
+import { compose } from 'redux';
+import { connect } from 'react-redux';
 
-export const RolesMapping = () => {
+const mapStateToProps = state => ({ currentPlatform: state.appStateReducers.currentPlatform });
```
Desvelao, line 24: To keep the component as a function, you could replace the HOC with the `useSelector` hook from 'react-redux'

```
import { useSelector, useDispatch } from 'react-redux';

export const RolesMapping  = () => {
  // Extract the variable form Redux store
  const currentPlatform= useSelector(state => state.appStateReducers.currentPlatform);
// ... code
}

jsanchez91 (approved):
LGTM!

PR #2647, opened 2020-11-20
feat(frontend): increase the API entities reserved IDs

### LINKED ISSUE #2643: Problem in the visualization of new reserved resources of the Wazuh API
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.1   | 7.9.1 (Open Distro 1.11) | 4009 |

**Description**
In Wazuh Cloud we have created reserved users and rules in the Wazuh API, but the APP does not reflect this. As @davidjiglesias 
 says, it is necessary to make any API resource with an ID between 1 and 99 reserved, that is, to appear as reserved and without giving the option to delete.

I believe that the option to increase the range (1 to 99) in future versions is contemplated.

**Steps to reproduce**

To be able to reproduce it, it is necessary to add a resource with an ID between 1 and 99. 

**Screenshots**

image

image

Desvelao:
Hi team, this PR resolves: 
 - Created a method to check if an API entity is reserved (ID < 100)

closes #2643


frankeros (approved):
LGTM!

jsanchez91 (approved):
LGTM!

PR #2655, opened 2020-11-24
fix(fronted): fixed the check of API and APP version in health check

Desvelao:
Hi team,

This PR resolves:
- Fix the check of API and APP version in the health check

image


jsanchez91 (approved):
LGTM!

Issue #2639, opened 2020-11-18
Security alerts and reporting issues when using private tenants


Phandora:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0   | 7.9.1     | 4008 |

## Description

When using **private tenants** the `security alerts are not displayed` and `reports can not be generated`.

## Steps to reproduce

- Multitenancy settings in `Kibana.yml`:

```
    opendistro_security.multitenancy.enabled: true
    opendistro_security.multitenancy.tenants.preferred: ["Private", "Global"]
```

1. Go to Wazuh app > Modules > Security events
2. No alerts are shown in `Security Alerts`
3. Click on Generate report
4. See error:

> 5029 - Reporting was aborted (Not Found)

**Screenshots**

Security Alerts:

![check_security_alerts](https://user-images.githubusercontent.com/33484080/99573752-c3d21f80-29d6-11eb-9797-62839cc772ba.png)

Report:

![generate_report](https://user-images.githubusercontent.com/33484080/99573766-c6cd1000-29d6-11eb-90f7-769f524722ac.png)

## Additional context

When performing the checks after accessing Kibana, the index pattern is created into the specific Kibana user index. However, when showing the security alerts or generating reports, the .kibana index is checked in order to get the index pattern.

<details>
<summary>Details</summary>

![create_index_pattern](https://user-images.githubusercontent.com/33484080/99574980-54f5c600-29d8-11eb-9a76-aca8ad4b7699.png)

![image](https://user-images.githubusercontent.com/33484080/99570242-f0d00380-29d1-11eb-92f1-87b4d88bde25.png)

</details>

### Workaround

 Using **global tenants** when accessing Kibana will create the index pattern into the `.kibana` index instead of the one for the specific Kibana user. This way, even if it is switched back to **private tenant**, the `.kibana` index will contain the index pattern.

- Multitenancy settings in `Kibana.yml`:

```
    opendistro_security.multitenancy.enabled: true
    opendistro_security.multitenancy.tenants.preferred: ["Global", "Private"]
```

Best regards.

Desvelao:
The endpoint used to retrieve the alerts for Security Alerts table in Module/Security events is using the internal user configurated in `kibana.yml`, same is occurring with the reports.

We need to fix:
- [x] Modify the endpoint used by the table, to use `callWithRequest` in the backend.
  - [x] Adapt the request to the backend with the query to Elasticsearch
[...]

PR #2663, opened 2020-11-27
[Fix] Discover supports private tenant

Desvelao:
Hi team,

this PR resolves:
- The Discover component uses `/elastic/esAlerts` and supports Private tennant in ODFE. This solves the table in Modules/Security Events.
- Added support when push the Refresh button to refresh the table

It is part of #2639

### DIFF
--- a/public/components/common/hooks/useResfreshAngularDiscover.ts
+++ b/public/components/common/hooks/useResfreshAngularDiscover.ts
@@ -18,7 +18,7 @@ export function useRefreshAngularDiscover(): number {
     let subscription;
     ModulesHelper.getDiscoverScope()
       .then(scope => {
-        subscription = scope.$watchCollection('fetchStatus',
+        subscription = scope.$watch('fetchStatus',
           (fetchStatus) => {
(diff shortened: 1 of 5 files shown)


public/components/common/hooks/useResfreshAngularDiscover.ts
```diff
@@ -0,0 +1,29 @@
...
+import { useState, useEffect } from 'react';
+import { ModulesHelper } from '../modules/modules-helper';
+
+export function useRefreshAngularDiscover(): number {
+  const [refresh, setRefresh] = useState(Date.now());
+  useEffect(() => {
+    let subscription;
+    ModulesHelper.getDiscoverScope()
+      .then(scope => {
+        subscription = scope.$watchCollection('fetchStatus',
```
frankeros, line 21: why did you use $watchCollection instead of $watch? `fetchStatus` is a string

Desvelao: You are right :+1:

Desvelao:
I replaced the `JSON.stringify` method to compare objects by `_isEqual` from `lodash` package.

jsanchez91 (approved):
LGTM!

Issue #2657, opened 2020-11-24
Wrong permissions on edit CBD


gabiwassan:
Parent issue | https://github.com/wazuh/wazuh-kibana-app/issues/2656
------------ | -------------
Branch | [fix/2657-wrong-permissions-on-edit-cdb](https://github.com/wazuh/wazuh-kibana-app/tree/fix/2657-wrong-permissions-on-edit-cdb)

### Description
In kibana configured as cluster, when trying to edit a CDB it asks for `manager: read_file, manager: read, lists: read` permissions which should change dynamically if it is **_Cluster_** or **_Manager_**.
These actions and resources are _**hardcoded**_.

When we try added **list_read(rule:file:*)**, we can read all CDB, and button edit is enabled, but we don't have access to this.
![image](https://user-images.githubusercontent.com/11060558/100134259-b13c6800-2e66-11eb-85ef-05fdf1db50ca.png)

And when we try added **upload_file_test** policies, and now We can press **+ Add new lists file** button:
- We can add a new entry, but when we try to Save we get the same permisions error.
![image](https://user-images.githubusercontent.com/11060558/100134529-1db76700-2e67-11eb-83e1-de3902d2711b.png)

### Possible solution
To the edit button assign so that the following actions and resources can be enabled:

**Actions:** `manager/cluster:read, manager/cluster:read_file, cluster:status`
**Resources:** `file:path:*, *:*:*`

Obtaining dynamically if it is `manager` or `cluster` the required actions.

### Important
To get the configuration context, the `cluster: status` policy is required to access and use this view.

PR #2665, opened 2020-11-27
[Fix] Wrong permissions on edit CDB

gabiwassan:
Hi team,

This PR resolves:
- Wrong permissions on edit CDB List with Cluster and Manager context configuration server.
- Added action to get context server and status.
- Context server is dynamically added to the CDBList edit button, add new, and delete.
Note: Policy `cluster:status` is necessary to get configuration context `manager` or `cluster`

Issue #2657

### DIFF
--- a/public/controllers/management/components/management/ruleset/actions-buttons.js
+++ b/public/controllers/management/components/management/ruleset/actions-buttons.js
@@ -304,7 +304,7 @@ class WzRulesetActionButtons extends Component {
         buttonType="empty"
         permissions={[
           {
-            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}::upload_file`,
+            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:upload_file`,
             resource: `file:path:/etc/${section}`,
           },
           {
--- a/public/controllers/management/components/management/ruleset/actions-buttons.js
+++ b/public/controllers/management/components/management/ruleset/actions-buttons.js
@@ -178,7 +178,12 @@ class WzRulesetActionButtons extends Component {
     // Add new rule button
     const addNewRuleButton = (
       <WzButtonPermissions
-        permissions={[{action: 'manager:upload_file', resource: `file:path:/etc/${section}`}]}
+        permissions={[
+          {
+            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:upload_file`,
+            resource: `file:path:/etc/${section}`,
+          },
+        ]}
         buttonType='empty'
         iconType="plusInCircle"
         onClick={() =>
@@ -197,7 +202,12 @@ class WzRulesetActionButtons extends Component {
     const addNewCdbListButton = (
       <WzButtonPermissions
         buttonType='empty'
-        permissions={[{action: 'manager:upload_file', resource: 'file:path:/etc/lists/files'}]}
+        permissions={[
+          {
+            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:upload_file`,
+            resource: 'file:path:/etc/lists/files',
+          },
+        ]}
         iconType="plusInCircle"
         onClick={() =>
           this.props.updateListContent({
@@ -215,7 +225,12 @@ class WzRulesetActionButtons extends Component {
     const manageFiles = (
       <WzButtonPermissions
         buttonType='empty'
-        permissions={[{action: 'manager:upload_file', resource: `file:path:/etc/${section}`}]}
+        permissions={[
+          {
+            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}::upload_file`,
+            resource: `file:path:/etc/${section}`,
+          },
+        ]}
         iconType={showingFiles ? 'apmTrace' : 'folderClosed'}
         onClick={async () => await this.toggleFiles()}
       >
@@ -252,6 +267,7 @@ class WzRulesetActionButtons extends Component {
         {(section === 'lists' || showingFiles) && (
           <EuiFlexItem grow={false}>
             <UploadFiles
+              clusterStatus={this.props.clusterStatus}
               msg={section}
               path={`etc/${section}`}
               upload={uploadFile}
--- a/public/controllers/management/components/management/ruleset/actions-buttons.js
+++ b/public/controllers/management/components/management/ruleset/actions-buttons.js
@@ -302,24 +302,7 @@ class WzRulesetActionButtons extends Component {
     const manageFiles = (
       <WzButtonPermissions
         buttonType="empty"
-        permissions={[
-          {
-            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:upload_file`,
-            resource: `file:path:/etc/${section}`,
-          },
-          {
-            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:read`,
-            resource: `file:path:/etc/${this.props.msg}`,
-          },
-          {
-            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:read_file`,
-            resource: `file:path:/etc/${this.props.msg}`,
-          },
-          {
-            action: `cluster:status`,
-            resource: `*:*:*`,
-          },
-        ]}
+        permissions={getPermissionsFiles()}
         iconType={showingFiles ? 'apmTrace' : 'folderClosed'}
         onClick={async () => await this.toggleFiles()}
       >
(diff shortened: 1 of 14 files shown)


public/redux/reducers/appStateReducers.js
```diff
@@ -120,6 +120,13 @@ const appStateReducers = (state = initialState, action) => {
     };
   }
 
+  if (action.type === 'UPDATE_CLUSTER_STATUS') {
+    return {
+      ...state,
+      clusterStatus: action.clusterStatus,
```
frankeros, line 126: you are missing add to initial state the prop `clusterStatus`  for this reducer

frankeros (changes requested):

public/redux/reducers/appStateReducers.js
gabiwassan: Thx!!! 😄

public/controllers/management/components/management/ruleset/ruleset-table.js
```diff
@@ -122,14 +125,41 @@ class WzRulesetTable extends Component {
     this._isMounted = false;
   }
 
+  isClusterOrManager = async () => {
+    await clusterReq()
+      .then((clusterStatus) => {
```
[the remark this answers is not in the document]
gabiwassan: Are you sure? I think you cannot use try-catch statements to handle exceptions thrown asynchronously, as the function has "returned" before any exceptions are thrown.

public/controllers/management/components/management/ruleset/actions-buttons.js
```diff
@@ -199,59 +199,79 @@ class WzRulesetActionButtons extends Component {
...
-    const addNewCdbListButton = (
-      <WzButtonPermissions
-        buttonType='empty'
-        permissions={[
-          {
-            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:upload_file`,
-            resource: 'file:path:/etc/lists/files',
-          },
+    const getPermissionsNewFile = () => {
+      let permissions = [
```
frankeros, line 228: `permissions` can be a `const`

public/controllers/management/components/management/ruleset/ruleset-table.js
```diff
@@ -236,29 +236,45 @@ class WzRulesetTable extends Component {
...
-          'data-test-subj': `row-${id || name}`,
-          className: 'customRowClass',
-          onClick: !WzUserPermissions.checkMissingUserPermissions(
-            [
-              {
-                action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:read_file`,
-                resource: `file:path:${item.relative_dirname}/${item.filename}`,
-              },
+        const getRequiredPermissions = (item) => {
+          let permissions = [
```
frankeros, line 240: same here

public/controllers/management/components/management/ruleset/utils/columns.js
```diff
@@ -249,21 +249,39 @@ export default class RulesetColumns {
       };
 
       const getEditButtonPermissions = (item) => {
-        return [
+        let permissions = [
```
frankeros, line 252: same

frankeros (changes requested):

public/controllers/management/components/management/ruleset/actions-buttons.js
gabiwassan: Sure! thanks

public/react-services/wz-user-permissions.ts
```diff
@@ -950,8 +951,32 @@ export class WzUserPermissions{
...
+              ? ![actionResource, actionResourceAll].includes(resource) &&
+                (resource.match(`/${actionResource}/`) || resource.match(`/${actionResourceAll}/`))
+              : ![actionResource, actionResourceAll].includes(resource) &&
+                (resource.match(actionResource.replace('*', '\\*')) ||
+                  resource.match(actionResourceAll.replace('*', '\\*')))
+          )
+        : undefined;
+
+      userPartialResources =
+        (userPartialResources || []).length > 0 ? userPartialResources : undefined;
```
frankeros, line 978: you can save a couple code lines with 
```js
      const userPartialResources: string[] | undefined = Object.keys(userPermissions[actionName] || {}).length
        ? Object.keys(userPermissions[actionName]).filter((resource) =>
            resource.match('&')
              ? ![actionResource, actionResourceAll].includes(resource) &&
                (resource.match(`/${actionResource}/`) || resource.match(`/${actionResourceAll}/`))
              : ![actionResource, actionResourceAll].includes(resource) &&
                (resource.match(actionResource.replace('*', '\\*')) ||
                  resource.match(actionResourceAll.replace('*', '\\*')))
          )
        : undefined;
```

frankeros (changes requested):

public/react-services/wz-user-permissions.ts
gabiwassan: Excellet! let me try and test 😄

jsanchez91 (changes requested):
I have conducted tests with these users and roles using X-Pack Security:
- **elastic**: administrator
- **test_user**: readonly, cluster_readonly

This is the result I have obtained in both tests

image

gabiwassan:
> I have conducted tests with these users and roles using X-Pack Security:
> 
> * **elastic**: administrator
> * **test_user**: readonly, cluster_readonly
> 
> This is the result I have obtained in both tests
> 
> image


Thanks, I'm checking it right now.

frankeros:
Hi @gabiwassan 

#### Testing this branch I found this bug when you set these policies to the role
image

You're able to see all files but trying to access one of them throw this error
image

The same error occurs 
image

----

#### Another bug that I found when you have a specific path in `file:path`
like this 
image

You see it as disabled
image

----

#### Also with the same policies, when you try to add new lists 
image
It requires a specific path (I'm not sure if it is a bug)
Screenshot from 2020-12-10 10-02-26

public/controllers/management/components/management/ruleset/actions-buttons.js
```diff
@@ -214,8 +301,25 @@ class WzRulesetActionButtons extends Component {
...
-        permissions={[{action: 'manager:upload_file', resource: `file:path:/etc/${section}`}]}
+        buttonType="empty"
+        permissions={[
+          {
+            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:upload_file`,
+            resource: `file:path:/etc/${section}`,
+          },
+          {
+            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:read`,
+            resource: `file:path:/etc/${this.props.msg}`,
```
Desvelao, line 312: Should it be the `section` variable instead `this.props.msg`? I checked the code and didn't found where it is coming from

public/controllers/management/components/management/ruleset/actions-buttons.js
```diff
@@ -214,8 +301,25 @@ class WzRulesetActionButtons extends Component {
...
+          },
+          {
+            action: `${((this.props || {}).clusterStatus || {}).contextConfigServer}:read_file`,
+            resource: `file:path:/etc/${this.props.msg}`,
```
Desvelao, line 316: Should it be the `section` variable instead `this.props.msg`? I checked the code and didn't found where it is coming from

Desvelao:
I tested the PR and found the next problem:
- Logged with deafult `admin` user in OpenDistro, I foud this problem in the `Management/Rules` section:
image

Desvelao:
I don't think we need check the permissions here, allowing edit/remove items of the list.
image
image

I think only the `Save button` is enough to do the validation to upload the file.

I found a problem using the default `admin` user that has the default `administrator` role. The `Save button` is disabled because the required `resource` is not included in the `action` that has this role.

image
image

Maybe modify the `resource` to `node:id:*` (Although the file should really have to be uploaded to the master node of the cluster, i mean the resource should be `node:id:<master_node_id>` instead `node:id:*` to be more exactly)

frankeros:
Testing the new commits I found these two bugs

- When the user has permissions to a specific file and permissions to `upload_file`
image
and try to save a new list file gets this error
image
but the new item is created anyway

- When the user (any with access to this action) tries to export a `list` detail, gets this error
image
```
code: 3034
message: "3034 - An error occurred fetching data from the Wazuh API : Extra query parameter(s) path not in spec"
statusCode: 500
```

frankeros (approved):
great job @gabiwassan!! :clap: :clap:

gabiwassan:
> Testing the new commits I found these two bugs
> 
> * When the user has permissions to a specific file and permissions to `upload_file`
>   image
>   and try to save a new list file gets this error
>   image
>   but the new item is created anyway
> * When the user (any with access to this action) tries to export a `list` detail, gets this error
>   image
> 
> ```
> code: 3034
> message: "3034 - An error occurred fetching data from the Wazuh API : Extra query parameter(s) path not in spec"
> statusCode: 500
> ```

The following issue was created https://github.com/wazuh/wazuh-kibana-app/issues/2692 for error 500

Desvelao (approved):
LGTM

Issue #2620, opened 2020-11-13
The index-pattern is not created when it does not have a Wazuh template


SitoRBJ:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0   | 7.9.1     | 4008 |

**Description**
When creating an Open Distro environment without adding the Wazuh template or a Wazuh API, the index pattern is not created. 

**Steps to reproduce**

1. Do not add Wazuh template. Disable the template check. Have a default API in the wazuh.yml
2. Logging in Open Distro Kibana

**Screenshots**

![image](https://user-images.githubusercontent.com/8913336/99096591-c0095c00-25d6-11eb-99b6-ca67d3940e36.png)

**Additional context**

- wazuh.yml
```yaml
---
#
# Wazuh app - App configuration file
# Copyright (C) 2015-2020 Wazuh, Inc.
#
# This program is free software; you can redistribute it and/or modify
# it under the terms of the GNU General Public License as published by
# the Free Software Foundation; either version 2 of the License, or
# (at your option) any later version.
#
# Find more information about this on the LICENSE file.
#
# ======================== Wazuh app configuration file ========================
#
# Please check the documentation for more information on configuration options:
# https://documentation.wazuh.com/current/installation-guide/index.html
#
# Also, you can check our repository:
# https://github.com/wazuh/wazuh-kibana-app
#
# ------------------------------- Index patterns -------------------------------
#
# Default index pattern to use.
pattern: "*:wazuh-alerts"
#
# ----------------------------------- Checks -----------------------------------
#
# Defines which checks must to be consider by the healthcheck
# step once the Wazuh app starts. Values must to be true or false.
#checks.pattern : true
checks.template: false
#checks.api     : true
#checks.setup   : true
#checks.metaFields  : true
#
# --------------------------------- Extensions ---------------------------------
#
# Defines which extensions should be activated when you add a new API entry.
# You can change them after Wazuh app starts.
# Values must to be true or false.
#extensions.pci       : true
#extensions.gdpr      : true
#extensions.hipaa     : true
#extensions.nist      : true
#extensions.tsc       : true
#extensions.audit     : true
#extensions.oscap     : false
#extensions.ciscat    : false
#extensions.aws       : false
#extensions.gcp       : false
#extensions.virustotal: false
#extensions.osquery   : false
#extensions.docker    : false
#
# ---------------------------------- Timeout ----------------------------------
#
[...]

PR #2671, opened 2020-12-02
Create Index Pattern without available indices

pablomarga:
Hi team! This PR fix the problem of creating a new index pattern in the health check when there aren´t any available indices that match with the pattern name.
Fixed [#2620 ](https://github.com/wazuh/wazuh-kibana-app/issues/2620)


jsanchez91 (approved):
LGTM!

PR #2710, opened 2020-12-24
fix route in getCurrentUser when AWS Elasticsearch is used

harduino:
When using AWS Elasticsearch 7.9.1 error "bad request" appearing. All works if "path" in OpendistroFactory.getCurrentUser() starts with "/".
image
image


Desvelao (approved):
Hi @harduino, thank you so much for your collaboration! 😄 

I changed the PR target branch to our current develop branch `4.1-7.10`. 

### Test
I tested the suggested change and it is working with Open Distro without problems too.

We will accept the PR and add a commit to the `CHANGELOG.md` file with the fix.

Thank you so much, again!

PR #2711, opened 2020-12-24
fix getListOfIndexPatterns() when required patterns are not on 1st page

harduino:
Wazuh UI retrieves index patterns and checks that at least one from list with required fields exists. By default in getListOfIndexPatterns() there is no "per page" parameter sending, so only first 20 index patterns are returning even if there are more than 20 index patterns. So in situation with >20 index patterns and when required index patterns are not on 1st page, Wazuh UI shows error "Sorry but no valid index patterns were found" even if valid index patterns are exist.
image
image
image
This PR changes getListOfIndexPatterns() method to retrieve 9999 index-patterns, so required index patterns will be always returned on check by Web UI.

### DIFF
--- a/public/react-services/saved-objects.js
+++ b/public/react-services/saved-objects.js
@@ -22,7 +22,7 @@ export class SavedObject {
     try {
       const result = await GenericRequest.request(
         'GET',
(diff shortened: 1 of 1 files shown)


gabiwassan:
Close: #3223

PR #2714, opened 2020-12-28
Feature/2713 error toast is showing about elasticsearch users for environments without security

### LINKED ISSUE #2713: Error toast is showing about Elasticsearch users for environments without security
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.x   | 7.x     | 400x |

**Description**
For Elasticsearch without security, we are trying to get the Elasticsearch users when it shouldn't for this case.

A possible solution will be, don't the request in this case.

We can know if there security plugin with the `appStateReducers.currentPlatform` variable saved in Redux's store. We can use it, to perform the request or not.

**Screenshots**
image

CPAlejandro:
Solving issue #2713 
For Elasticsearch without security, we are trying to get the Elasticsearch users when it shouldn't for this case.

- We have fixed it by creating a conditional that only if you have the security plugin can make the request
Closes #2713


frankeros (approved):
LGTM!

Desvelao (approved):

gabiwassan (approved):
👍🏼

Issue #2692, opened 2020-12-15
Error 500 on Export formatted CDB list


gabiwassan:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.3   | 7.9.1  | 4014 |

**Description**
Error 500 when trying to Export formatted any CBD list.

**Steps to reproduce**

1. Go to Management CDB List.
2. Select 1 list.
3. Click on the `Export formatted` button.
4. See error:

**Screenshots**
![image](https://user-images.githubusercontent.com/11060558/102217411-4b6a5b80-3ebb-11eb-8480-d8b5376ea905.png)
![image](https://user-images.githubusercontent.com/11060558/102217462-5e7d2b80-3ebb-11eb-95e3-fc9e7a406d3c.png)

PR #2717, opened 2020-12-30
Fixed error due to new use of path in the export of lists

CPAlejandro:
Hi team, we resolve the issue #2692 with this steps:
- We change `path` to `relative_dirname` and `filename` (which is the new way to use it) and instead of concatenating name, we pass another filter passing this name to filename

### DIFF
--- a/public/controllers/management/components/management/ruleset/list-editor.js
+++ b/public/controllers/management/components/management/ruleset/list-editor.js
@@ -593,12 +593,17 @@ class WzListEditor extends Component {
                         try {
                           this.setState({ generatingCsv: true });
                           await exportCsv(
-                            `/lists?path=${path}/${name}`,
+                            `/lists`,
                             [
                               {
                                 _isCDBList: true,
-                                name: 'path',
-                                value: `${path}/${name}`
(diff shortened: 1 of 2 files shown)


public/controllers/management/components/management/ruleset/list-editor.js
```diff
@@ -593,12 +593,17 @@ class WzListEditor extends Component {
...
                           await exportCsv(
-                            `/lists?path=${path}/${name}`,
+                            `/lists`,
                             [
                               {
                                 _isCDBList: true,
-                                name: 'path',
-                                value: `${path}/${name}`
+                                name: 'relative_dirname',
+                                value: `${path}`
```
Desvelao, line 601: If you don't need to interpolate any variable into the string, you can set the variable as `value` object key.

CPAlejandro: Well, I have applied your advice. Thank you!

Desvelao (approved):
LGTM!

frankeros (approved):
LGTM!

PR #2749, opened 2021-01-07
Fix ApiInterceptor.buildRequestOptions to destructurate data parameters

Desvelao:
Hi team, this PR resolves:
- Desestrucutrate the data parameters received by the `buildRequestOptions` method of the ApiInterceptor into:
  - `body`
  - `headers`
  - `params` 
  - `rest`
- Use the `rest` if `body` property doesn't exist in the `data` parameter

This should solve the problem with some requests in the `Security` app section.


frankeros (approved):
LGTM!

PR #2769, opened 2021-01-12
Changed the way index pattern are refreshed

pablomarga:
Hi teams this PR solves a problem that was caused due to before to refresh the fields of the index pattern we used an endpoint that was removed.

### DIFF
--- a/public/react-services/pattern-handler.js
+++ b/public/react-services/pattern-handler.js
@@ -111,12 +111,11 @@ export class PatternHandler {
    */
   static async changePattern(selectedPattern) {
     try {
-      let prueba = AppState.getCurrentPattern();
       AppState.setCurrentPattern(selectedPattern);
       await this.refreshIndexPattern();
       return AppState.getCurrentPattern();
     } catch (error) {
-      throw new Error('Error Pattern Handler (changePattern)',error);
+      throw new Error('Error Pattern Handler (changePattern)');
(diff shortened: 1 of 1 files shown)


public/react-services/pattern-handler.js
```diff
@@ -111,15 +111,12 @@ export class PatternHandler {
    */
   static async changePattern(selectedPattern) {
     try {
+      let prueba = AppState.getCurrentPattern();
```
frankeros, line 114: remove this

public/react-services/pattern-handler.js
```diff
@@ -111,15 +111,12 @@ export class PatternHandler {
...
-      await GenericRequest.request(
-        'GET',
-        `/elastic/known-fields/${selectedPattern}`,
-        {}
-      );
+      await this.refreshIndexPattern();
       return AppState.getCurrentPattern();
     } catch (error) {
-      throw new Error('Error Pattern Handler (changePattern)');
+      throw new Error('Error Pattern Handler (changePattern)',error);
```
frankeros, line 119: new Error only gets one parameter for the message.

frankeros (changes requested):

Desvelao (approved):
LGTM!

frankeros (approved):
LGTM!

PR #2779, opened 2021-01-13
Fixed wrong shards and replicas for statistics indices. Fixed wrong prefix for monitoring indices

### LINKED ISSUE #2732: The wazuh-statistics index shards do not match the cron.statistics.shards value
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.3  | 7.9.1    | 4015|

## Description

The `wazuh-statistics` index is created with 2 shards regardless of the `cron.statistics.shards` setting value.

## Steps to reproduce

1. Use the following setting in the _wazuh.yml_ file:

```
cron.statistics.index.creation: m
cron.statistics.shards: 1
...
```

2. Check the wazuh-statistics index:

```bash
green open wazuh-statistics-2021.01                aeStzeKiSCaMKAfLQU8kqw 2 0  2  0  22.4kb  22.4kb
```

The index is created monthly but with 2 primary shards instead of 1, and 0 replicas (default value)

## Additional context

After deleting the `wazuh-statistics-2021.01` index, it is recreated again with the same number of primary shards.

pablomarga:
Hi team, this PR solves two bugs when creating new indices. First one was that when it created new indices dindn´t match with the cron.statistics.shards configuration. The second one was that when the creation period was modified for the monitoring indices it changed the prefix name of the indices

Closes #2732 .


server/lib/cron-scheduler/save-document.ts
```diff
@@ -59,7 +58,9 @@ export class SaveDocument {
             }
           }
         });
-        
+        }
+      }catch(error){
+        log(this.logPath, `Index '${index}' exists? false`, 'debug');
         log(this.logPath, `Status of create a new index: ${JSON.stringify(response)}`, 'debug');
```
Desvelao, line 64: If the index pattern doesn't exist, the request doesn't throw an error, so it doesn't enter in the `catch` block.

The `response` variable is decalred in the `try` block. Should you move the log there.

server/lib/cron-scheduler/save-document.ts
```diff
@@ -43,25 +43,25 @@ export class SaveDocument {
...
             }
-          }
-        });
+          });
         }
-      }catch(error){
-        log(this.logPath, `Index '${index}' exists? false`, 'debug');
         log(this.logPath, `Status of create a new index: ${JSON.stringify(response)}`, 'debug');
+      } catch (error) {
+        log(this.logPath, `Index '${index}' exists? false`, 'debug');
```
Desvelao, line 64: Could you log the error in the `catch` block?

Desvelao (approved):
LGTM!

frankeros (approved):
LGTM!

PR #2780, opened 2021-01-13
Check monitoring and statistics index patterns

frankeros:
Hi team, this resolves
 - Check if monitoring and statistics index pattern already exists, if not then create them
 - Also covering the creation of index patterns using tenants with ODFE
 
 Closes https://github.com/wazuh/wazuh-kibana-app/issues/2753#issuecomment-758649742

image
image


Desvelao:
I tested the PR using X-Pack with a custom space and user and it works. Good job! 👏 

I think it would be a good idea to show a toast indicating that the index pattern goes to be created and/or is created successfully like the toast for `wazuh-alerts-*` index pattern. Add a toast if an error occurs.

server/lib/cron-scheduler/save-document.ts
```diff
@@ -31,7 +31,7 @@ export class SaveDocument {
       const createDocumentObject = this.createDocument(doc, indexCreation, mapping);
       const response = await this.esClientInternalUser.bulk(createDocumentObject);
       log(this.logPath, `Response of create new document ${JSON.stringify(response)}`, 'debug');
-      // await this.checkIndexPatternAndCreateIfNotExists(index);
+      await this.checkIndexPatternAndCreateIfNotExists(index);
```
Desvelao, line 34: This creates an index pattern through the backend. We want to create it in the frontend.

server/lib/cron-scheduler/save-document.ts
```diff
@@ -40,13 +40,60 @@ export class SaveDocument {
...
+      });
+      log(
+        this.logPath,
+        `The indexPattern no exist, response of createIndexPattern: ${JSON.stringify(response)}`,
+        'debug'
+      );
+    } catch (error) {
+      this.checkDuplicateIndexPatterError(error);
+    }
+  }
```
Desvelao, line 84: This logic creates an index pattern in the backend. We create it in the frontend. You could remove these methods too.

gabiwassan (approved):
LGTM!

public/components/health-check/health-check.tsx
```diff
@@ -247,6 +239,40 @@ export class HealthCheck extends Component {
...
+                getToasts().remove(toast.id);
+                getToasts().addSuccess(`${pattern} index pattern created successfully`)
+                results[itemId].description = <span><EuiIcon type="check" color="secondary" ></EuiIcon> Ready</span>;                
+            } catch (error) {
+                getToasts().remove(toast.id);
+                errors.push(`Error trying to create ${pattern} index pattern: ${error.message}`);
+                results[itemId].description = <span><EuiIcon type="alert" color="danger" ></EuiIcon> Error</span>;
+            }
+        }
+    }
```
Desvelao, line 274: If you mute the `results` variable that is in the component's state. You should set the state after the mutation, like this:

```
this.setState( { results } );
```

You could do this after the `try/catch` block.

pablomarga:
LGTM! Great job!

pablomarga (approved):
LGTM! Great job!

public/components/health-check/health-check.tsx
frankeros: I missed that thanks

Desvelao (approved):
LGTM!

PR #2784, opened 2021-01-15
Fix the report's creation date

### LINKED ISSUE #2772: Report's creation dates set to 1970-01-01T00:00:00.000Z
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.3  | 7.9.1    | 4015|

## Description

When checking the generated reports in the `Management > Reporting` tab, the report's creation dates are set to **1970-01-01T00:00:00.000Z**.

The dates do not match the ones showed in the `/usr/share/kibana/optimize/wazuh/downloads/reports` directory. 

## Steps to reproduce

1. Generate report.
2. Go to `Management > Reporting`
3. Check the report's creation dates

## Screenshots

- **Response using the Developer Tools**:

image

- **Content of the `/usr/share/kibana/optimize/wazuh/downloads/reports` path**:

image

## Additional context

OS: `Oracle Linux 7.9`

> On filesystems where birthtime is not available, this field may instead hold either the ctime or 1970-01-01T00:00Z (ie, Unix epoch timestamp 0). This value may be greater than atime or mtime in this case. On Darwin and other FreeBSD variants, also set if the atime is explicitly set to an earlier value than the current birthtime using the utimes(2) system call. https://nodejs.org/api/fs.html#fs_stats_birthtimems

Desvelao:
Hi team, this PR resolves:
- Fix the report's creation date showing 1970-01-01T00:00:00.000Z

Some OSs, like Oracle Linux 7.9, have the problem where the file's `birthtime` stat is  `1970-01-01T00:00:00.000Z`.

With these PR, returns the first date whose value is truthy using the `[fieldName]Ms` with the next stat preference:
`birthtime` > `mtime` > `ctime` > `atime`

Closes #2772


gabiwassan (approved):
LGTM!

PR #2806, opened 2021-01-21
[4.0-7.10] Removed kibana alert for security

frankeros:
Hi team this resolves 
- Remove Kibana alert for security https://github.com/wazuh/wazuh-kibana-app/issues/2801#issuecomment-764840354

### DIFF
--- a/public/plugin.ts
+++ b/public/plugin.ts
@@ -65,7 +65,8 @@ export class WazuhPlugin implements Plugin<WazuhSetup, WazuhStart, WazuhSetupPlu
 
   public async start(core: CoreStart, plugins: AppPluginStartDependencies): Promise<WazuhStart> {
     // hide security alert 
-    plugins.security_oss.insecureCluster.hideAlert(true);
+    if(plugins.security_oss)
+      plugins.security_oss.insecureCluster.hideAlert(true);
     
     // we need to register the application service at setup, but to render it
     // there are some start dependencies necessary, for this reason
--- a/public/plugin.ts
+++ b/public/plugin.ts
@@ -65,8 +65,9 @@ export class WazuhPlugin implements Plugin<WazuhSetup, WazuhStart, WazuhSetupPlu
 
   public async start(core: CoreStart, plugins: AppPluginStartDependencies): Promise<WazuhStart> {
     // hide security alert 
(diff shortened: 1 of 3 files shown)


public/plugin.ts
```diff
@@ -64,6 +64,10 @@ export class WazuhPlugin implements Plugin<WazuhSetup, WazuhStart, WazuhSetupPlu
   }
 
   public async start(core: CoreStart, plugins: AppPluginStartDependencies): Promise<WazuhStart> {
+    // hide security alert 
+    if(plugins.security_oss)
+      plugins.security_oss.insecureCluster.hideAlert(true);
```
gabiwassan, line 69: Thi's it only for cluster or it's generic?

gabiwassan (approved):
LGTM!

Desvelao:
I tested this PR with an Elasticsearch without security and the toast about the sure your data is showing.

I researched about this, and the `security_oss` plugin is not included in the Wazuh plugin `kibana.json`. In addition, its plugin ID is `securityOss`.

public/plugin.ts
Desvelao, line 69: It would be `plugins.securityOss` instead `plugins.security_oss`. 

As a personal note, I don't like the conditional inline statement without curly braces.

public/types.ts
```diff
@@ -4,14 +4,16 @@ import { DiscoverStart } from '../../../src/plugins/discover/public';
...
+import { UiActionsSetup } from '../../../src/plugins/ui_actions/public';
+import { SecurityOssPluginStart } from '../../../src/plugins/security_oss/public/';
 
 export interface AppPluginStartDependencies {
   navigation: NavigationPublicPluginStart;
   data: DataPublicPluginStart;
   visualizations: VisualizationsStart;
   discover: DiscoverStart;
   charts: ChartsPluginStart
+  security_oss: SecurityOssPluginStart
```
Desvelao, line 16: It would be `securityOss`

public/plugin.ts
frankeros: It is the only way to manage the events of this alert.

Desvelao (approved):
LGTM!

Issue #2839, opened 2021-01-25
Sample data doesn't use the current manager or cluster name.


jctello:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.4   | 7.10.0     | 4016 |

**Description**
When sample data is added the value of the `manager.name` field is `master` and `cluster.name` is `wazuh`. 
Given that the Wazuh Kibana Plugin filters by its current manager name or, when enabled, the cluster name, this makes the data invisible (except through the Discover app).
 

**Steps to reproduce**

1. Install the wazuh manager on a computer whose hostname is not master (in my screenshot dev410)
2. Add Sample Data by going into Settings > Sample Data in the drop-down menu.
3. Observe there is no new data visible in the Wazuh plugin dashboard (although there is a warning of sample data being present, the amount of data does not significantly increase).
4. Select the Discover app from the left-most menu and find the data. 

**Screenshots**
![image](https://user-images.githubusercontent.com/36071202/105763035-88fbe300-5f55-11eb-94fa-13757e0062c5.png)

![image](https://user-images.githubusercontent.com/36071202/105762766-21de2e80-5f55-11eb-9f30-854d40c2cef2.png)

**Additional context**
This is not observed in the plugin for 4.0.3 - 7.9.1 Revision 4013

PR #2857, opened 2021-01-27
Fix body-payload of Sample Alerts POST endpoint

asteriscos:
Hi team, this PR fixes:

**Sample data doesn't use the current manager or cluster name.**
Turns out the endpoint wasn't receiving the payload to use the current manager, but the feature was already developed.

Closes: #2839


frankeros (approved):
LGTM!

Issue #2855, opened 2021-01-27
Debug information in Dev Tools is not precise enough


Rebits:
Hello team,

| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.4   | 7.10.0     | 4016|

## Description

`Dev Tool` debug information does not inform correctly where the wrong syntax is.

## Steps to reproduce

1. Go to Dev Tools 

2. Perform a malformed query like `GEET /agents`

3. Debug is not enough precise, error line is `1` not `2`

![devtools_bug](https://user-images.githubusercontent.com/11089305/106019550-0ee96c80-60c3-11eb-87c0-7f7e59064cde.png)

**Additional context**
This happens on Firefox and Chrome in Linux and Windows OS

PR #2863, opened 2021-01-28
Fix debug error line number to match code editor line

asteriscos:
Hi team, this PR fixes:

**Debug info mismatch to code editor line number.**
The line number of the error did not count but the line of the error itself.

Closes: #2855


frankeros (commented):
I tested it but I got the same result 
image

But it happens because the editor takes the first query `GET /agents` as wrong, it's taking `GEET` as a continuation of the first query. 
I'm not sure if it's a bug.

Desvelao:
I tested and get the same result as @frankeros. It should be considered a bug.

Issue #2801, opened 2021-01-21
Testing 4.0.4-7.10.0


Desvelao:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.4   | 7.10.0     | 4016|

**Description**
Testing.

Test with browsers:
- Chrome
- Firefox
- Safari

Test with the theme:
- Light
- Dark

Checks:
- [ ] General checks
- [ ] ODFE 1.12.0
- [ ] X-Pack

Desvelao:
## Chrome testing

#### Filter checks (Modules)    

- [ ] "Module/Dashboard -> Add `rule.level:7` -> Go to Module/Events" Filters shouldn't change
- [ ] "Module/Dashboard -> Add `rule.level:7` and make it pinned -> Go to Module/Events" Filters shouldn't change
- [ ] "Module/Dashboard -> Add `rule.level:7` -> Go to Module/FIM/Dashboard" The filter "rule.level:7" is removed
- [ ] "Module/Dashboard -> Add `rule.level:7` and make it pinned -> Go to Module/FIM/Dashboard" The filter "rule.level:7" persists
- [ ] "Module/Events -> Add `rule.level:7` -> Go to Module/FIM/Dashboard" The filter "rule.level:7" is removed
- [ ] "Module/Events -> Add `rule.level:7` and make it pinned -> Go to Module/FIM/Dashboard" The filter "rule.level:7" persists
- [ ] "Module/Events -> Add `rule.level:7` -> Go to Module/Dashboard" Filters shouldn't change
- [ ] "Module/Events -> Add `rule.level:7` and make it pinned -> Go to Module/Dashboard" Filters shouldn't change
- [ ] "Module/Dashboard -> Add `rule.level:7` -> Go to Agent any Module/Dashboard" The filter "rule.level:7" is removed
- [ ] "Module/Dashboard -> Add `rule.level:7` and make it pinned -> Go to Agent any Module/Dashboard" The filter "rule.level:7" persists
- [ ] "Agent any Module/Dashboard -> Add `rule.level:7` -> Go to Agent any Module/Events" Filters shouldn't change
- [ ] "Agent any Module/Dashboard -> Add `rule.level:7` and make it pinned -> Go to Agent any Module/Events" Filters shouldn't change
- [ ] "Agent any Module/Dashboard -> Add `rule.level:7` -> Go to Agent FIM/Dashboard" The filter "rule.level:7" is removed
[...]

PR #2864, opened 2021-01-28
[4.0.4-7.10] Fixed filter pinned crash returning from agents

frankeros:
Hi team, this resolves

 -  Blank page on modules after pinning a filter, go agents and return.
 
Resolves [#2801 comment](https://github.com/wazuh/wazuh-kibana-app/issues/2801#issuecomment-769171853)

### DIFF
--- a/public/components/add-modules-data/sample-data.tsx
+++ b/public/components/add-modules-data/sample-data.tsx
@@ -123,7 +123,7 @@ export default class WzSampleData extends Component {
         }
       });
       await WzRequest.genericReq('POST', `/elastic/samplealerts/${category.categorySampleAlertsIndex}`, { params: this.generateAlertsParams });
-      this.showToast('success', `${category.title} alerts installed`, 'Date range for sample data is now-7 days ago', 5000);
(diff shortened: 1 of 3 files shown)


gabiwassan (approved):
LGTM!

Desvelao:
I tested the PR and the problem seems to continue to occur when the filter is pinned.

The selected tab for the module is set from `tabView` URL quey parameter. For some reason, the `tabView` query parameter is not present in the URL when there a filter pinned.

`public/components/common/modules/main-overview.tsx`

```ts
async componentDidMount() {
    const tabView = AppNavigate.getUrlParameter('tabView');
    const tab = AppNavigate.getUrlParameter('tab');
    if(tabView && tabView !== this.props.selectView){
      if(tabView === 'panels' && tab=== 'sca' ){ // SCA initial tab is inventory
        this.props.onSelectedTabChanged('inventory');
      }else{
        this.props.onSelectedTabChanged(tabView);
      }
    }
    
    this.setGlobalBreadcrumb();
  }
```

Maybe we could set the default tab if `tabView` URL query param is not present?

Desvelao (approved):
I tested the PR after the changes and it is working now. Good job!

PR #2876, opened 2021-02-01
Feature/add warning when configuring new role mapping

### LINKED ISSUE #2851: Feature request: Warn users when configuring role mapping if run_as is set to false
Since `run_as` is disabled by default in the `wazuh.yml`, users may configure RBAC using the Wazuh Kibana plugin without knowing that their changes will not take effect until the configuration file is edited. I suggest that if `run_as` is set to false, a message is shown after applying a role mapping that states:

`
"These changes will not take effect until run_as is activated in /usr/share/kibana/data/wazuh/config/wazuh.yml and the Kibana service is restarted".`

pablomarga:
Hi team this PR adds a toast warning the user when configuring role mapping if run_as is set to false. 

Captura de pantalla 2021-02-01 a las 18 00 07

Closes #2851

### DIFF
--- a/public/components/security/main.tsx
+++ b/public/components/security/main.tsx
@@ -68,6 +68,7 @@ export const WzSecurity = compose(
 
   const checkRunAsUser = async () => {
     const currentApi = AppState.getCurrentAPI();
+    console.log("HIOLOA")
     try {
       const ApiCheck = await GenericRequest.request('POST',
         '/api/check-api',
@@ -83,7 +84,7 @@ export const WzSecurity = compose(
   useEffect(() => {
     checkRunAsUser()
       .then(result => setAllowRunAs(result))
-      .catch(error => console.log(error,'Error checking if run_as user is enabled'))
+      .catch(error => console.log(error, 'Error checking if run_as user is enabled'))
   }, [])
 
   // This allows to redirect to a Security tab if you click a Security link in menu when you're already in a Security section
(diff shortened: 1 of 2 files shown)


public/components/security/roles-mapping/components/rule-editor.tsx
```diff
@@ -227,7 +253,11 @@ export const RuleEditor = ({ save, initialRule, isLoading, isReserved, internalU
     }
   };
 
-  const saveRule = () => {
+  const saveRule = async () => {
+    const isRunAS = await checkRunAsUser();
+    if (!isRunAS.data.allow_run_as) {
+      showToast('warning', 'These changes will not take effect until run_as is activated in /usr/share/kibana/data/wazuh/config/wazuh.yml and the Kibana service is restarted');
```
frankeros, line 259: the path could come from a constant, and you should add " and the Wazuh user has the `allow_run_as` as true

frankeros (changes requested):

public/components/security/main.tsx
```diff
@@ -55,11 +60,33 @@ export const WzSecurity = compose(
...
+
+    } catch (error) {
+      ErrorHandler.handle(error, 'Error checking the current API');
+    }
+  }
+  const [allowRunAs, setAllowRunAs] = useState();
+  useEffect(() => {
+    checkRunAsUser()
+      .then(result => setAllowRunAs(result))
+      .catch(error => console.log('Hubo un error'))
```
Desvelao, line 86: Remove the log in Spanish and manage the error otherwise or do nothing in the `.catch`.

Desvelao:
I tested the PR and I found the callout is showing in all `Security` section tabs, like this:
image

Do we want to show this message in all tabs of the Security section or it should only appear in `Roles mapping` tab?

frankeros:
I tested it, works well but when you have the setting `run_as` as `true` the callout is displayed for a while and then it disappears.
It looks like the callout is displayed by default until the request for `allowedRunAs` is got.

public/components/security/main.tsx
```diff
@@ -103,10 +145,13 @@ export const WzSecurity = compose(
             <Policies></Policies>
           }
           {selectedTabId === 'roleMapping' &&
-            <RolesMapping></RolesMapping>
+            <>
+              {allowRunAs !== API_USER_STATUS_RUN_AS.ENABLED && isNotRunAs()}
```
frankeros, line 149: you could add another state to wait for the request to complete, maybe
```
{allowRunAs !== API_USER_STATUS_RUN_AS.ENABLED && !allowRunAsIsLoading && isNotRunAs()}
```

frankeros (approved):
LGTM!

Issue #2888, opened 2021-02-02
When clicking on a decoder element its information is not displayed


jmv74211:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.0   | 7.10 | 4101-1 |

Hello team,

When clicking in one element of the decoder list to see its details, the info does not appear.

https://user-images.githubusercontent.com/23462183/106595232-87cd4600-6553-11eb-85bd-d15113f8e454.mp4

![image](https://user-images.githubusercontent.com/23462183/106595201-7edc7480-6553-11eb-9294-ae0167e042fe.png)

It is requested to solve this error.

Best regards.

Desvelao:
The view is crashing due to some details key were of `object` type and these weren't managed.

Fixed in the PR: https://github.com/wazuh/wazuh-kibana-app/pull/2903

PR #2903, opened 2021-02-02
[4.1-7.10] Fix the decoder details for keys as objects

Desvelao:
Hi team, this PR resolves:
- Fix the crash in the decoder details view when the property was an object

Closes: #2888


frankeros (commented):
I test it, but when back from the first decoder detail and try to open another it crashes
image

gabiwassan (approved):
LGTM!

frankeros (approved):
I'm tested and everything seems ok

PR #2931, opened 2021-02-03
Check user permissions before render

### LINKED ISSUE #2919: An unauthorized user can retrieve agents groups information
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.0  | 7.10.0     | 4101-1 |

**Description**

An unauthorized user can retrieve agents groups information, which is also shown momentarily when refreshing the groups' page before saying is forbidden.

Two issues can be seen here:
- The user is not authorized to read agents groups but can retrieve groups' information using `GET /groups` in Dev Tools.  
- When refreshing the page `Managements > Groups` the groups' information is shown momentarily before saying "You have no permissions". This behaviour is also observed in other sections like CDB lists. 

output

gabiwassan:
Hi team,
This PR added a control before check permissions to render components and changed the initial state of userPermissions.
+ Applied prettier.

Closes #2919

### DIFF
--- a/public/components/common/hooks/useUserPermissions.ts
+++ b/public/components/common/hooks/useUserPermissions.ts
@@ -26,7 +26,7 @@ export const useUserPermissionsRequirements = (requiredPermissions) => {
   }
 
   if (!userPermissions) {
-    return requiredPermissions;
+    return [requiredPermissions, []];
   }
 
   const requiredPermissionsArray =
--- a/public/components/common/hooks/useUserPermissions.ts
+++ b/public/components/common/hooks/useUserPermissions.ts
@@ -15,25 +15,35 @@ import { WzUserPermissions } from '../../../react-services/wz-user-permissions';
 
 // It retuns user permissions
 export const useUserPermissions = () => {
-  const userPermissions = useSelector(state => state.appStateReducers.userPermissions);
-  return userPermissions;
-}
+  return useSelector((state) => state.appStateReducers.userPermissions);
+};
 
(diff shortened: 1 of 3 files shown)


public/components/common/hooks/useUserPermissions.ts
```diff
@@ -15,25 +15,35 @@ import { WzUserPermissions } from '../../../react-services/wz-user-permissions';
...
+  if (!userPermissions) {
+    return requiredPermissions;
+  }
+
+  const requiredPermissionsArray =
+    typeof requiredPermissions === 'function' ? requiredPermissions() : requiredPermissions;
+  return [
+    WzUserPermissions.checkMissingUserPermissions(requiredPermissionsArray, userPermissions),
+    userPermissions,
+  ];
```
Desvelao, line 37: This hook is returning the next interface:
```
[requiredPermissions, userPermissions]
```

but in the conditional is returned
```
requiredPermissions
```
The return in the conditional should be consistent to the interface

gabiwassan: If it is correct, this variable `requiredPermissions` comes by parameters an array, for ex: 
`[{action: 'agent:read', resource: 'agent:id:*'}]`

gabiwassan: Thanks! this is the correct return `[requiredPermissions, []]`

frankeros (approved):
I tested it and it seems ok!

PR #2932, opened 2021-02-03
Fix clear visualizations list when changing tabs

### LINKED ISSUE #2874: Wrong filters shown in PDF report
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.4   | 7.10.0     | 4017 |

**Description**
When a filter is added on a dashboard and then removed before generating a report, the report will some times still wrongly reflect the filters is being applied.
Sometimes even a newer filter will not be reflected on the report.

**Steps to reproduce**

1. Go to Security Events
2. Click on any value on a visualization to create a filter
3. Go to Events 
4. Remove the filter
5. Go back to Dashboard and remove the filter
6. Click on Generate Report
7. Navigate to the report and see the filters shown on the PDF.
8. Note how the graphs shown do not reflect the previously applied filter (if a specific rule level was selected, all levels are still shown).

**Screenshots**
BadFiltersOnKibana Report

**Additional context**
The steps shown above are not the only way to achieve this, but they reflect a reliable way to reproduce it.
This issue also affects previous versions of the Wazuh Kibana Plugin.

asteriscos:
Hi team, this fixes:

**Filters applied to PDF reports.** 
Clears the visualizations list when changing tabs so as to check which filter is applied to the current chart list.

closes #2874

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 ### Fixed
 
 - Fix server error Invalid token specified: Cannot read property 'replace' of undefined [#2899](https://github.com/wazuh/wazuh-kibana-app/issues/2899)
+- Fixed clear visualizations manager list when switching tabs. Fixes PDF reports filters [#2932](https://github.com/wazuh/wazuh-kibana-app/pull/2932)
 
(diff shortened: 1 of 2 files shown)


gabiwassan (approved):
LGTM Good job!

Desvelao (approved):
LGTM!

Issue #2895, opened 2021-02-02
Mitre detailed view: filters don't work at all


SpotHound:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.0  | 7.10.0     | 4101-1 |

**Description**
At Mitre ATT&ACK detailed view there is an option to set filters but, if you try to add one filter it doesn't' work. Watch the attached video.

https://user-images.githubusercontent.com/15269938/106600448-bc90cb80-655a-11eb-8fd4-56692fe2f9c0.mp4

**Steps to reproduce**

1. Go to Modules/Mitre ATT&AK
2. Select a technique
3. Add a filter

**Extra info**
This happens in all OS and all browsers where the mitre view works.

PR #2935, opened 2021-02-04
Fixed bug in the filters of flyouts

CPAlejandro:
Hello team, we have resolved a bug that did not allow us to use filters in both Miter flyouts and Regulatory Compliance tables.
Closes. #2895.


gabiwassan (approved):
LGTM! 
Checked works properly

public/components/common/modules/discover/discover.tsx
```diff
@@ -108,7 +108,7 @@ export const Discover = compose(
   constructor(props) {
     super(props);
     this.KibanaServices = getDataPlugin();
-    this.filterManager = props.shareFilterManager ? this.KibanaServices.query.filterManager : new FilterManager(getUiSettings);
+    this.filterManager = props.shareFilterManager ? this.KibanaServices.query.filterManager : this.KibanaServices.query.filterManager;
```
Desvelao, line 111: Here, `this.filterManager` always has the same value, don't?

Why not, simplify to:
```
this.filterManager = this.Kibanaservices.query.filerManager
```

Desvelao:
I tested the PR and it is working!

frankeros (commented):
I tested it the filters are applied but are using the same Filter manager instance that the kibana discover search bar, it is not expected behavior. 
I proceed to change it.

gabiwassan (approved):

Issue #2889, opened 2021-02-02
 Selected agents are not unchecked when switching from Available agents to Current agents in the group in Manage agents option


Rebits:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.0   | 7.10.0     | 4101-1 |

**Description**

Hello team,

When you are managing agents in `Manage agent` of a group menu, if you change from `Available agents` to `Current agents`, selected agents don't unmark and interface become confusing due to the lack information of which set of agents did you selected.

**Steps to reproduce**

- Go to `Management > Groups`
- Select a group
- Select `Manage agents`
- Select a set of agents from `Current agents` and then, select a set of agents from `Available agents`. `Add selected items` and r`emove selected items` options change but selected agents are not unmarked

**Video**

https://user-images.githubusercontent.com/11089305/106595055-463c9b00-6553-11eb-9754-522842b00ef9.mp4

**Additional context**
This happens in all browsers and system

PR #2947, opened 2021-02-08
Bugfix agents are not unchecked

gabiwassan:
Hi team,
This PR fixed when you change selections options of right select or left select the options are not unchecked.
- Changed `selectedOptions` for `options`, because after editing the first option in the forEach, it was ignoring a selected option.

Closes: #2889


frankeros (approved):
LGTM!

PR #2952, opened 2021-02-09
Refactored registry values for FIM

frankeros:
Hi team, this resolves

Currently, the items for `registry_key` and `registry_value` types in FIM are displayed at the same level (table). But the items with the type `registry_value` should be displayed inside the detail of their corresponding `registry_key` item.

image


gabiwassan (approved):
LGTM. Good refactor!

PR #2953, opened 2021-02-09
deleted bucket property from json structure of tagclouds

asteriscos:
Hi team, this fixes:
**Since `Kibana 7.10`  didn't load properly the tagcloud visualization json.**
image

Deleting the `bucket` property of the json definition fixed the issue. _Kibana_ seems to have a changed the inner properties of the `tagcloud visualization`

In addition, deleted the class `kbnTopNavMenu` from the `common.scss`, as it was hiding the visualize section navbar action buttons. This made saving new visualizations impossible.


frankeros (approved):
LGTM! Good catch :tada:

PR #2961, opened 2021-02-11
Bugfix/2887 null pointer in filter wrong type

### LINKED ISSUE #2887: Null pointer exception when creating a filter with wrong type + Duplicated warnings
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.0   | 7.10.0     | 4101-1 |

**Description**

If you try to create an alert filter with the wrong type (for example, introducing a string into a numeric data type filter) there is a null pointer exception. Also, there are multiple duplicate warnings saying that the data may be wrong or incomplete.

**Steps to reproduce**

- Go to `Modules > events`
- Add a filter for rule.level = "error"

**Screenshot**
image
image
image

**Additional context**
This seems to happen in all browsers and system.

Nevertheless, on chrome/chromium, you can only add one letter: 
image

While in firefox you can introduce full words.

image

asteriscos:
Hi team, this fixes:

**When you tried to create an alert filter with the wrong type or with an empty value an exception was thrown.**
E.g when typing letters in a number input field.

Closes #2887


Desvelao:
I tested the PR and I got a rare problem. I get a blank page in Kibana with this error in the console:

I am not sure if this is my environment or it is happening in this branch. Could someone test it?
image

Desvelao:
The change in the `CHANGELOG.md` is missing. You should add it.

pablomarga (changes requested):
I have tested it and seems it is working when you select `is` or `is not` options. But when you tried for example  `is one of` option it is still happening.

https://user-images.githubusercontent.com/46006326/108112912-11722d00-7097-11eb-92a7-8075a2e0901b.mov

Desvelao:
I tested again the PR and it is working. The next toast is displayed:
image

In my opinion, I think we should not add custom behavior to the discover search bar. The Kibana Discover is working in the same way, showing the toast with the error.

asteriscos:
> I have tested it and seems it is working when you select `is` or `is not` options. But when you tried for example `is one of` option it is still happening.
> 
>  Grabacion.de.pantalla.2021-02-16.a.las.20.37.37.mov

I was able to reproduce this issue. Unfortunately, that operator allows the use of strings for number type inputs and I didn't find a way to detect when a filter field is a number type. So I cannot stop this behavior as string type fields will use this as well and I won't know which is which.
image

asteriscos:
Added changes to `CHANGELOG.md`

pablomarga (approved):
LGTM!

PR #2964, opened 2021-02-12
Fix/wrong number of alerts

### LINKED ISSUE #2922: Wrong number of security events momentarily shown while aplying filter
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.0  | 7.10.0     | 4101-1 |

**Description**

We found that when the cluster name changes the `Security events` dashboard shows the total number of alerts during a short moment while the filter on `cluster.name` is being aplied.

**Steps to reproduce**

1. Create an all in one environment.
2. Wait for alert to be generated.
3. Set a Wazuh cluster.
4. Access `Security events`

**Video**

https://user-images.githubusercontent.com/19505384/106755753-c97eef00-662e-11eb-8f1c-a873eb5b155f.mp4

**Additional context**

Maybe this number should not be  shown until the filter has been applied.

Regards,
Daniel Folch

pablomarga:
Hi teams, this PR fix the error that showed wrong number of security events were shown while aplying the filter.
Closes #2922


Desvelao:
LGTM!

This PR solves the problem when there some alerts whose first filter applied in the Dashboard is not set and the metrics are loading without filters, then some filter is applied by the app (`cluster.name` or `manager.name`) and the metrics reload and show another quantity of alerts. The PR tricks the render of the `Metrics` component when the visualizations are ready.

asteriscos (approved):
Tested it as the issue was described and works as expected. LGTM!

Issue #2789, opened 2021-01-19
Properly logout of Wazuh API when logging out of the application


davidjiglesias:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.4   | 7.9.1/7.9.3     | 4015 |

**Description**
The Kibana APP is not calling [DELETE /security/user/authenticate](https://documentation.wazuh.com/4.0/user-manual/api/reference.html#operation/api.controllers.security_controller.logout_user) to logout current Wazuh API user and revoke the issued JWT token when logging out of the application.

This causes the JWT token to remain valid and allows to replay a previously-captured request after a user logs out of the application.

Regards,

David J. Iglesias

gabiwassan:
Working on this issue on this [PR](https://github.com/wazuh/wazuh-kibana-app/pull/2972)

gabiwassan:
> Working on this issue on this [PR](https://github.com/wazuh/wazuh-kibana-app/pull/2972)

PR on revision. For the moment only work for Open Distro on an explicit logout

PR #2972, opened 2021-02-12
Bugfix delete old token

gabiwassan:
Hi team,
This PR is to removed old tokens on api.

Closes: #2789


Desvelao:
I tested the PR and it seems it is working.

I checked the API request received and it seems to receive multiple `DELETE /security/user/authenticate`. Could someone check this?

How to reproduce:
- Login to Kibana and go to the app
- Copy the token from the cookie `wz-token` and test the token works using a curl request.
- Open the menu to log out and click the `Log out` button
- Redo the curl request using the token and check the response is unauthorized.

gabiwassan:
This fix only found for Open Distro plugin security

Issue #2963, opened 2021-02-12
Non-stop error messages when trying to manage agents on an unauthorized group 


s-ocando:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.0   | 7.10.0     | 4101-2 |

**Description**
A user with permission to read agent 001 can access the default group section by clicking on the group label. If the user clicks on "Manage agents" error messages start appearing and will not stop until the session is terminated. 

![output](https://user-images.githubusercontent.com/61882981/107745852-35d7ad80-6d15-11eb-87f7-621295655f84.gif)

PR #2976, opened 2021-02-15
Fix a non-stop error in Manage agents when the user has no permissions

Desvelao:
Hi team, this PR resolves:

- Fix non-stop error in Manage agents when the user has no permissions
- Added an authorization prompt to the group detail view

Closes: #2963


sortizowlh:
Tested:
Now, if the user does not have permission to view groups, a message will be displayed stating that the user does not have permissions.

sortizowlh (approved):
Check!

pablomarga (approved):
LGTM!

PR #2997, opened 2021-02-22
Fixed health check runs after every refresh

frankeros:
Hi team, 
this resolves health-check runs after every refresh

### Steps to reproduce it
1. Open Wazuh app for first time (with cache cleared) 
2. Health-check runs (OK)
3. Navigate to any section of the app
4. Press F5
5. Health-check runs (KO)


asteriscos (approved):
tested. LGTM!

Issue #3000, opened 2021-02-23
Clicking on police L1 it shows me data from policy L1 and data from policy L2 on SCA


CPAlejandro:
By clicking on police L1 it shows me data from policy L1 and data from policy L2.
Also if I enter L1 and refresh, I get the L2 data
https://images.zenhubusercontent.com/62186486/daa4996d-4add-41be-b5e5-7b7fcbd207c2/scabug.mp4

Desvelao:
### Changes
The request to get the SCA policy with the results and name is wrong formulated and all policies are returned and selected the first. The parameters to filter the selected SCA policy is built wrong way.

PR #3007, opened 2021-02-24
Fix SCA selection policy showing information about another policy

Desvelao:
Hi team, this PR resolves:
 
- Fixed a bug when an agent has more than 1 SCA policy and select one that is not the first in the SCA inventory table. The data about the first policy was shown in the policy details. The policy checks are working well.

Closes: #3000


frankeros (approved):
Tested :heavy_check_mark: 
LGTM!

Issue #3001, opened 2021-02-23
Creating a role mapping before the existing ones are loaded, the page bursts


CPAlejandro:
If we try to create a role mapping before the existing ones are loaded, the page bursts.

https://user-images.githubusercontent.com/46006326/108984453-3c104700-7690-11eb-95a9-57d133f9aece.mov

PR #3013, opened 2021-02-24
Checking if data is loaded before show button for create role mapping

sortizowlh:
Hi team,
This PR hides the create role-mapping button while the data is loading.

Closes: [#3001](https://github.com/wazuh/wazuh-kibana-app/issues/3001)

### DIFF
--- a/public/components/security/users/users.tsx
+++ b/public/components/security/users/users.tsx
@@ -120,13 +120,11 @@ export const Users = () => {
         <EuiPageContentHeaderSection>
           {
             !rolesLoading
-            ?
+            &&
             <div>
               <EuiButton onClick={() => setIsCreateFlyoutVisible(true)}>Create user</EuiButton>
               {createFlyout}
             </div>
-            :
-            null
           }
         </EuiPageContentHeaderSection>
       </EuiPageContentHeader>
--- a/public/components/security/users/users.tsx
+++ b/public/components/security/users/users.tsx
@@ -118,10 +118,16 @@ export const Users = () => {
           </EuiTitle>
         </EuiPageContentHeaderSection>
         <EuiPageContentHeaderSection>
-          <div>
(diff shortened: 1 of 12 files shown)


pablomarga:
It is working as expected but I find this behaviour strange. On the other tabs the create button appears instantly.

https://user-images.githubusercontent.com/46006326/109306281-3c9b1000-783f-11eb-9852-74c56b131718.mov

sortizowlh:
> 
> 
> It is working as expected but I find this behaviour strange. On the other tabs the create button appears instantly.
> Grabacion.de.pantalla.2021-02-26.a.las.14.21.21.mov

Going to add the same behavior to all `create` buttons for `security `tabs

frankeros:
> It is working as expected but I find this behaviour strange. On the other tabs the create button appears instantly.

I agree we should use the same behavior in all tabs of this section.

sortizowlh:
Now, all create buttons for the Security module has the same behavior

public/components/security/users/users.tsx
```diff
@@ -118,10 +118,16 @@ export const Users = () => {
           </EuiTitle>
         </EuiPageContentHeaderSection>
         <EuiPageContentHeaderSection>
-          <div>
-            <EuiButton onClick={() => setIsCreateFlyoutVisible(true)}>Create user</EuiButton>
-            {createFlyout}
-          </div>
+          {
+            !rolesLoading
```
frankeros, line 122: you could use `&&` instead of a nullable ternary operator 
```
{ 
  !rolesLoading && <div> .. </div>
}
```

sortizowlh: All nullable ternary operators have been removed.

frankeros (approved):
tested locally :heavy_check_mark: 
LGTM! :tada: 
Good job!

Issue #2981, opened 2021-02-17
Roles Mapping Editor unexpected behavior


elwali10:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.0.4   | 7.9.3     | 4016 |

**Description**

Opening/Editing the role mapping causes adding an AND operator affecting the mapping of users using it. 

**Demo :** 

![bugJsonEditor](https://user-images.githubusercontent.com/16334770/108196695-b33a5e00-7119-11eb-8128-6e24f90b4da3.gif)

PR #3028, opened 2021-02-26
2981/roles mapping unespected behavior

sortizowlh:
Hi team,
This PR checks an invalid behavior that added a logical operator AND or OR without needed.

Closes: #2981


public/components/security/roles-mapping/helpers/rule-editor.helper.ts
```diff
@@ -11,24 +11,34 @@ export const getJsonFromRule = (internalUserRules, rules, logicalOperator) => {
...
-        OR: usersRulesArray,
-      },
-      {
-        [logicalOperator]: rulesArray,
-      },
-    ];
-  } else {
-    if (usersRulesArray.length) {
-      ruleObject['OR'] = usersRulesArray;
+    if(rulesArray.length == 1){
```
frankeros, line 17: this is causing bad behavior because is removing the `userRulesArray`

frankeros: please add some unit test to this function with this uses cases
- Giving userRulesArray NO empty and rulesArray NO empty, should return an object with one `'OR'` array with both arrays as items with all items
- Giving userRulesArray with 1 item  and rulesArray empty, should return the single item
- Giving rulesArray with 1 item  and userRulesArray empty, should return the single item
- Giving rulesArray with 2 o more items  and userRulesArray empty, should return an object with one item equals to the `logicalOperator` (OR or AND) array with all items
 - Giving userRulesArray with 2 o more items  and rulesArray empty, should return an object with one `'OR'` array with all items

frankeros (changes requested):
Tested locally :x: 

-- 

Adding both kind of filters from the UI editor and switch to JSON editor, is removing the user filter
image
image

sortizowlh:
Testing behavior. When adding some code to the JSON editor, the visual editor keeps these changes.

public/components/security/roles-mapping/helpers/rule-editor.helper.test.ts
```diff
@@ -201,4 +201,159 @@ describe('Rule Editor Helper', () => {
       });
     });
   });
-});
+
+  describe('Test role-mapping behavior', () => {
+    describe('Given a non null rules array and non null internal users array ', () => {
```
gabiwassan, line 206: Here the title says that rules and internal users are null, but they are not. They are correct? Or maybe you can copy this test and create a new one with null values and cover this case 😄 

Great job!

gabiwassan (changes requested):
Great job!
Just a simple comment

Desvelao (approved):
LGTM!

tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "jsx": "react",
```
Desvelao, line 4: Is this necessary?

Desvelao:
The change should be added to `CHANGELOG.md` file.

frankeros (approved):
tested locally :heavy_check_mark: 
Great job :tada:

pablomarga (approved):
LGTM!

Desvelao:
## Changes
- Updated the branch with the last changes of `4.1-7.10` branch
- Added change into the changelog

PR #3066, opened 2021-03-10
Fix deployment command line macos agents

### LINKED ISSUE #3058: Bad variable declaration for macOS agent deployment
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
|4.1.0 | 7.10.0 | 4101-3|
|4.1.0 | 7.10.2 | 4101-3|
| 4.1.2   | 7.10.0     | 4103 |

**Description**
As mentioned in the additional context of #2988 the wrong syntax is used for `launchctl setenv`.

**Steps to reproduce**

1. Go to Agents
2. If agents have been registered before "Deploy new agent" must be clicked on.
3. Select all options necessary to deploy a macOS agents
4. Copy and paste command into a macOS computer
5. See that if no group or manager password is specified the command results in `Usage: launchctl setenv <<key> <value>, ...>`
6. If more than one variable is specified then the second variable is taken as the value for the first resulting in an unregistered agent with a malformed configuration (for example with `<address>my.wazuh.manager.com=WAZUH_AGENT_GROUP=default</address>`

**Additional context**
This is because it uses the same [OptionalDeploymentVariables](https://github.com/wazuh/wazuh-kibana-app/blob/4.1-7.10/public/controllers/agent/components/register-agent.js#L285-L293) function ([here](https://github.com/wazuh/wazuh-kibana-app/blob/4.1-7.10/public/controllers/agent/components/register-agent.js#L408-L409)) that is valid in Linux environments and not valid for macOS.

pablomarga:
Hi team, this PR solve:
- Bad variable declaration for macOS agent deployment

Closes #3058

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Change Log
 
 All notable changes to the Wazuh app project will be documented in this file.
+
+## Wazuh v4.2.0 - Kibana 7.10.0 , 7.10.2 - Revision 4201
+
+### Fixed
+
+- Fix wrong variable declaration for macOS agents [#3066](https://github.com/wazuh/wazuh-kibana-app/pull/3066)
+
+
 ## Wazuh v4.1.2 - Kibana 7.10.0 , 7.10.2 - Revision 4103
 
 ### Added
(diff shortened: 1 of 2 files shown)


frankeros (approved):
LGTM!

PR #3097, opened 2021-03-23
Fix Index pattern setting not used 

### LINKED ISSUE #3074: Index pattern setting not used for choosing from existing patterns
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.2   | 7.10     | - |

**Description**
The value set for the "default index pattern to use on the app" ([pattern setting](https://documentation.wazuh.com/current/user-manual/kibana-app/reference/config-file.html#pattern)) is not used for choosing a pattern to load when multiple valid Wazuh indices exist.

Furthermore since the request used in [saved-objects.js](https://github.com/wazuh/wazuh-kibana-app/blob/master/public/react-services/saved-objects.js#L28) does not use the `sort_field` option the index pattern used can be somewhat random, since apparently Kibana's API does not have a default sorting.

**Steps to reproduce**

1. Configure multiple valid Wazuh index patterns (e.g. wazuh-alerts-* and wazuh-archives-*)
2. Access the Wazuh app "for the first time" (clear all cookies or use private browsing mode)
3. The Wazuh app chooses one of the available index patterns, basically the first valid one from the list (not the default pattern one). Depending how the Kibana API sorts them, it can differ when trying multiple times.

**Additional context**
During initialization of the Wazuh app [checkPatterns()](https://github.com/wazuh/wazuh-kibana-app/blob/master/public/components/health-check/health-check.tsx#L60) is used to check for valid indices and set it. When no cookie exists, it calls [getListOfWazuhValidIndexPatterns()](https://github.com/wazuh/wazuh-kibana-app/blob/master/public/react-services/saved-objects.js#L45) and uses the first index from the list. At this point the default index pattern from the configuration could be chosen, if it exists, and the first one from the list as a backup, if it does not exist.

pablomarga:
Hi team this PR solves:
- Index pattern setting not used for choosing from existing patterns

Closes #3074

### How to Test:
To test this PR first you need to create a new index pattern in Stack Management > Index pattern (if it is not created it will be created in the health check). Then you have to add this configuration to your `wazuh.yml` `pattern: '<INDEX_PATTERN_NAME>'`. After this you should clean the cookies and the caché before entering to the app. Finally click on Wazuh and the default index pattern should be the same that you had created.


Desvelao:
I tested the PR and it is working. Good job!

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to the Wazuh app project will be documented in this file.
 
+## Wazuh v4.1.3 - Kibana 7.10.0 , 7.10.2 - Revision 4105
+
+### Fixed
+
+- Index pattern setting not used for choosing from existing patterns [#3097](https://github.com/wazuh/wazuh-kibana-app/pull/3097)
+
```
Desvelao, line 10: This needs to be updated to `4.2.0-4201`

PR #3128, opened 2021-04-07
Fix redirect to settings after the health check

pablomarga:
Hi team this PR solves:
- The app redirect to Settings section after the health check 
- If the wazuh-manager service was stopped the button did not appeared.
- Remove some unused code.

### How to test
To test this PR you have to `stop the manager service` and then try to pass the health check. After that the button should appeared and then you need to `start the manager service` again before clicking on the button. Finally after the health check is passed you should be redirected to `overview` section

### DIFF
--- a/public/components/health-check/health-check.tsx
+++ b/public/components/health-check/health-check.tsx
@@ -183,21 +183,21 @@ export class HealthCheck extends Component {
     let errors = this.state.errors;
     let apiChanged = false;
     const buttonRestartApi = <div> <span><EuiIcon type="alert" color="danger" ></EuiIcon> Error</span>
-            {<EuiToolTip
-              position='top'
-              content='Try to reconnect to the API'
-            >
-              <EuiButtonIcon
-                display="base"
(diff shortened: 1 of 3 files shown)


public/components/health-check/health-check.tsx
```diff
@@ -118,22 +118,29 @@ export class HealthCheck extends Component {
       const response = await GenericRequest.request('GET', '/hosts/apis');
       const hosts = response.data;
       const errors = [];
+      const results = this.state.results;
 
       if (hosts.length) {
         for (let i = 0; i < hosts.length; i++) {
-          let tries = 36;
+          const maxTries = 5;
```
gabiwassan, line 125: `maxTries` is declared for every loop on for?

gabiwassan (approved):
LGTM

public/components/health-check/health-check.tsx
frankeros: yes, your are right

PR #3140, opened 2021-04-12
Fixed problems with permissions in Rules, Decoders and Groups

CPAlejandro:
Hi guys, we have fixed a problem that causes us to not be able to interact with the buttons of manage rules files / decoders and add new rules / decoders file together with the bug of not being able to enter groups having permissions.


frankeros (approved):
LGTM!
We need to support it without changes, for now, its ok

PR #3145, opened 2021-04-13
Fix not retrying the API connection in the healthcheck

Desvelao:
### Description
Hi team, this PR resolves:
- Fix when the API is down not retrying the connection.

Note: this has a temporal fix

### Checks
- Stop the Wazuh manager service/binary
```
service wazuh-manager stop
```
- Stop the container/down the machine with the Wazuh manager of the used Wazuh API

In both cases, it should allow to retry the connection and it should appear the `Retrying` in the API connection and API version checks


frankeros (approved):
LGTM!

PR #3167, opened 2021-04-15
Fixing when most active agent doesn't apper

CPAlejandro:
Hello guys, in this PR we solved an error that caused that when you had permission for all users in / Agents -> Most Active Agent nothing was shown


Desvelao (approved):
LGTM!

I don't like the conditional without curly braces `{}` but it works anyways

frankeros (approved):
LGTM!

PR #3184, opened 2021-04-19
[FEATURE] Add redirect to group details using the URL

Desvelao:
### Description

Redirect to group detail using the `group` parameter in the URL when the `Management > Groups` is accesed.

  - Redirect to group details when the URL contains group query param
  
### Checks
In a new tab, use the next URL to access to group details directly:
```
KIBANA_HOST_ADDRESS/app/wazuh#/manager/?tab=groups&group=GROUP_NAME
```
- If the group:
  - exists and the user can get information about the group, it should display information about the group.
  - not exists: it should display the main groups section.
  - without permission to that group: it should display the main groups section.

Replace `KIBANA_HOST_ADDRESS` and `GROUP_NAME` for your case

### DIFF
--- a/public/controllers/management/components/management/groups/groups-main.js
+++ b/public/controllers/management/components/management/groups/groups-main.js
@@ -24,6 +24,7 @@ import {
 } from '../../../../../redux/actions/groupsActions';
 import { connect } from 'react-redux';
 import { updateGlobalBreadcrumb } from '../../../../../redux/actions/globalBreadcrumbActions';
+import { WzRequest } from '../../../../../react-services/wz-request';
 
 class WzGroups extends Component {
   constructor(props) {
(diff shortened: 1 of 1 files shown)


public/controllers/management/components/management/groups/groups-main.js
```diff
@@ -39,8 +40,20 @@ class WzGroups extends Component {
     store.dispatch(updateGlobalBreadcrumb(breadcrumb));
   }
 
-  componentDidMount() {
+  async componentDidMount() {
     this.setGlobalBreadcrumb();
+    console.log(window.location.href)
```
gabiwassan, line 45: Remove this console log pls :)

public/controllers/management/components/management/groups/groups-main.js
```diff
@@ -39,8 +40,20 @@ class WzGroups extends Component {
...
+    // Check if there is a group in the URL
+    const [_, group] = window.location.href.match(new RegExp('group=' + '([^&]*)')) || [];
+    window.location.href = window.location.href.replace(new RegExp('group=' + '[^&]*'), '');
+    if(group){
+      try{
+        // Try if the group can be accesed
+        const responseGroup = await WzRequest.apiReq('GET', '/groups', {params: {groups_list: group}});
+        const dataGroup = responseGroup?.data?.data?.affected_items?.[0];
```
gabiwassan, line 53: `?.` it's ok? we don't have problems whits these?

public/controllers/management/components/management/groups/groups-main.js
```diff
@@ -39,8 +40,20 @@ class WzGroups extends Component {
...
+        this.props.updateGroupDetail(dataGroup);
+      }catch(error){};
```
gabiwassan, line 55: In this catch, could be set `updateGroupDetails` with some value or isn't necessary?

gabiwassan (changes requested):
Some questions, great job!

public/controllers/management/components/management/groups/groups-main.js
Desvelao: Yes, it should work now when the app package is built.

public/controllers/management/components/management/groups/groups-main.js
Desvelao: :+1:

public/controllers/management/components/management/groups/groups-main.js
Desvelao: It is not necessary.

frankeros (approved):
tested :heavy_check_mark: 
LGTM!

PR #3217, opened 2021-05-04
Fixed fields overlap in the agent summary screen

### LINKED ISSUE #3204: Fields overlap in the agent summary screen
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.2.0  | 7.10.2     | 4201-8 |

**Description**
Some fields including operating system, cluster node and registration date overlap on the agent summary dashboard.

**Screenshots**
image

CPAlejandro:
Hi team, 
We have fixed that when we have a very large operating system or cluster name, these overlap with the rest of the fields, within the overview of an agent
Closes #3204 

**Steps to test it:**
In `agents-info.js` we modify in line 177, after the if, the variable `agent.node_name`, and we put a long node name and then in line 78 we change `osName` to another super long name, which will change the name of the operating system to display.
Then we go to `/Agents`, we select an agent and the names of both should not overlap.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@ All notable changes to the Wazuh app project will be documented in this file.
 ## Wazuh v4.2.0 - Kibana 7.10.0 , 7.10.2 - Revision 4202
 
 ### Fixed
-
+Hola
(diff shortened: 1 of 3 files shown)


pablomarga (approved):
Tested it and works fine.LGTM!

PR #3245, opened 2021-05-13
Fix broke format in discover after manual refresh of the index pattern

### LINKED ISSUE #3215: Broken format in Discover due to checks.metaFields setting
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.5  | 7.10.0  |  4106 |

**Description**

If we refresh the index pattern **manually** from the Kibana settings, the format in Discover is broken (screenshot 3).

**Details**
- There are 5 meta fields. If  `_source` and `_index` meta fields are removed, the **manual** refresh of the index pattern fails. This causes the broken format in Discover.
- By default, the app uses a setting (`checks.metaFields: true`) to remove all the fields starting with `_`.

**Screenshots**

<details>
<summary>1</summary>

image

</details>

<details>
<summary>2</summary>

image

</details>

<details>
<summary>3</summary>

image

</details>

<details>
<summary>4</summary>

image

</details>

<details>
<summary>5</summary>

image

</details>

> **The main goal is to remove all meta fields (or most of them) without breaking the Discover format and allowing the index patterns to be refreshed manually without any issues**

pablomarga:
Hi team this PR fix an error that broked the format of Kibana's Discover after a manual refresh of  the index pattern. It happened due to we used to removed all metafields from the index pattern and the one that is in charge of the format is `_source`. Because an error from Kibana I have also added `_index` to the index pattern.

### How to test
- Go to Stack management > Advance settings and look for Meta fields setting, then change it to default and remove the current index pattern
- After this pass the health check and look for that setting again, there should only be `_source` and `_index`.
- If you change that setting and add another metafield and then pass the health check the setting should remain the same as you have previously setted

Closes #3215


public/components/health-check/lib/check-kibana-settings.ts
```diff
@@ -24,13 +26,15 @@ async function getKibanaSettings(): Promise<responseKbnSettings> {
...
+  return !!metaFields && !isEqual;
 }
 
 async function updateMetaFieldsSetting(isModified:boolean) {
+  console.log("UPDATE",!isModified)
   return !isModified && await GenericRequest.request(
     'POST',
     '/api/kibana/settings',
-    {"changes":{"metaFields":[]}}
+    {"changes":{"metaFields":["_source","_index"]}}
```
Desvelao, line 38: Why hardcode the metafields? You could use a constant as:
```
WAZUH_KIBANA_SETTING_METAFIELDS = ['_source', '_index']
```

Add the contant to the `constants.ts` file and adapt the comparation with `_isEqual` (not use userValue in the constant)

public/components/health-check/lib/check-kibana-settings.ts
```diff
@@ -24,13 +26,15 @@ async function getKibanaSettings(): Promise<responseKbnSettings> {
 
 async function checkMetafieldSetting({settings}: responseKbnSettings) {
   const { metaFields } = settings;
-  return !!metaFields && !!metaFields.userValue.length;
+  const isEqual = _.isEqual(WAZUH_METAFIELDS, metaFields)
+  return !!metaFields && !isEqual;
 }
 
 async function updateMetaFieldsSetting(isModified:boolean) {
+  console.log("UPDATE",!isModified)
```
Desvelao, line 34: Remove the `console.log`

public/utils/known-fields.js
```diff
@@ -15,8 +15,8 @@ export const KnownFields = [
     "esTypes": [
       "_index"
     ],
-    "searchable": true,
-    "aggregatable": true,
+    "searchable": false,
+    "aggregatable": false,
```
Desvelao, line 19: Is this necessary? If it doesn't, remove it.

frankeros: Both should be `false`

gabiwassan (approved):
LGTM!

public/components/health-check/lib/check-kibana-settings.ts
```diff
@@ -24,13 +26,14 @@ async function getKibanaSettings(): Promise<responseKbnSettings> {
 
 async function checkMetafieldSetting({settings}: responseKbnSettings) {
   const { metaFields } = settings;
-  return !!metaFields && !!metaFields.userValue.length;
+  const isEqual = _.isEqual(WAZUH_METAFIELDS_SETTINGS, metaFields && metaFields.userValue);
+  return !!metaFields && !isEqual;
```
frankeros, line 30: you could simplify with
```
return !!metaFields && _.isEqual(WAZUH_METAFIELDS_SETTINGS, metaFields.userValue);
```

frankeros (changes requested):
Merge with 4.2-7.10 and move it to the new health-check architecture pls

PR #3256, opened 2021-05-17
Fix/graph single api 3239

### LINKED ISSUE #3239: The evolution graph of agents shows the agents for all configured APIs
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.2.0   | 7.x     | 4201 |

**Description**
The evolution graph of agents shows the agents for all configured APIs

**Acceptance Criteria**
The evolution graph must display only the agents included by the current API. But this only can be applied by the manager and workers' names included in the monitoring indices. 
This will only work properly if the clusters' names and their workers' names from each APIs are different.

By @wazuh/frontend
**Steps to reproduce**

1. Go to 'Agents'

**Screenshots**
image

**Tasks:** 
- [x] Add filter by managers names corresponding to the current API only for the Agents evolution graph
- [x] Add the business rule to the Troubleshooting for Kibana in the documentation web.

mpRegalado:
Closes #3239
**Filtered evolution of agents by manager**
Added code to the Evolution graph so that it only shows agents for the currently selected manager.

**Possible issues**
The `wazuh-monitoring-*` indexes only store info about the API by name. For cluster APIs, this name is the name of the cluster; for single manager APIs, this name is the name of the manager.
Should two clusters or two independent managers have the same name, there wouldn't be enough unique information to identify which API they belong to and both would show on the graph.

**Steps to reproduce**
Visit the Agents section in the wazuh menu

**Screenshots**
Graph of an API with 1 disconnected agent, it does not show the connected agent from the other active API

> Graph of an API with 1 disconnected agent, it does not show the connected agent from the other active API

Graph of the other API with the opposite situation
Graph of the other API with the opposite situation


eze9252 (approved):
LGTM!

CPAlejandro (approved):
LGTM!

gabiwassan (approved):
LGTM!

pablomarga (approved):
LGTM!

Desvelao:
Test :x:

In an environment with:
2 managers nodes (1 master and 1 worker):
- Register an agent to the `worker` node.
The indexed document has the `manager` key as the worker, so the filter by `manager` is not useful to solve the issue if only filer by the name of one manager.
image
image

Possible solutions:
- Add to the visualization filters, the name of all manager nodes of the API. Something like:
```
{
 "query": {
   "term": {
     "manager.keyword": ["manager-master-node", "manager-worker-node1", "manager-worker-node2", etc...]
   } 
 }
}
```
- Add to the document the `id` of API. This will be a breaking change with the current data indexed to the document in the monitoring indices

Desvelao (changes requested):
PR requirements:
:x: Add the change to `CHANGELOG.md`

Test :x:

In an environment with:
2 managers nodes (1 master and 1 worker):
- Register an agent to the `worker` node.
The indexed document has the `manager` key as the worker, so the filter by `manager` is not useful to solve the issue if only filer by the name of one manager.
image
image

Possible solutions:
- Add to the visualization filters, the name of all manager nodes of the API. Something like:
```
{
 "query": {
   "term": {
     "manager.keyword": ["manager-master-node", "manager-worker-node1", "manager-worker-node2", etc...]
   } 
 }
}
```
- Add to the document the `id` of API. This will be a breaking change with the current data indexed to the document in the monitoring indices

mpRegalado:
I implemented the change suggested by @Desvelao and added a function to retrieve the list of managers from the current API to use it in the filter.
image

The issue of multiple managers having the same name still exists

frankeros (approved):
CR: LGTM!

public/controllers/agent/components/agents-preview.js
```diff
@@ -93,7 +94,12 @@ export const AgentsPreview = compose(
       return prev;
     }, {});
   };
-
+  async getManagers() {
+    const agents = await WzRequest.apiReq('GET', '/agents', {});
+    const managers = agents.data.data.affected_items.map((agent) => agent.manager);
+    const uniqueManagers = managers.reduce((unique,manager) => unique.includes(manager) ? unique : [...unique,manager], []);
+    return uniqueManagers;
+  }
```
Desvelao, line 102: This could cause bugs due to:
- The request is limited to the first 500 agents. If we need more agents, we need to paginate. It costs so much to get the managers.
- This could work if we get all agents because could exist an agent registered with other managers that are not returned with the request.

public/kibana-integrations/kibana-vis.js
```diff
@@ -215,7 +215,25 @@ class KibanaVis extends Component {
...
+          key: "manager.keyword",
+          alias: null,
+          negate: false,
+          disabled: false,
+        },
+        query: {       
+          bool: {
+            should: this.props.managers.map((manager) => {
+              return {term: {"manager.keyword" : manager}}
+            }),
```
Desvelao, line 230: You could use the `terms` query instead of multiple `term`.

mpRegalado:
Changed the filter so it search by cluster name. This means the conflict may arise now on clusters with same name instead of managers with same name

public/controllers/agent/components/agents-preview.js
```diff
@@ -51,6 +51,7 @@ export const AgentsPreview = compose(
     super(props);
     this.state = { data: [], loading: false, showAgentsEvolutionVisualization: false, agentTableFilters: [] };
     this.wazuhConfig = new WazuhConfig();
+    this.managers = [];
```
Desvelao, line 54: Remove if it is not required.

pablomarga (approved):
CR: :heavy_check_mark: 
Testing: :heavy_check_mark:

CPAlejandro (commented):
CR: LGTM!
Testing: LGTM!

PR #3279, opened 2021-05-20
Replace empty registry value name in FIM inventory

### LINKED ISSUE #3220: Replace empty registry value name in FIM inventory
**Description**
We've been working on wazuh/wazuh#8253.

Due to how Windows works, whenever a registry key is created a default value inside of it is also created which has an empty string as its name and is visualized as `(Default)` in regedit. In order to solve the false positive alerts we need to store the default value's name as an empty string on the agent DB, since Windows allows you to create a value named `(Default)` and would collide as it is doing with the name `@`. This causes the UI to show an empty field and might look a little weird:

default_value

We have considered changing the name through a view on the manager's DB, but that could create an increased load on every query done on the FIM inventory. Similarly, we thought about changing it in the API but since it only does a simple SQL query, adding additional logic could prove challenging. Also, neither of these solutions would allow the user to easily distinguish between a default registry value and a registry value named `(Default)`.

mpRegalado:
Closes #3220 
**Presented empty fields with a notice**
Added a function that returns a render method compatible with EuiColumn objects that, if the object is either an empty string or undefined, shows a label stating as such (for now only in Value Name on Registry details).
Also replicated the same functionality for the `syscheck.value_name` in the FIM Discovery dashboard to address the situation in the issue #3220

**Testing**
Create a enviroment with at least 1 windows agent with `syscheck` enabled
- Any empty fields on the FIM inventory should display the label, one can be easily found in the files inventory for a windows agent.
- Replacing line 129 in `public/kibana-integrations/discover/application/components/table/table.tsx` with `value={""}` will empty all the information displayed in all the extended rows of FIM discovery, checking the `syscheck.value_name` field should show the label.

**Screenshots**
image

image

### DIFF
--- a/public/components/agents/fim/inventory/lib/empty-field-handler.js
+++ b/public/components/agents/fim/inventory/lib/empty-field-handler.js
@@ -10,7 +10,7 @@
  * Find more information about this on the LICENSE file.
  */
 
-import { EuiCode, EuiIcon } from '@elastic/eui';
+import { EuiCode, EuiIcon, EuiFlexGroup, EuiFlexItem } from '@elastic/eui';
 import React from 'react';
 
 /* This function can be used to render possibly empty fields.
@@ -19,10 +19,10 @@ export const emptyFieldHandler = (renderFn = (value, record) => value) => {
   return (value, record) => {
     if (value === '' || value === undefined) {
       return (
-        <>
-          <EuiIcon type="iInCircle" />
-          <EuiCode>Empty field</EuiCode>
-        </>
+        <span style={{display:"flex", minWidth:"0"}}>
+            <EuiIcon type="iInCircle" />
+            <EuiCode className="wz-ellipsis" style={{whiteSpace:"nowrap"}}>Empty field</EuiCode>
+        </span>
       );
     } else {
       return renderFn(value, record);
--- a/public/components/agents/fim/inventory/lib/empty-field-handler.js
+++ b/public/components/agents/fim/inventory/lib/empty-field-handler.js
@@ -10,24 +10,22 @@
  * Find more information about this on the LICENSE file.
  */
 
-import { EuiCode, EuiIcon } from '@elastic/eui'
-import React from 'react'
+import { EuiCode, EuiIcon } from '@elastic/eui';
+import React from 'react';
 
 /* This function can be used to render possibly empty fields.
 It takes a render function suitable for an EuiTable and returns another. */
-export const emptyFieldHandler = (renderFn = (value,record) => value) => {
-    return (value, record) => {
-        if (value === "" || value === undefined) {
-            return (
-                <>
-                    <EuiIcon type="iInCircle"/>
-                    <EuiCode>
-                        Empty field
-                    </EuiCode>
-                </>
-            )
-        } else {
-            return renderFn(value,record);
-        }
+export const emptyFieldHandler = (renderFn = (value, record) => value) => {
+  return (value, record) => {
+    if (value === '' || value === undefined) {
+      return (
+        <>
+          <EuiIcon type="iInCircle" />
+          <EuiCode>Empty field</EuiCode>
+        </>
+      );
+    } else {
+      return renderFn(value, record);
     }
-}
+  };
+};
--- a/public/components/agents/fim/inventory/lib/emptyFieldHandler.js
+++ b/public/components/agents/fim/inventory/lib/emptyFieldHandler.js
@@ -1,0 +1,33 @@
+/*
+ * Wazuh app - Integrity monitoring components
+ * Copyright (C) 2015-2021 Wazuh, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Find more information about this on the LICENSE file.
+ */
+
+import { EuiCode, EuiIcon } from '@elastic/eui'
+import React from 'react'
+
+/* This function can be used to render possibly empty fields.
+It takes a render function suitable for an EuiTable and returns another. */
+export const emptyFieldHandler = (renderFn = (value,record) => value) => {
+    return (value, record) => {
+        if (value === "" || value === undefined) {
+            return (
+                <>
+                    <EuiIcon type="iInCircle"/>
+                    <EuiCode>
+                        Empty field
+                    </EuiCode>
+                </>
+            )
+        } else {
+            return renderFn(value,record);
+        }
+    }
+}
--- a/public/components/agents/fim/inventory/registryValues/registryValues.tsx
+++ b/public/components/agents/fim/inventory/registryValues/registryValues.tsx
@@ -47,25 +47,23 @@ export const RegistryValues = (props) => {
       field: 'date',
       name: 'Date',
       sortable: true,
-      render: emptyFieldHandler()
     },
     {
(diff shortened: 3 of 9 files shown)


public/components/agents/fim/inventory/lib/emptyFieldHandler.js
```diff
@@ -0,0 +1,33 @@
...
+                    <EuiCode>
+                        Empty field
+                    </EuiCode>
+                </>
+            )
+        } else {
+            return renderFn(value,record);
+        }
+    }
+}
```
gabiwassan, line 33: - The name of this file should be with kebab-case: `empty-field-handler.js` (if possible we have legacy files with camel case)
- We use 2 spaces on code style. Maybe you are not using `prettier` on your IDE.

gabiwassan (changes requested):
Just some conventions to apply. Great Job!

public/components/agents/fim/inventory/registryValues/registryValues.tsx
```diff
@@ -15,6 +15,7 @@ import { WzRequest } from '../../../../../react-services';
 import React, { useEffect, useState } from 'react';
 import valuesMock from './values.json';
 import { DIRECTIONS } from '@elastic/eui/src/components/flex/flex_group';
+import { emptyFieldHandler } from '../lib'
```
Desvelao, line 18: Semicolon `;`

public/components/common/modules/events-enhance-discover-fields.ts
```diff
@@ -106,7 +106,14 @@ export const EventsEnhanceDiscoverCell = {
   }), {
     contentRegex: /(\w+)/g,
     element: 'span'
-  })
+  }),
+  'syscheck.value_name': (content, rowData, element, options) => {
+    if (content) return;
+    const container = document.createElement("tr");
```
Desvelao, line 112: why is the `container` an `tr` HTML element?

Desvelao (changes requested):
CR: ❌ Review the suggested changes
Test: ✅ 
With a window size lower, the cell looks a little strange.
image

I couldn't get the agent report changes about the registry, so I did the suggested trick in the PR comment to get the event detail field:
image

public/components/common/modules/events-enhance-discover-fields.ts
mpRegalado: It was meant to be a `td` element to match the rest of the cells in the table. Fixed in [97821ec](https://github.com/wazuh/wazuh-kibana-app/pull/3279/commits/97821ec47caf3d8512f3928f093c2240bcb0e59c)

public/components/common/modules/events-enhance-discover-fields.ts
```diff
@@ -106,7 +106,14 @@ export const EventsEnhanceDiscoverCell = {
   }), {
     contentRegex: /(\w+)/g,
     element: 'span'
-  })
+  }),
+  'syscheck.value_name': (content, rowData, element, options) => {
+    if (content) return;
+    const container = document.createElement("td");
```
Desvelao, line 112: Maybe a `div` or `span` instead of an HTML related to a table. The `td`/`tr` elements are set by the table itself.

mpRegalado: Just checked and you were right, replaced the element with a `span`

pablomarga (approved):
CR: :heavy_check_mark: 
Testing: :heavy_check_mark: it works fine but as @Desvelao commented when you reduce the window the cell looks a little strange.
Screenshot_20210526_101735

public/styles/common.scss
```diff
@@ -808,6 +808,13 @@ md-switch.md-checked .md-thumb {
   display: block;
 }
 
+.wz-ellipsis{
+  white-space: nowrap !important;
+  overflow: hidden !important;
+  text-overflow: ellipsis !important;
+  min-width: 0;
```
asteriscos, line 815: when creating very generic classes I would be mindful of adding specific attributes that may affect others in an unexpected way.

asteriscos (commented):
CR: small improvements

public/components/agents/fim/inventory/lib/empty-field-handler.js
```diff
@@ -19,10 +19,10 @@ export const emptyFieldHandler = (renderFn = (value, record) => value) => {
   return (value, record) => {
     if (value === '' || value === undefined) {
       return (
-        <>
-          <EuiIcon type="iInCircle" />
-          <EuiCode>Empty field</EuiCode>
-        </>
+        <span style={{display:"flex", minWidth:"0"}}>
+            <EuiIcon type="iInCircle" />
+            <EuiCode className="wz-ellipsis" style={{whiteSpace:"nowrap"}}>Empty field</EuiCode>
```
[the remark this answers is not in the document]
mpRegalado: Due to the way `EuiCode` works, it generates 2 html elements nested inside each other. The outer one is a `span` that receives the `wz-ellipsis` class. The inner one is a code that gets the `style` object. The inner code tag has a `white-space: normal` that overrides the one on the class, so the only way I've found for it to display correctly is to manually set it again.

public/styles/common.scss
mpRegalado: Right, those were adding during testing and they slipped my mind when pushing the commit. Removed them in the last one, thank you!

public/components/agents/fim/inventory/lib/empty-field-handler.js
```diff
@@ -0,0 +1,31 @@
...
+ * Find more information about this on the LICENSE file.
+ */
+
+import { EuiCode, EuiIcon, EuiFlexGroup, EuiFlexItem } from '@elastic/eui';
+import React from 'react';
+
+/* This function can be used to render possibly empty fields.
+It takes a render function suitable for an EuiTable and returns another. */
+export const emptyFieldHandler = (renderFn = (value, record) => value) => {
+  return (value, record) => {
```
frankeros, line 19: It doesn't work when the value is an object that needs to be mapped, for example in registryValues.tsx
```
    {
      field: 'value',
      name: 'Value name',
      sortable: true,
      render: emptyFieldHandler((item) => item.name),
    },
    {
      field: 'value',
      name: 'Value type',
      sortable: true,
      render: emptyFieldHandler((item) => item.type),
    },
```

frankeros (changes requested):

public/components/agents/fim/inventory/registryValues/registryValues.tsx
```diff
@@ -46,23 +47,25 @@ export const RegistryValues = (props) => {
...
       name: 'Date',
       sortable: true,
+      render: emptyFieldHandler()
     },
     {
       field: 'value',
       name: 'Value name',
       sortable: true,
-      render: (item) => item.name,
+      render: emptyFieldHandler((item) => item.name),
```
frankeros, line 56: remove please `emptyFieldHandler` from all but this.
we'll only notify this field when is empty.

frankeros (changes requested):

public/components/agents/fim/inventory/lib/empty-field-handler.js
mpRegalado: addressed in [1e90416](https://github.com/wazuh/wazuh-kibana-app/pull/3279/commits/1e904165481426a44912e81766bfd8e7b1dfea73)

frankeros (approved):
Test: :heavy_check_mark: 
CR: :heavy_check_mark: 
LGTM! Great job :tada:

PR #3303, opened 2021-05-27
Fix alerts summary reports

### LINKED ISSUE #3275: [Reports] The alerts of vulnerabilities is not completly displayed in the report
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.x   | 7.x     | 4xxx |

**Description**
The alert count displayed in the Vulnerabilities dashboard module is not according to the showing in the report. We should review why it is happening. We should display 

**Steps to reproduce**

1. Go to the Vulnerabilities dashboard module and see the alert count by criticality
2. Generate a report
3. Go to Management > Reporting and open the last report generated
4. See the `Alerts summary section and review the alert count. These are not according to stats displayed in the dashboard

**Screenshots**
image
image

pablomarga:
Hi team this PR solves some problems related to the Alert summary of the reports and also fix the last scan date of the CIS CAT module. Closes #3275 

### How to test
Check if the last scan date of the CIS CAT module is displayed as is set in the kibana settings. Also create some reports from vulnerabilities, Security Alerts and other modules to check if the number of alerts is correct.


CPAlejandro (approved):
CR: LGTM!
Testing: 
The date in `CIS CAT` works perfectly, the reports in `Vulnerabilities` too. But when testing the reports in `VirusTotal` I find that as each alert is different, only some are shown, perhaps the name of `alerts summary` could be changed to `some alerts` or comment at the end that since there are so many alerts, we only show some of them.
image

mpRegalado (approved):
**CR:** LGTM! :heavy_check_mark: 
**Testing:** LGTM! :heavy_check_mark:
As a warning I would note the case where there are over a thousand elements to visualize in a report, but that seems like an unlikely edge case.

Issue #3313, opened 2021-05-31
Create an error boundary component as a HOC.


gabiwassan:
| Parent Issue |
|:---:|
| #3262  |

**Description**
Create an error boundary component as a HOC for handling errors when rendering react components,
improve that the error handling and the prompt to be displayed is specific for the component with error.

PR #3321, opened 2021-06-01
Added ErrorBoundary HOC and component.

gabiwassan:
Hi team,
This PR includes the implementation of the ErrorBoundary component and HOC to catch components with errors on rendering.
Plus added `loglevel` dependency to log errors, warnings, etc.
e.g of error prompt:
image

Closes: #3313

### DIFF
--- a/public/components/common/hocs/error-boundary/with-error-boundary.tsx
+++ b/public/components/common/hocs/error-boundary/with-error-boundary.tsx
@@ -1,3 +1,15 @@
+/*
+ * Wazuh app - React HOCs to manage user authorization requirements
+ * Copyright (C) 2015-2021 Wazuh, Inc.
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Find more information about this on the LICENSE file.
+ */
+
 import React, { Component } from 'react';
 import log from 'loglevel';
 import { EuiEmptyPrompt } from '@elastic/eui';
@@ -13,7 +25,7 @@ export default class ErrorBoundary extends Component {
   render() {
     // @ts-ignore
     if (this.state.errorInfo) {
-      return handleError(this);
+      return <HandleError props={this} />;
     }
     return this.props.children;
   }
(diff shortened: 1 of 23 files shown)


public/components/common/hocs/error-boundary/with-error-boundary.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    render() {
+      debugger;
+      // @ts-ignore
+      if (this.state.errorInfo) {
+        return handleError(this);
+      }
+      return <WrappedComponent {...this.props} />;
+    }
+  };
```
Desvelao, line 41: I think that the HOC could be refactored to:
```ts
export const withErrorBoundary = WrappedComponent => props => (
  <ErrorBoundary>
    <WrappedComponent {...props} />
  </ErrorBoundary>
)
```

If the refactoring is possible, I would move the `catchFunc`  and `handleError` functions as components methods.

Desvelao:
I suggest exporting the HOC in the `public/components/common/hocs`, so the HOC can be imported more easily.

public/components/common/hocs/error-boundary/with-error-boundary.tsx
gabiwassan: Great!
Maybe we can use [this](https://www.npmjs.com/package/react-error-boundary)

public/components/common/hocs/error-boundary/with-error-boundary.tsx
```diff
@@ -0,0 +1,75 @@
...
+const catchFunc = (error, errorInfo, ctx) => {
+  ctx.setState({
+    error: error,
+    errorInfo: errorInfo,
+  });
+
+  log.error({ error, errorInfo });
+};
+
+const ErrorComponent = (props: { ctx: any }) => {
```
frankeros, line 49: it would be nice if was a child component for future enhancements

frankeros:
please add unit test for the new components

public/components/common/hocs/error-boundary/with-error-boundary.tsx
gabiwassan: Great!

public/components/common/hocs/error-boundary/with-error-boundary.tsx
```diff
@@ -0,0 +1,20 @@
+/*
+ * Wazuh app - React HOCs to manage user authorization requirements
```
Desvelao, line 2: Fix the file description

Desvelao:
As a suggestion, you could add the license block to the new files (not to snapshots).

PR #3327, opened 2021-06-02
Feature/3316 error handler orchestrator

### LINKED ISSUE #3316: Create new error handler to be responsible for the error orchestration
| Parent Issue |
|:---:|
| #3262 |

 - [x] Create a new error handler to be responsible for the error orchestration (business rules for errors).

**Acceptance criteria**
 - Not all errors should be sent to the backend, only logs categorized with a certain level or higher. 
 - The errors displayed in the UI must be clear to the user and not prevent the use of the app. 
 - The component must display only NO business errors, the business errors must be handled by the corresponding component.
  - The error orchestrator must be responsible for manage what to do with the errors according to the error level and interfaces.
 
 **Alternatives**
 Check the libs [loglevel](https://github.com/pimterry/loglevel) and [loglevel-plugin-remote](https://github.com/kutuluk/loglevel-plugin-remote).
 [Here](https://www.loggly.com/blog/best-practices-for-client-side-logging-and-error-handling-in-react/) an example with error boundary and `loglevel`.

Machi3mfl:
Hi Team, This PR added:
- First attempt to Logger Service (orchestrator), this service is responsible for application logs.
Closes #3316


common/constants.ts
```diff
@@ -220,3 +220,38 @@ export enum WAZUH_MODULES_ID{
...
+// Logger
+// log criticality 
+
+/*export enum LoggerCriticality { 
+  UI = 'UI',
+  BUSSINESS = 'BUSSINESS',
+  CRITICAL = 'CRITICAL'
+}*/
+// creo que queda mejor el type para el intelligence, no se en cuanto a performance cual es mejor
+export type LoggerCriticality = 'UI' | 'BUSSINESS' | 'CRITICAL';
```
frankeros, line 253: I suggest use UILogSeverity

common/constants.ts
```diff
@@ -220,3 +220,38 @@ export enum WAZUH_MODULES_ID{
...
+
+// levels
+export type LoggerLevel = 'WARNING' | 'INFO' | 'ERROR';
```
frankeros, line 256: I suggest use UILogLevel

common/utils/logger/logger-service.ts
```diff
@@ -0,0 +1,108 @@
...
+    LoggerCriticality,
+    LoggerLevel
+} from '../../constants';
+
+export interface ILoggerOptions {
+    context: string;
+    level: LoggerLevel;
+    criticality: LoggerCriticality;
+    display?: boolean; // falta setearlo por defecto en false 
+    log?: boolean; // falta setearlo por defecto en false
```
frankeros, line 12: store instead of log (maybe)

common/utils/logger/logger-service.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    /**
+     * 
+     * @param message 
+     * @param options 
+     */
+    private launchLog(message: string, options: ILoggerOptions){
+        let display = options.display || false;
+        let log = options.log || false;
```
frankeros, line 62: both are const

common/constants.ts
```diff
@@ -220,3 +220,31 @@ export enum WAZUH_MODULES_ID{
...
+export const KIBANA_SETTING_NAME_MAX_BUCKETS = 'timelion:max_buckets';
+
+// Default metaFields Kibana setting set by the app
+export const WAZUH_KIBANA_SETTING_METAFIELDS = ['_source', '_index'];
+export const KIBANA_SETTING_NAME_METAFIELDS = 'metaFields';
+
+
+// Logger
+// log severity
+export type UILogSeverity = 'UI' | 'BUSSINESS' | 'CRITICAL';
```
frankeros, line 246: It would be nice to get severity options and level options from enum or const too
here is an example 
image

public/react-services/logger-service.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+import loglevel from 'loglevel';
+
+jest.mock('loglevel');
+jest.mock('../kibana-services', () => ({
+  getToasts: () => ({ add: () => {} }),
+}));
+
+describe('Wazuh logger service', () => {
+  describe('Login service test error', () => {
+    it('Should use only service of login without toast', () => {
```
frankeros, line 11: typo `logging` instead of `login`

frankeros: maybe `Should use logging service only, but no toast`

public/react-services/logger-service.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+          stack: 'jest testing',
+        },
+      };
+
+      logError('unit test error message', options);
+
+      expect(loglevel.error).toHaveBeenCalledTimes(1);
+    });
+
+    it('Should use only service of login and toast', () => {
```
frankeros, line 30: typo `logging` 
maybe `Should use logging service and toast service`

public/react-services/logger-service.tsx
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+const showToast = (color, title, text, time) => {
+  getToasts().add({
+    color: color,
+    title: title,
+    text: text,
+    toastLifeTimeMs: time,
+  });
+};
```
frankeros, line 34: why no call directly to getToasts()?

public/react-services/logger-service.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+};
+
+/**
+ * @param message
+ * @param options
+ */
+const logError = (message: string, options: ILoggerOptions) => {
+  //const level : UILogLevel = 'ERROR';
+  // i think is better to set level inside method, maybe in future.
```
frankeros, line 67: +1 if don't what is the difference between `logError` `logInfo` and `logWarning` ?

common/constants.ts
```diff
@@ -220,3 +220,72 @@ export enum WAZUH_MODULES_ID{
...
+
+export type ErrorOrchestrator = {
+  loadErrorLog: (uiErrorLog: UIErrorLog) => void;
+};
+
+export const ToastColor = {
+  SUCCESS: 'success',
+  WARNING: 'warning',
+  DANGER: 'danger',
+};
```
Desvelao, line 291: This file contains the defined constants. Maybe you could move the types to another localtion. Maybe a local types for the service. Same for the constants, these seem locals to the service.

package.json
```diff
@@ -48,6 +48,7 @@
     "js2xmlparser": "^3.0.0",
     "json2csv": "^4.1.2",
     "jwt-decode": "^2.2.0",
+    "loglevel": "^1.7.1",
```
Desvelao, line 51: Do we need a dependency for this? Will we use some advanced feature that the package offers and we don't want to develop?

public/react-services/error-orchestrator/error-orchestrator-base.ts
```diff
@@ -0,0 +1,58 @@
...
+  private async storeError(errorLog: UIErrorLog) {
+    await GenericRequest.request('POST', `/utils/logs/ui`, {
+      body: {
+        message: errorLog.error.message,
+        level: errorLog.level,
+        location: errorLog.location,
+      },
+    })
+      .then(loglevel.info)
+      .catch(loglevel.error);
```
Desvelao, line 56: Maybe this could be refactored to use `try/catch` block instead `.then/.catch` to manage the Promise.

public/react-services/error-orchestrator/error-orchestrator.factory.ts
```diff
@@ -0,0 +1,29 @@
...
+  switch (severity) {
+    case UIErrorSeverities.UI:
+      return new ErrorOrchestratorUI();
+    case UIErrorSeverities.BUSINESS:
+      return new ErrorOrchestratorBusiness();
+    case UIErrorSeverities.CRITICAL:
+      return new ErrorOrchestratorCritical();
+    default:
+      break;
+  }
```
Desvelao, line 28: Why do we have multiple constructors of loggers depending on `severity`? Could this be a property/param and use a logger?

common/constants.ts
gabiwassan: Yes, I am going to reorganize this, some constants will be used at a general level in the app, and those that are specific to the service I will move. Thx!

public/react-services/error-orchestrator/error-orchestrator-base.ts
gabiwassan: Absolutely, working on that! thx

public/react-services/error-orchestrator/error-orchestrator.factory.ts
gabiwassan: The idea is that they are independent, that they can be used individually, and thus by extending `ErrorOrchestratorBase`, they can scale better.
But it is open to suggestions obviously 😃

gabiwassan (approved):

PR #3355, opened 2021-06-10
Fix/3320 visualization label overlap

### LINKED ISSUE #3320: Agent evolution X axis overlap
| Wazuh | Elastic |
| ----- | ------- |
| 4.2   | 7.12     |

**Description**
The agent monitoring graph's x axis labels overlap each other at small container widths. This is not an issue I could replicate with elastic version 7.10

**Steps to reproduce**

1. Go to Agents
2. See error

**Screenshots**

mpRegalado:
Closes #3320 
**The issue**
In kibana 7.11+ there seems to have been changes that affect how elements change with the `viewport` width, this led to situations where containers could be too narrow and the X-axis labels overlap one another.
**The solution**
Setting the `params.categoryAxes.lablels.filter = true` for each of the visualizations that have that property allows the visualization plugin to automatically remove labels as they overlap.
**Additional notes**
Also, in this PR, I rewrote how the visualizations are stored in their files. Several properties of a raw visualization consist of a `JSON` as a `string`. This used to be in a single line as well. I refactored these to be objects inside a `JSON.stringify` call for identical results but much better readability and ease of use.
**Testing**

- [ ] Every visualization still loads properly, this should check that the refactor has not affected anything.

- [ ] On each of the visualizations of the types `area`, `histogram` and `line`, at low container widths, X axis labels should be progressively removed as to not overlap.


asteriscos (approved):
CR: :heavy_check_mark: 
Tested 7.10 - 7.12: :heavy_check_mark:

sortizowlh (approved):
CR: ✔️
Tested 7.11.2 ✔️

PR #3376, opened 2021-06-16
feat(error-orchestrator): Improved on createGetterSetter

gabiwassan:
Hi team,
This PR applies the ErrorOrchestratorService enhancement by applying createGetterSetter.
ex: 

```
try {
    ...........
} catch (error) {
    const options = {
      level: 'INFO',
      severity: 'CRITICAL',
      error: error;
    };

    getErrorOrchestrator().handleError(options);
}
```

### DIFF
--- a/public/kibana-services.ts
+++ b/public/kibana-services.ts
@@ -1,22 +1,22 @@
-let angularModule: any = null;
-let discoverModule: any = null;
-
+import { ErrorOrchestratorService } from './react-services';
 import {
   ChromeStart,
+  CoreStart,
   HttpStart,
   IUiSettingsClient,
-  ToastsStart,
-  SavedObjectsStart,
   OverlayStart,
+  SavedObjectsStart,
   ScopedHistory,
-  CoreStart,
+  ToastsStart,
 } from 'kibana/public';
 import { createGetterSetter } from '../../../src/plugins/kibana_utils/common';
 import { DataPublicPluginStart } from '../../../src/plugins/data/public';
 import { VisualizationsStart } from '../../../src/plugins/visualizations/public';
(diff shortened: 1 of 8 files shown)


Machi3mfl (approved):
LGTM!

pablomarga (approved):
LGTM!

public/kibana-services.ts
```diff
@@ -40,9 +40,9 @@ export const [getVisualizationsPlugin, setVisualizationsPlugin] = createGetterSe
 export const [getNavigationPlugin, setNavigationPlugin] = createGetterSetter<
   NavigationPublicPluginStart
 >('NavigationPlugin');
-export const [getErrorOrchestrator, setErrorOrchestrator] = createGetterSetter<ErrorOrchestrator>(
-  'ErrorOrchestrator'
-);
+export const [getErrorOrchestrator, setErrorOrchestrator] = createGetterSetter<
+  ErrorOrchestratorService
+>('ErrorOrchestratorService');
```
Desvelao, line 45: I don't undestand what is the reason to use a `createGetterSetter` for this.

gabiwassan: The idea is that we can use the orchestrator throughout the app without having to create a new instance every time, like the other services (toast, https, plugins, etc)

frankeros: I think that is a good approach to get the common services from one place, you need to only know from where import `kibana-services` (maybe it need a new name)

frankeros: Also, maybe not for this case where the instance is static but for another service that needs a constructor for its instance is good to get the same instance from a getterSetter

frankeros (approved):
LGTM!

asteriscos (approved):
CR: :heavy_check_mark:

PR #3383, opened 2021-06-18
Merge latest 4.2-7.10 into 4.3-7.10

frankeros:
Hi team this merges the latest changes in 4.2-7.10 into 4.3-7.10 branch


pablomarga:
After passing the Health Check this error happens and does not load the main page
Screenshot_20210621_115139

Desvelao:
I am experiencing the same problem that @pablomarga reported. 

I was investigating this and I found that some React services use the getter for the AngularJS module but using it out service. Example:
```
const app = getAngularModule();

export class MyService{
  constructor(){
    // Using the app variable. It expected that contains the module. At start, its value is null (app === null),
    this.$rootScope = app.$injector.get('$rootScope');
    // ....
  }
}
```

The React services are being exported in `public/react-services/index.ts`.

The problem comes when importing something from the `public/react-services/index.ts`, as `getErrorOrchestrator`. It does, the rest of the files exported are executed, and the `app` variable defined on some of these files has `null` as its value.

The `public/plugin.ts` file contains the import statement displayed in the screenshot. This executes all the React services and creates the bug indirectly.
image

Exported getter/setter in `public/react-services/index.ts` which is used before set the Angular Module.
image

Possible solutions:
- Move the `getter/setter` for the `ErrorOrchestratorService` to another file and import in `public/intex.ts` from this file instead of the `public/react-services/index.ts`
- Move the use of the `getter` to set the `app` variable to inside the function/method which uses it.. Example:
```
export class MyService{
  constructor(){
    // Using the app variable. It expected that contains the module. At start, its value is null (app === null),
    const app = getAngularModule(); // Move the line when requiring the Angular module
    this.$rootScope = app.$injector.get('$rootScope');
    // ....
  }
}
```

public/react-services/common-services.ts
```diff
@@ -0,0 +1,6 @@
+import { ErrorOrchestratorService } from './error-orchestrator/error-orchestrator.service';
```
gabiwassan, line 1: I think you forgot the copyright section

PR #3387, opened 2021-06-22
Identity components with hocs

gabiwassan:
Hi team,
In order to improve the identity of each component and for ErrorBoundary to provide us with more precise information, I suggest the following change (which we should implement in all our hocs) following [Convention: Wrap the Display Name for Easy Debugging](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging).
At the moment since the parent components are rendered from Angular we have to give identity explicitly, as I apply it for Management:

```
WzManagement.displayName = 'WzManagement';
ManagementWelcomeWrapper.displayName = 'ManagementWelcomeWrapper';
WzManagementConfiguration.displayName = 'WzManagementConfiguration';
```

After:
image

Before:
image

All hocs with displayName:
image

### DIFF
--- a/public/components/common/hocs/error-boundary/with-error-boundary.tsx
+++ b/public/components/common/hocs/error-boundary/with-error-boundary.tsx
@@ -12,9 +12,13 @@
 
 import React from 'react';
 import ErrorBoundary from '../../error-boundary/error-boundary';
+import { getDisplayName } from '../utils/utils';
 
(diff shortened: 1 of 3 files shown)


frankeros (approved):
LGTM! :mechanical_arm:

pablomarga (approved):
LGTM!

Issue #3392, opened 2021-06-23
Refactor try catch on Settings section


gabiwassan:
| Parent Issue |
|:---:|
|#3265| 

**Description**
In order to improve better error handling in the front-end, we should implement a better try-catch strategy in the Settings section, services, and functions.

PR #3393, opened 2021-06-23
Refactor try catch in Settings section

gabiwassan:
Hi team,
This PR implements a new strategy to catch and store/show errors in the Settings section.

Close: #3392


public/components/settings/api/api-table.js
```diff
@@ -122,16 +127,34 @@ export const ApiTable = compose(withErrorBoundary, withReduxProvider)(class ApiT
...
+        const downReason =
+          typeof error === 'string'
+            ? error
+            : (error || {}).message ||
+              ((error || {}).data || {}).message ||
+              'Wazuh is not reachable';
         const status = code === 3099 ? 'down' : 'unknown';
-        entries[idx].status = { status, downReason }; 
+        entries[idx].status = { status, downReason };
+        throw error;
```
frankeros, line 138: I think this is a expected error and it should not be trowed, the component is handling the error

public/components/settings/api/add-api.js
```diff
@@ -93,6 +97,20 @@ export const AddApi = withErrorBoundary (class AddApi extends Component {
...
+        severity: UI_ERROR_SEVERITIES.UI,
+        store: true,
+        error: {
+          error: error,
+          message: error.message || error,
+          title: `Wazuh API not reachable, please review your configuration: ${error.message || error}`,
+        },
+      };
+
+      getErrorOrchestrator().handleError(options);
```
frankeros, line 113: the error is handled by two different logics, I think only one must be the responsible for doing it.

public/components/settings/modules/modules.js
```diff
@@ -82,7 +120,19 @@ export class EnableModulesWrapper extends Component {
     this.setState({ extensions });
     try {
       this.currentApi && AppState.setExtensions(this.currentApi, extensions);
-    } catch (error) {} //eslint-disable-line
+    } catch (error) {
+      const options = {
+        level: UI_LOGGER_LEVELS.ERROR,
+        severity: UI_ERROR_SEVERITIES.BUSINESS,
+        error: {
```
frankeros, line 127: it is not displayed and not stored, right?

public/components/settings/api/api-table.js
gabiwassan: Yes in this case the component catches these errors and show them on a `ToolTip` like this:
image
I think could be better to log them on our file `.log` too.

public/components/settings/api/add-api.js
gabiwassan: In this case, the orchestrator will save the error in our `.log` and write a `loglevel.error` in the console. (because it is UI severity) like this:
image

public/components/settings/modules/modules.js
gabiwassan: By default is
`display: true`
`store: false`

pablomarga:
The `api-is-down.js ` try catch did not change.

pablomarga (approved):
CR :heavy_check_mark: 
Testing :heavy_check_mark:

Machi3mfl (approved):
CR ✔️
Testing ✔️

PR #3430, opened 2021-07-02
Notify when you are registering an agent without permissions

### LINKED ISSUE #3406: Attempting to register an agent - no warning to notify that the user does not have the required permissions
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.1.5  | 7.10.0   | 4106 |

## Description

When attempting to add a new agent, if the user does not have enough permissions, the Kibana command to register the agent **would not display the authd password**. Also, the group to assign the agent can not be selected.

image

<details>
<summary>Request details</summary>

- **Status Code: 500 Internal Server Error.** 
  - message: "3013 - Permission denied: Resource type: agent:id"
  - Request Payload:

```
{method: "GET", path: "/agents/000/config/auth/auth", body: {}, id: "1"}
body: {}
id: "1"
method: "GET"
path: "/agents/000/config/auth/auth"
```

- **Status Code: 500 Internal Server Error.** 
  - message: "3013 - Permission denied: Resource type: agent:id"
  - Request Payload:
```
{method: "GET", path: "/agents/000/config/request/remote", body: {}, id: "1"}
body: {}
id: "1"
method: "GET"
path: "/agents/000/config/request/remote"
```

</details>

Regards.

CPAlejandro:
Hi guys,
In this PR, we have added a notification when it is not possible to change the group for this agent due to lack of permissions and we have hidden the command option when we do not have permission to access the manager's password since this made the command erroneous.

**In order to test It:**

- We would have to enter with a user with group permissions and access the manager's password and see that neither of the two posters is shown.

- Then we would have to enter with a user without both permissions and see that both permissions are activated.

Close #3406


eze9252 (approved):
LGTM!

public/controllers/agent/components/register-agent.js
```diff
@@ -473,7 +485,14 @@ export const RegisterAgent = withErrorBoundary(
 
       const guide = (
         <div>
-          {this.state.selectedOS && (
+          {(!this.state.needsPassword && this.state.wazuhPassword=== "") ? (
+            <EuiCallOut
+              color="warning"
+              title={<> This section could not be displayed because you do not have permission to get access to the registration service. </>}
```
frankeros, line 491: why `registration service`? maybe `authorization coniguration` (?)

Desvelao: **reply**:
Wazuh documentation about the configuration that is requested:
https://documentation.wazuh.com/current/user-manual/reference/ossec-conf/auth.html#auth

image

public/controllers/agent/components/register-agent.js
```diff
@@ -473,7 +485,14 @@ export const RegisterAgent = withErrorBoundary(
 
       const guide = (
         <div>
-          {this.state.selectedOS && (
+          {(!this.state.needsPassword && this.state.wazuhPassword=== "") ? (
```
Desvelao, line 488: **thought**: I am not sure about this check to display the callout. Maybe we should have a state when the request fails.

eze9252: hmmm... I'm going to check this

public/controllers/agent/components/register-agent.js
eze9252: I think to have a relationship with the documentation. we should leave it like this

public/controllers/agent/components/register-agent.js
Desvelao: **review**:

I tested the PR and when there isn't a configurated password, the callout is displying:
image

Desvelao:
### Changes
  - Applied the callout to some steps where there is an error to get the registration service info
  - Changed the callout from warning to danger
  - Fix not reported errors, when the user has not permissions, some selectors didn't display

Issue #3334, opened 2021-06-04
Api name cuts off if the name of the api is too long


mpRegalado:
| Wazuh | Elastic |
| ----- | ------- | 
| 4.2   | 7.x     |

**Description**
If there is more than one API registered, and one of them has a name too long, it cuts off without any elipsis.

**Steps to reproduce**

1. Register a second API in `kibana/data/wazuh/config/wazuh.yml`, it doesn't need to be active
2. Set the name of the active API (`default` if no other is set) to a very long name
3. Open the app

**Screenshots**
![image](https://user-images.githubusercontent.com/80431234/120829802-1bdd0780-c55e-11eb-82ba-c9b4481bd97e.png)
![image](https://user-images.githubusercontent.com/80431234/120830237-9574f580-c55e-11eb-8bcf-1d5eb2173bcb.png)

**Additional context**
It's possible that a similar issue happens to the Index Pattern selector but I have not tested it yet, worth looking into.

PR #3445, opened 2021-07-07
fix size api selector and add basic test

eze9252:
Hi team, this resolves:

- Adapt API selector when api name is too long

**How to Test:** 

- Register a second API in kibana/data/wazuh/config/wazuh.yml, it doesn't need to be active
- Set the name of the active API (default if no other is set) to a very long name
- The API selector should show full name

Closes [#3334](https://github.com/wazuh/wazuh-kibana-app/issues/3334)

### DIFF
--- a/public/components/wz-menu/wz-menu.js
+++ b/public/components/wz-menu/wz-menu.js
@@ -632,7 +632,7 @@ export const WzMenu = withWindowSize(class WzMenu extends Component {
   }
 
   getApiSelectorComponent() {
-    let style = { minWidth: 'max-content' };
+    let style = { minWidth: 100, textOverflow: 'ellipsis' };
     if (this.showSelectorsInPopover){
       style = { width: '100%', minWidth: 200 };
     }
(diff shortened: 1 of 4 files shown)


pablomarga (approved):
CR: :heavy_check_mark: 
Testing: :heavy_check_mark:

public/components/wz-menu/wz-menu.js
```diff
@@ -632,7 +632,7 @@ export const WzMenu = withWindowSize(class WzMenu extends Component {
   }
 
   getApiSelectorComponent() {
-    let style = { maxWidth: 100 };
+    let style = { minWidth: 'max-content' };
```
mpRegalado, line 635: increasing the size the component can take is a good first step, but not limiting it in any way can lead to other issues, if the name is too long still, the other components in the menu will be squished and rendered incorrectly
image
I suggest you specify a max size and `text-overflow: ellipsis`

mpRegalado (changes requested):

public/components/wz-menu/wz-menu.js
eze9252: It is a very difficult case to happen (that the api name is so long) but I will consider it.

mpRegalado (approved):
CR: LGTM :heavy_check_mark:

matiasmoreno876 (approved):
CR: LGTM!

Issue #3457, opened 2021-07-09
Wrong body parse in PUT /active-response endpoint using Dev Tools


vicferpoy:
| Wazuh |
| ----- |
| 4.1.5  |

**Description**
`PUT /active-response` JSON body does not match specification when using `Dev Tools`.

**Steps to reproduce**

1. Go to Dev Tools (Kibana).
2. Try to use the `PUT /active-response` endpoint adding `arguments` to the JSON body.
3. See how the error reported in the response does not match.
4. Check the `api.log` and confirm that the JSON is different from the one introduced in the `Dev Tools`.

**Screenshots**
1. `Dev Tools` using a **list** of arguments.
![image](https://user-images.githubusercontent.com/37274979/125044606-67725c00-e09c-11eb-91f0-6df8253f9ae0.png)

2. `api.log` showing that the Wazuh API received a **string** instead of the list.
![image](https://user-images.githubusercontent.com/37274979/125044708-85d85780-e09c-11eb-83ec-c1f2f04253a3.png)

**Additional context**

This was tested using the `wazuh-docker` environment with `v4.1.5`.

Regards,
Víctor

PR #3466, opened 2021-07-13
Wrong active-response endpoint parameter from dev tools

sortizowlh:
Hi team!
This PR solves an error when trying to send an array to an `/active-response` path.
Before the change, the request retrieved the next error with this example:

EXAMPLE:
`PUT /active-response?agents_list=000
{
	"arguments": [
    	"-","1.1.1.1"
    ],
	"command":"firewall-drop.sh",
	"custom": true
}`
ERROR:
`error: 2021/07/12 15:32:7: wazuh-api:makeRequest: {"title":"Bad Request","detail":"'-,1.1.1.1' is not of type 'array' - 'arguments'"}`

Now, the same example should not show this error.

To test it: 

1. Go to Management/Groups

- **Given** the browser is in wazuh app
- **When** the user navigates to Tools/Api Console
- **Then** the Api Console section will be shown

**Precondition 1**:
**Command**: `PUT /active-response?agents_list=000
{
	"arguments": [
    	"-","1.1.1.1"
    ],
	"command":"firewall-drop.sh",
	"custom": true
}`

2. Run this command

- **Given** the browser is Api Console
- **When** the user will run the command according to the **Precondition 1**
- **Then** the return that will be shown after the command ran will be different to: `error: 2021/07/12 15:32:7: wazuh-api:makeRequest: {"title":"Bad Request","detail":"'-,1.1.1.1' is not of type 'array' - 'arguments'"}`

Closes: #3457


gabiwassan (dismissed):

Desvelao (approved):
**review**:
Code Review ✅

Desvelao:
### Changes
- Added the change to changelog

mpRegalado (approved):
CR: LGTM :heavy_check_mark:

PR #3495, opened 2021-07-21
Agg Table component and useEsSearch hook

mpRegalado:
Closes https://github.com/wazuh/wazuh-kibana-app/issues/3476
**This PR adds a hook to make searches to the Elasticsearch database and consume the raw data received.**

It also creates a component that consumes this hook to create a table with an aggregation that shows the available, although it is not implemented in any component yet.

In order to see the component live:

1. Checkout this branch
2. Apply this patch
[aggTable.patch.txt](https://github.com/wazuh/wazuh-kibana-app/files/6854787/aggTable.patch.txt)
3. Visit `Modules/Office 365/Panel`

image


public/components/common/hooks/useEsSearch.js
```diff
@@ -0,0 +1,70 @@
+import { useState, useEffect } from 'react';
```
frankeros, line 1: the name of the files must be in kebab-case `use-es-search` and all new files must be in typescript `ts` or `tsx`

public/components/common/hooks/useEsSearch.js
```diff
@@ -0,0 +1,70 @@
...
+  };
+
+  const nextPage = () => {
+    setPage(page + 1);
+  };
+  const prevPage = () => {
+    setPage(page - 1 < 0 ? 0 : page - 1);
+  };
+
+  return {
```
frankeros, line 62: it should return an array instead of an object to keep standard

public/components/common/panels/agg_table.js
```diff
@@ -0,0 +1,54 @@
+import { EuiBasicTable, EuiPanel, EuiTitle } from '@elastic/eui';
```
frankeros, line 1: kebab-case for file names

gabiwassan (approved):
LGTM

PR #3513, opened 2021-07-29
Moved filter subscription to useFilterManager

mpRegalado:
The filter manager now handles it's own subscriptions and changes it's own state when filters update, returning an array of filters used globally.

This PR also refactors files that already consumed `useFilterManager` so they use the property `filterManager` of the return object

closes https://github.com/wazuh/wazuh-kibana-app/issues/3512

### DIFF
--- a/public/components/common/hooks/use-filter-manager.ts
+++ b/public/components/common/hooks/use-filter-manager.ts
@@ -16,17 +16,15 @@ import _ from 'lodash';
 
 export const useFilterManager = () => {
   const filterManager = useMemo(() => getDataPlugin().query.filterManager, []);
-  const [filters, setFilters] = useState<Filter[]>([])
+  const [filters, setFilters] = useState<Filter[]>([]);
 
   useEffect(() => {
-    let filterSubscriber = filterManager.getUpdates$().subscribe(() => {
+    const { unsubscribe } = filterManager.getUpdates$().subscribe(() => {
       const newFilters = filterManager.getFilters();
       if (!_.isEqual(filters, newFilters)) {
         setFilters(newFilters);
       }
-      return () => {
-        filterSubscriber.unsubscribe();
-      };
+      return unsubscribe;
     });
   }, []);
   return { filterManager, filters };
--- a/public/components/common/hooks/use-filter-manager.ts
+++ b/public/components/common/hooks/use-filter-manager.ts
@@ -24,8 +24,8 @@ export const useFilterManager = () => {
       if (!_.isEqual(filters, newFilters)) {
         setFilters(newFilters);
       }
-      return unsubscribe;
     });
+    return unsubscribe;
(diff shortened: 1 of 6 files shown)


public/components/common/hooks/use-filter-manager.ts
```diff
@@ -10,10 +10,24 @@
...
+import { Filter } from 'src/plugins/data/public';
+import _ from 'lodash';
 
 export const useFilterManager = () => {
-    const [filterManager, setFilterManager] = useState(getDataPlugin().query.filterManager);
-    return filterManager;
-}
\ No newline at end of file
+  const filterManager = useMemo(() => getDataPlugin().query.filterManager, []);
+  const [filters, setFilters] = useState<Filter[]>([])
```
Desvelao, line 19: **thoughts**: Why the initial value of `filters` state is an empty array `[]`. Should we use the `filterManager.getFilters()` as the initial value?

public/components/common/hooks/use-filter-manager.ts
```diff
@@ -10,10 +10,24 @@
...
+
+  useEffect(() => {
+    let filterSubscriber = filterManager.getUpdates$().subscribe(() => {
+      const newFilters = filterManager.getFilters();
+      if (!_.isEqual(filters, newFilters)) {
+        setFilters(newFilters);
+      }
+      return () => {
+        filterSubscriber.unsubscribe();
+      };
```
Desvelao, line 29: **issue**: The effect function of `useEffect` should return a function that will be called when the component will unmount to unsubscribe.

**suggestion**: Fix the returned function
```ts
useEffect(() => {
  // Suggestion change: destructuring the `unsubscribe` method
  const { unsubscribe } = filterManager.getUpdates$().subscribe(() => {
    const newFilters = filterManager.getFilters();
    if (!_.isEqual(filters, newFilters)) {
      setFilters(newFilters);
    }
  }
  // The returned function that will be called when the component will unmount is returned from the effect function. In the current code, a function is returned from the subscription callback function.
  return unsubscribe
}

public/components/common/hooks/use-filter-manager.ts
mpRegalado: Since the call for filters is asynchronous we couldn't use it as an innitial value. We could create an explicit `useEffect` that calls for the filters to set them, but in my testing this doesn't seem to be necessary.

public/components/common/hooks/use-filter-manager.ts
mpRegalado: It seems functionally the same but I do like the elegance of it. I'll commit it right away.

asteriscos (approved):
CR: :heavy_check_mark:

Desvelao (approved):
**review**:
Code review ✅

PR #3517, opened 2021-07-30
Enhancement/3512 refactor use filter manager

mpRegalado:
This PR is the same as https://github.com/wazuh/wazuh-kibana-app/pull/3513 but fixing a critical bug that was reverted.

The filter manager now handles it's own subscriptions and changes it's own state when filters update, returning an array of filters used globally.

This PR also refactors files that already consumed useFilterManager so they use the property filterManager of the return object
closes https://github.com/wazuh/wazuh-kibana-app/issues/3512


asteriscos (approved):
CR: :heavy_check_mark: 
Test: :heavy_check_mark:

eze9252 (approved):
LGTM!

PR #3529, opened 2021-08-03
change filter from is to is one of

eze9252:
Hi team, this resolve: 

- change custom search bar from is to is one of to support multiple values

closes #3529

### DIFF
--- a/public/components/common/custom-search-bar/custom-search-bar.tsx
+++ b/public/components/common/custom-search-bar/custom-search-bar.tsx
@@ -20,7 +20,6 @@ export const CustomSearchBar = ({ filtersValues, ...props }) => {
 
     const KibanaServices = getDataPlugin().query;
     const filterManager = KibanaServices.filterManager;
-    const indexPattern = getIndexPattern();
     const defaultSelectedOptions = () => {
         const array = []
         filtersValues.forEach(item => {
@@ -51,15 +50,13 @@ export const CustomSearchBar = ({ filtersValues, ...props }) => {
 
 
     const buildCustomFilter = (isPinned: boolean, values?: any): Filter => {
-        const newFilters = []
-        values.forEach(element => {
-            const filter = {
-                match_phrase: {
-                    [element.value]: element.label
-                }
+        const newFilters = values.map(element => ({
+            match_phrase: {
+              [element.value]: {
+                query: element.label
+              }
             }
-            newFilters.push(filter);
-        });
+        }));
         const params = values.map(item => item.label)
         const meta: FilterMeta = {
             disabled: false,
@@ -94,13 +91,11 @@ export const CustomSearchBar = ({ filtersValues, ...props }) => {
 
     const refreshCustomSelectedFilter = () => {
         setSelectedOptions(defaultSelectedOptions)
-        const filterCustom = []
         const filters = filterManager.getFilters().filter(item => item.meta.type === 'phrases' || Object.keys(selectedOptions).includes(item.meta.key)).map(element => ({ params: element.meta.params, key: element.meta.key })) || [];
-        filters.forEach(item => {
-            item.params.forEach(element => {
-                filterCustom.push({ label: element, value: item.key })
-            })
-        })
+        const getFilterCustom = (item) => {
+             return item.params.map(element => ({ label: element, value: item.key }))
+        }
+        const filterCustom = filters.map((item) => getFilterCustom(item))[0] || []
 
         if (filterCustom.length != 0) {
             filterCustom.forEach(item => {
--- a/public/components/common/custom-search-bar/custom-search-bar.tsx
+++ b/public/components/common/custom-search-bar/custom-search-bar.tsx
@@ -105,7 +105,6 @@ export const CustomSearchBar = ({ filtersValues, ...props }) => {
       return item.params.map((element) => ({ label: element, value: item.key }));
     };
     const filterCustom = filters.map((item) => getFilterCustom(item)) || [];
-    console.log(filters,filters.map((item) => getFilterCustom(item)))
     if (filterCustom.length != 0) {
       filterCustom.forEach((item) => {
         item.forEach(element =>{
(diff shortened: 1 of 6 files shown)


public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -49,51 +50,57 @@ export const CustomSearchBar = ({ filtersValues, ...props }) => {
...
+    const buildCustomFilter = (isPinned: boolean, values?: any): Filter => {
+        const newFilters = []
+        values.forEach(element => {
+            const filter = {
+                match_phrase: {
+                    [element.value]: element.label
+                }
+            }
+            newFilters.push(filter);
+        });
```
Desvelao, line 62: **nitpick**: You could use `Array.prototype.map` instead of `Array.prototype.forEach`;

```ts
const newFilters = values.map(element => {
  const filter = {
    match_phrase: {
        [element.value]: element.label
    }
  }
});
```

The `match_phrase` give problem with some app logic. Use
```
match: {
  [element.value]: {
    query: element.value,
    type: 'phrase'
  }
}
```

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -49,51 +50,57 @@ export const CustomSearchBar = ({ filtersValues, ...props }) => {
...
-                value: item.meta.key,
-                label: item.meta.params.query,
-            }
-        }).filter(element => Object.keys(selectedOptions).includes(element.value))
+        const filterCustom = []
+        const filters = filterManager.getFilters().filter(item => item.meta.type === 'phrases' || Object.keys(selectedOptions).includes(item.meta.key)).map(element => ({ params: element.meta.params, key: element.meta.key })) || [];
+        filters.forEach(item => {
+            item.params.forEach(element => {
+                filterCustom.push({ label: element, value: item.key })
+            })
```
Desvelao, line 102: **nitpick**: We could use `Array.prototype.map` instead of `Array.protype.forEach` + `Array.prototype.push`

CPAlejandro (changes requested):
The changelog is missing

CPAlejandro (approved):
LGTM!

Machi3mfl (approved):
LGTM!

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -1,165 +1,178 @@
...
+  const [selectedOptions, setSelectedOptions] = useState(defaultSelectedOptions);
+
+  useEffect(() => {
+    let filterSubscriber = filterManager.getUpdates$().subscribe(() => {
+      onFiltersUpdated();
+      return () => {
+        filterSubscriber.unsubscribe();
+      };
+    });
+  }, []);
```
Desvelao, line 41: **nitpick**: The returned function to unsubscribe to the changes on the `filterManager` should be the returned on the `useEffect` callback, not the subscription function.

**suggestion**: You could use the `useFilterManager` hook instead of use the `filterManager.getUpdates$().subscribe` and listen to the chagens of `filters` variable returned by the mentioned hook to apply the effect.

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -1,165 +1,177 @@
...
\ No newline at end of file
+  };
+
+  const onChange = (values: any[]) => {
+    setKibanaFilters(values);
+    refreshCustomSelectedFilter();
+  };
+
+  const getComponent = (item: any) => {
+    var types = {
```
Desvelao, line 126: **nitpick**: You could use `const` instead of `var` here. Anyways, we prefer `let` over `var` in the variable declaration.

Desvelao:
**issue**: We should be able to remove the filters applied to the selectors. The button to remove the filter is hidden with CSS
image
image

CPAlejandro (approved):
The second revision of CR and testing: LGTM!

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -1,165 +1,182 @@
...
-        setAvancedFiltersState(state => !state);
+    return { meta, $state, query };
+  };
+
+  const setKibanaFilters = (values: any[], selectReference: String) => {
+    const currentFilters = filterManager
+      .getFilters()
+      .filter((item) => item.meta.key != selectReference);
+    filterManager.removeAll();
+    filterManager.addFilters(currentFilters);
```
mpRegalado, line 90: Why remove and add filters?

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -1,165 +1,182 @@
...
+  const defaultSelectedOptions = () => {
+    const array = [];
+    filtersValues.forEach((item) => {
+      array[item.key] = [];
+    });
+
+    return array;
+  };
+  const [avancedFiltersState, setAvancedFiltersState] = useState(false);
+  const [selectedOptions, setSelectedOptions] = useState(defaultSelectedOptions);
```
mpRegalado, line 29: Over the code, setSelectedOptions is used to set references to functions rather than running the functions and setting the state as their return, but the variable selectedOptions is used with the structure of an array.

public/components/common/hooks/use_previous.ts
```diff
@@ -0,0 +1,9 @@
+import { useEffect, useRef } from 'react';
+
+export function usePrevious(value) {
```
mpRegalado, line 3: How does this hook work, when does the value get updated?

sortizowlh (approved):
LGTM!

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -1,165 +1,182 @@
...
+  const [values, setValues] = useState([]);
+  const [selectReference, setSelectReference] = useState('');
+  const prevValues = usePrevious(values);
+
+  useEffect(() => {
+    if (prevValues != values) {
```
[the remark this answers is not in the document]
eze9252: done!

PR #3531, opened 2021-08-03
Integrating dynamic filters

### LINKED ISSUE #3510: Filter suggested values by the already active filters
**Describe the solution you'd like**
The `useValueSuggestions` hook currently returns possible values for the provided field regardless of whether they would offer any results given the current search status. It could be useful to limit the recommendations based on the already applied filters and timeframe.

**Additional context**
The `autocomplete.getValueSuggestions` accepts an optional paramter called `boolFilter` of type `any[]`, this value is passed on to the `/api/kibana/suggestions/values/${index}` endpoint and, from experimentation, seems to require an array of objects with similar structure to the one outilined in the [elasticsearch filter context](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-filter-context.html)

It seems like the best course of action to guarantee that the `useValueSuggestions` hook would be to use the the filters provided by the `filterManager` and process them in a way that can be used in the previous endpoint.

CPAlejandro:
Hi guys,
In this PR we integrate the dynamic filters and add the filter fields from the GitHub module.

To test it:
We will go to the Github tab, try to enter the tables, and see if the filters are applied correctly when entering the Drilldown tables and if when applying filters both in Github and in the Drilldown tables, the simple filters are modified as well.

Closes #3510

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 - Created a separate component to check for sample data [#3475](https://github.com/wazuh/wazuh-kibana-app/pull/3475)
 - Added a new hook for getting value suggestions [#3506](https://github.com/wazuh/wazuh-kibana-app/pull/3506)
 - Added base Module Panel view with Office365 setup [3518](https://github.com/wazuh/wazuh-kibana-app/pull/3518)
+- Added dinamic simple filters and adding simple GitHub filters fields [3531](https://github.com/wazuh/wazuh-kibana-app/pull/3531)
 
 ### Changed
 
--- a/public/components/overview/github-panel/config/search-bar-config.ts
+++ b/public/components/overview/github-panel/config/search-bar-config.ts
@@ -1,70 +1,33 @@
+/*
+ *   Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License").
+ *   You may not use this file except in compliance with the License.
+ *   A copy of the License is located at
+ *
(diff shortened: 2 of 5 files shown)


public/components/overview/github-panel/config/search-bar-config.ts
```diff
@@ -1,70 +1,33 @@
...
+    type: 'combobox',
+    key: 'data.github.actor',
+  },
+  {
+    type: 'combobox',
+    key: 'data.github.org',
+  },
+  {
+    type: 'combobox',
+    key: 'data.github.repository',
```
Desvelao, line 27: **nitpick**: Change the field to `data.github.repo`.

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 - Created a separate component to check for sample data [#3475](https://github.com/wazuh/wazuh-kibana-app/pull/3475)
 - Added a new hook for getting value suggestions [#3506](https://github.com/wazuh/wazuh-kibana-app/pull/3506)
 - Added base Module Panel view with Office365 setup [3518](https://github.com/wazuh/wazuh-kibana-app/pull/3518)
+- Added dinamic simple filters and adding simple GitHub filters fields [3531](https://github.com/wazuh/wazuh-kibana-app/pull/3531)
```
Desvelao, line 49: **thought**: I am not sure if we want to add this to the changelog.

Desvelao (approved):
**review**:
Code review ✅ 
Testing ✅

public/components/overview/github-panel/config/search-bar-config.ts
Desvelao: **action**:
- Replaced the field by `data.github.repo`

Issue #3532, opened 2021-08-03
Specify filters for Office365 search bar


mpRegalado:
We should decide which filters exactly will be shown in the office365 module.

These filter selectors are defined in `public/components/overview/office-panel/config/search-bar-config.ts`
A few valid options could be:

- Filter by subscriptions, the values of `data.office365.Subscription`
- Filter by userType, the values of `data.office365.UserType`, they should be mapped from their numeric values to strings according [to this reference](https://docs.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema#user-type)
- Filter by ResultStatus, the values of `data.office365.ResultStatus`

More possible fields can be found in the [office365 documentation](https://docs.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema#common-schema), or might be related to alert data such as the level.

PR #3533, opened 2021-08-03
Specify filters for Office365 search bar

gabiwassan:
Hi team,
This PR added specifics filters for Office 365 search bar.
The `optionsTitles` parameter is added in order to be able to give a mask to values that number, as in the case of `UserTypes`
Added `placeholder` value on `filtersValues` to assign a simple label on combobox

Added new component multis-select implemented with `EuiFilterGroup` `EuiFilterSelectItem` `EuiFieldSearch`

image

Close: #3532


public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -112,15 +112,15 @@ export const CustomSearchBar = ({ filtersValues, ...props }) => {
...
-            'default': <></>,
-            'combobox': <Combobox
-                item={item}
-                selectedOptions={selectedOptions[item.key] || []}
-                onChange={onChange}
-            />
-        };
-        return types[item.type] || types['default'];
+      const types = {
+        'default': <></>,
```
frankeros, line 116: it should be typed and should not use strings for keys
similar to 
```ts
const types: {[key: string]: object} = {
		default: <></>,
		combobox: <Combobox>...</Combobox>
};

return types[item.type] || types.default;

}

```

public/components/common/hooks/use-value-suggestions.ts
```diff
@@ -29,13 +28,28 @@ export interface IValueSuggestiions {
...
+  type: 'string' | 'boolean' = 'string'
+): IValueSuggestiions => {
   const [suggestedValues, setSuggestedValues] = useState<string[] | boolean[]>([]);
   const [query, setQuery] = useState<string>('');
   const [isLoading, setIsLoading] = useState(true);
   const data = getDataPlugin();
   const indexPattern = useIndexPattern();
 
+  const getValueSuggestions = async (field) => {
+    if (field.name === 'data.office365.UserType') {
```
frankeros, line 42: this hook is generic should not know about `'data.office365.UserType'`

frankeros (approved):
LGTM!

public/components/overview/office-panel/config/search-bar-config.ts
```diff
@@ -11,17 +11,36 @@
...
     type: 'combobox',
-    key: 'agent.name',
+    key: 'data.office365.Subscription',
+    placeholder: 'Subscription',
   },
   {
     type: 'combobox',
-    key: 'agent.ip',
+    key: 'data.office365.UserType',
+    placeholder: 'UserType',
```
Desvelao, line 33: **issue**: The `placeholder` should be readable by the user.

**suggestion**: Maybe we could use `User type`.

public/components/overview/office-panel/config/search-bar-config.ts
```diff
@@ -11,17 +11,36 @@
...
+    optionTitle: true,
+  },
+  {
+    type: 'combobox',
+    key: 'data.office365.ResultStatus',
+    placeholder: 'ResultStatus',
```
Desvelao, line 39: **issue**: The `placeholder` should be readable by the user.

**suggestion**: Maybe we could use `Result status`. I am not sure if the `data.office365.ResultStatus` field could have a more comprehensible name.

public/components/common/hooks/useValueSuggestion/use-value-suggestion.ts
```diff
@@ -9,33 +9,51 @@
...
+  const getValueSuggestions = async (field) => {
+    const result = await data.autocomplete.getValueSuggestions({
+      query,
+      indexPattern: indexPattern as IIndexPattern,
+      field,
+    });
+
+    if (!result.length) {
+      return getCustomValueSuggestion(field);
+    }
```
Desvelao, line 52: **thought**: I am not sure what is this doing. does it get fixed values? It is getting some prefixed values, if this is the case, maybe we could let the component receive these as props. We should discuss the solution.

Desvelao:
**issue**: It looks a little weird to use a group with a value on this selector.
image

**suggestion**: Maybe we could display a text in the filter with a different value. We could research about the `value` property of `EuiComboBox` options. More information: https://elastic.github.io/eui/#/forms/combo-box#EuiComboBoxOptionOption

sortizowlh:
usertype_office


It seems that it's only selectable the number in the combobox options instead of the whole button.

Also, when a UserType is selected, the filters shows that a UserType has been selected, but don't show the number or name. Is only shown  a Warning message.

When more than one UserType is selected, is not possible to know what filter is for what UserType.

And, when a UserType is selected, on the right side of the number appears an arrow, but only refreshes the page.

sortizowlh (approved):
CR: LGTM

public/components/common/hooks/useValueSuggestion/use-value-suggestion.ts
gabiwassan: If the search does not obtain results, use the new helper that I have created to obtain them (the idea is to separate that responsibility from the hook) the helper can be called here or before calling the hook, if we could define it better 👍🏼

gabiwassan:
> **issue**: It looks a little weird to use a group with a value on this selector.
> image
> 
> **suggestion**: Maybe we could display a text in the filter with a different value. We could research the `value` property of `EuiComboBox` options. More information: https://elastic.github.io/eui/#/forms/combo-box#EuiComboBoxOptionOption

Yes is so weird. 
Right there I am using EuiComboBoxOptionOption, which generates the groups, it is an alternative, perhaps the ideal would be to use another component for these cases that we do not have values, only numeric keys.

gabiwassan:
> usertype_office
> 
> It seems that it's only selectable the number in the combobox options instead of the whole button.
> 
> Also, when a UserType is selected, the filters shows that a UserType has been selected, but don't show the number or name. Is only shown a Warning message.
> 
> When more than one UserType is selected, is not possible to know what filter is for what UserType.
> 
> And, when a UserType is selected, on the right side of the number appears an arrow, but only refreshes the page.

It generates a warning because you do not have the data. You could install the SampleData for it and try again. Maybe we could handle alerts better by not having data 👍🏼

eze9252 (approved):
LGTM!

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -1,165 +1,190 @@
-import React, { useState, useEffect, useLayoutEffect } from 'react';
-
-import { getIndexPattern } from '../../overview/mitre/lib'
+import React, { useEffect, useState } from 'react';
+import { getIndexPattern } from '../../overview/mitre/lib';
```
mpRegalado, line 2: Why get the index pattern from Mitre instead of `useIndexPattern`?

public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -1,165 +1,190 @@
...
+  const [selectedOptions, setSelectedOptions] = useState(defaultSelectedOptions);
+
+  useEffect(() => {
+    let filterSubscriber = filterManager.getUpdates$().subscribe(() => {
+      onFiltersUpdated();
+      return () => {
+        filterSubscriber.unsubscribe();
+      };
+    });
+  }, []);
```
mpRegalado, line 36: This `useEffect` does not properly cancel the subscription as the return statement is the subscriber's rather than `useEffect`'s
The hook `useFilterManager` already manages the subscription

public/components/common/custom-search-bar/components/multi-select.tsx
```diff
@@ -0,0 +1,157 @@
...
+            key: key,
+            label: value,
+            value: item.key,
+            filterByKey: item.filterByKey,
+            checked: OFF as FilterChecked,
+          }))
+          .sort((a, b) => a.label - b.label)
+      );
+    }
+  }, [suggestedValues, isLoading, setQuery]);
```
mpRegalado, line 56: Same question about `setQuery`, why as a dependency

public/components/common/custom-search-bar/components/multi-select.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  useEffect(() => {
+    items.forEach((item) => {
+      item.checked = selectedOptions.find((element) => element.label === filterBy(item)) ? ON : OFF;
```
mpRegalado, line 60: `items` is an state object returned by `useState` and should be immutable. Switching the `checked` value directly goes against this as it doesn't make React aware of the changes to the object and may cause unexpected issues.

public/components/common/custom-search-bar/components/multi-select.tsx
```diff
@@ -0,0 +1,157 @@
...
+    });
+    setActiveFilters(selectedOptions.length);
+  }, [selectedOptions]);
+
+  const filterBy = (item) => {
+    return item.filterByKey ? item.key.toString() : item.label;
+  };
+
+  const toggleFilter = (item) => {
+    item.checked = item.checked === ON ? OFF : ON;
```
mpRegalado, line 70: Is there a reason to need to use this logic rather than have `item.checked` be a boolean?

mpRegalado (changes requested):

public/components/common/custom-search-bar/components/multi-select.tsx
gabiwassan: Yes, unfortunately that definition is from EuiCompontent 😢 
image

public/components/common/custom-search-bar/components/multi-select.tsx
gabiwassan: Is not! thanks

mpRegalado (approved):
CR: LGTM :heavy_check_mark:

public/components/common/custom-search-bar/components/multi-select.tsx
```diff
@@ -0,0 +1,160 @@
...
+  EuiPopoverTitle,
+  EuiSpacer,
+  FilterChecked,
+} from '@elastic/eui';
+import { IValueSuggestion, useValueSuggestion } from '../../hooks';
+
+const ON = 'on';
+const OFF = 'off';
+
+export const MultiSelect = ({ item, onChange, selectedOptions, onRemove }) => {
```
mpRegalado, line 31: This component should have a similar behaviour as the one implemented to the Combobox in [this PR](https://github.com/wazuh/wazuh-kibana-app/pull/3529).
Changing the component used in the search bar from the Combobox to MultiSelect may have been outside the scope of the issue, but if we are keeping this decision we should make sure it keeps an expected behavior, and that means that selecting several options should return values that match any of the selected rather than all of them at once, because any 2 options in a single field are mutually exclusive.

mpRegalado (changes requested):
- `Actor ID` is not guaranteed to exist, it belongs to the [Azure Active Directory Schema](https://docs.microsoft.com/en-us/office/office-365-management-api/office-365-management-activity-api-schema#azure-active-directory-schema) and thus only will be present on alerts that use that schema. I personally don't see enough merit to having it as a featured filter, but if it were, the component would need an additional option to filter items that do not have that property.
- `Client IP` is already defined in one of the aggregation tables. Having it also as a featured filter is redundant and any selection will lead the aggregation table to only have 1 result.

frankeros (approved):
CR: LGTM!
Testing: Passed

Comment: 
I found a bug whit pinned filters across the Office 365 panel, I'm creating a new issue with this
pin filter no appearing in panel

Machi3mfl (approved):
CR ✅
Testing ✅

PR #3534, opened 2021-08-04
Pagination in Drilldown tables

CPAlejandro:
Hi guys,

In this PR we switch from basic tables to memory tables to be able to show the paginated tables.

image

To test it we will only have to get into the GitHub module and go to the Panel tab

### DIFF
--- a/public/components/common/modules/panel/components/agg-table.tsx
+++ b/public/components/common/modules/panel/components/agg-table.tsx
@@ -62,6 +62,10 @@ export const AggTable = ({
       direction: 'desc',
     },
   };
+  const pagination = {
+    hidePerPageOptions: true,
+    pageSize: 10,
+  };
   return (
     <EuiPanel data-test-subj={`${aggTerm}-aggTable`} {...panelProps}>
       <EuiTitle {...titleProps}>
@@ -73,7 +77,7 @@ export const AggTable = ({
         loading={isLoading}
         rowProps={getRowProps}
         error={error ? error.message : undefined}
-        pagination={{ pageSizeOptions: [5, 10, 20] }}
+        pagination={pagination}
         sorting={sorting}
       />
     </EuiPanel>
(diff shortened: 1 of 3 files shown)


public/components/overview/github-panel/config/main.tsx
```diff
@@ -28,7 +28,7 @@ export const MainViewConfig = {
                 tableTitle='Actors'
                 aggTerm='data.github.actor'
                 aggLabel='Actor'
-                maxRows={5} 
+                maxRows={100}
```
Desvelao, line 31: **thought**: I am not sure why this change.

sortizowlh (approved):
CR: 🟢 
Test: 🟢

CPAlejandro:
**Final result:**

image

public/components/overview/github-panel/config/main.tsx
CPAlejandro: I have modified it by a variable, to extract the same data from the 4 fields. Thanks for the warning.

public/components/common/modules/panel/components/agg-table.tsx
```diff
@@ -56,18 +56,29 @@ export const AggTable = ({
       },
     };
   };
-
+  const sorting = {
```
mpRegalado, line 59: Adding sorting to an EuiInMemoryTable without knowing that all the possible data is already in memory can lead to unaccurate results.
You could use `onTableChange` as [documented by Eui](https://elastic.github.io/eui/#/tabular-content/in-memory-tables) to change the value of `preAppliedAggs.buckets.terms.order` for either `_count` or `_key` to be `asc` or `desc` as described in the [Elastic doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-order). It would just require mapping the `field` value to one of those.

mpRegalado (changes requested):

sortizowlh:
Tested all new changes ✔️ 
Now GitHub panel tab has 10 rows per table.

sortizowlh:
Second revision. LGTM!
Now, all tables have 10 rows for GitHub Panel tab

PR #3535, opened 2021-08-04
Changed props for Discover

mpRegalado:
This pr closes https://github.com/wazuh/wazuh-kibana-app/issues/3527

On this PR, I have refactored the properties used in the component `Discover` and all the components that use it.

[The first commit](https://github.com/wazuh/wazuh-kibana-app/commit/02f941ef98fb8ac42939fcc433e4fce4ab11d723) simply applies `prettier` to the files that will be modified.

[The second commit](https://github.com/wazuh/wazuh-kibana-app/commit/5c5287e68a3194eaf8b4e673cc50e2af9169c329) changes the behavior of Discover. A brief summary of the changes as intended:

- A new interface is defined within the component:
```
interface ColumnDefinition {
  field: string;
  label?: string;
}
```

- The prop `initialColumns` is set as an array of `ColumnDefinition` , the prop `initialAgentColumns` is created as optional to be used when an agent is pinned.
- The method `getColumns` of `Discover` now extracts the selected fields from one of these properties, if `initialAgentColumns` is not defined it should always use `initialColumns` regadless of whether there is an agent pinned.
- The names for the columns are extracted from the `label` field in `ColumnDefinition`, defaults to `field` if not specificed
- All files that used the property initialColumns have been edited to use this new structure.

### DIFF
--- a/public/components/agents/fim/inventory/fileDetail.tsx
+++ b/public/components/agents/fim/inventory/fileDetail.tsx
@@ -479,12 +479,22 @@ export class FileDetails extends Component {
                 kbnSearchBar
                 shareFilterManager={this.discoverFilterManager}
                 initialColumns={[
-                  'icon',
-                  'timestamp',
-                  'syscheck.event',
-                  'rule.description',
-                  'rule.level',
-                  'rule.id',
+                  { field: 'icon' },
+                  { field: 'timestamp' },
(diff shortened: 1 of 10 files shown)


matiasmoreno876 (approved):
CR: LGTM ✅

eze9252 (approved):
LGTM!

PR #3551, opened 2021-08-10
Removed office365 section from agent sections

mpRegalado:
This PR closes https://github.com/wazuh/wazuh-kibana-app/issues/3550

The office365 section should not appear under the agents sections as it is not applicable. This PR removes it.
To test:

1. Visit an agent panel
2. Click on more
3. See that Office365 is missing
image


CPAlejandro (approved):
CR:🟢 
Testing:🟢 

I have tested with and without activating the module and all seems correct to me!

eze9252 (approved):
LGTM!

PR #3562, opened 2021-08-11
block filter when drilldown is open

eze9252:
Hi team, this resolve:

- When you enter a specific drilldown, filter should be blocked to prevent you from choosing another option-

### DIFF
--- a/public/components/common/custom-search-bar/components/multi-select.tsx
+++ b/public/components/common/custom-search-bar/components/multi-select.tsx
@@ -38,7 +38,6 @@ export const MultiSelect = ({ item, onChange, selectedOptions, onRemove, isDisab
     { key: any; label: any; value: any; checked: FilterChecked }[]
   >([]);
   const [activeFilters, setActiveFilters] = useState<number>(0);
-  const disableStyles = {pointerEvents: isDisabled ? "none" : "all"}
 
   useEffect(() => {
     if (!isLoading) {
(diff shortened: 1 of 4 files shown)


public/components/common/custom-search-bar/custom-search-bar.tsx
```diff
@@ -18,7 +18,7 @@ import { getCustomValueSuggestion } from '../../../components/overview/office-pa
 
 export const CustomSearchBar = ({ filtersValues, ...props }) => {
   const { filterManager, filters } = useFilterManager();
-  const indexPattern = useIndexPattern();
+  const filterDrillDownValue = sessionStorage.getItem('filterDrillDownValue')
```
Desvelao, line 21: **issue**: Could we get a solution without using the `sessionStorage`?

**suggestion**: Maybe we should use some component `prop` to control when the selector is disabled.

public/components/common/custom-search-bar/components/multi-select.tsx
```diff
@@ -28,7 +28,7 @@ import { IValueSuggestion, useValueSuggestion } from '../../hooks';
 const ON = 'on';
 const OFF = 'off';
 
-export const MultiSelect = ({ item, onChange, selectedOptions, onRemove }) => {
+export const MultiSelect = ({ item, onChange, selectedOptions, onRemove, isDisable }) => {
```
mpRegalado, line 31: Nitpick: gramatically is more correct to use `isDisabled` rather than `isDisable`

Machi3mfl (approved):
Test y CR ✅

frankeros:
disabled the entire selector when the multi-select is disabled
image

frankeros (approved):
LGTM!

PR #3568, opened 2021-08-18
Simple filters change between panel and drilldown panel

### LINKED ISSUE #3563: Simple filters options doesn't change between Panel and in Drilldown tables
Hi guys,
The options that appear in the simple filters selecting an option in Github or Office Panel tables aren´t changing from the Panel ones.

CPAlejandro:
Hi guys,

In this PR we try to change the simple filters between the vies of Panel and the view of Drilldown Panel (selecting a row in Panel)

To test it, go to GitHub Panel, see the Simple filters allocated there, then select a row and check if the filters have changed. Of course, only some of these rows have to change. Do more than 1 attempt.

Example:
- Panel view:
image

- Drilldown view:
image

Closes #3563

### DIFF
--- a/public/components/common/hooks/use-value-suggestion.ts
+++ b/public/components/common/hooks/use-value-suggestion.ts
@@ -36,7 +36,7 @@ interface BoolFilter {
 
 export const useValueSuggestion = (
   filterField: string,
-  filterDrillDownValue: BoolFilter,
+  boolFilterValue: BoolFilter,
   options?: string[],
   type: 'string' | 'boolean' = 'string'
 ): IValueSuggestion => {
@@ -52,11 +52,11 @@ export const useValueSuggestion = (
 
   const getValueSuggestions = async (field) => {
     const boolFilter =
-      filterDrillDownValue && filterDrillDownValue.value !== ""
+      boolFilterValue && boolFilterValue.value !== ""
         ? [
             {
               term: {
-                [filterDrillDownValue.field]: `${filterDrillDownValue.value}`,
+                [boolFilterValue.field]: `${boolFilterValue.value}`,
               },
             },
           ]
@@ -99,7 +99,7 @@ export const useValueSuggestion = (
         }
       })();
     }
-  }, [indexPattern, query, filterField, type, filterDrillDownValue]);
+  }, [indexPattern, query, filterField, type, boolFilterValue]);
 
   return { suggestedValues, isLoading, setQuery };
 };
--- a/public/components/common/hooks/use-value-suggestion.ts
+++ b/public/components/common/hooks/use-value-suggestion.ts
@@ -30,13 +30,16 @@ export interface IValueSuggestion {
 }
 
 interface BoolFilter {
-  field:string;
-  value:string;
+  field: string;
+  value: string;
 }
 
 export const useValueSuggestion = (
   filterField: string,
-  boolFilterValue: BoolFilter,
+  boolFilterValue: BoolFilter = {
+    field: '',
+    value: '',
+  },
   options?: string[],
   type: 'string' | 'boolean' = 'string'
 ): IValueSuggestion => {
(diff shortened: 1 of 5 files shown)


public/components/common/hooks/use-value-suggestion.ts
```diff
@@ -31,6 +31,7 @@ export interface IValueSuggestion {
 
 export const useValueSuggestion = (
   filterField: string,
+  filterDrillDownValue,
```
mpRegalado, line 34: I am not sure about this name inside the `useValueSuggestion` hook. A hook should concern itself with it's own behavior and not where it is used, it is not a bad idea to accept filters as a parameter but I feel that naming it `DrillDownValue` ties it too closely with the behavior of other components. Maybe replace it by `boolFilters` internally.

This parameter should also have a type. I would suggest
```
interface BoolFilter {
    field:string;
    value:string;
}
export const useValueSuggestion = (
    boolFilters : BoolFilter[] = [],
    ...
```
This snippet of code also adds a default value to this parameter so you can always count on having an array here.

public/components/common/hooks/use-value-suggestion.ts
```diff
@@ -39,19 +40,29 @@ export const useValueSuggestion = (
...
+    const boolFilter =
+      filterDrillDownValue && filterDrillDownValue.value !== ""
+        ? [
+            {
+              term: {
+                [filterDrillDownValue.field]: `${filterDrillDownValue.value}`,
+              },
+            },
+          ]
+        : [];
```
mpRegalado, line 58: If the previous comment was followed, this could be replaced by
```
const boolFilter = boolFilters.map(filter => ( 
    {
        term: {
            [filterDrillDownValue.field]:  filterDrillDownValue.value
        }
    } ) );
```
because we would always know boolFilters is, at least, an empty array. And if it isn't, it will have a key value pair.

public/components/overview/github-panel/github-panel.tsx
```diff
@@ -11,18 +11,23 @@
...
 import { withErrorBoundary } from '../../common/hocs';
 import { CustomSearchBar } from '../../common/custom-search-bar';
 import { ModuleConfiguration } from './views';
 import { ModuleConfig, filtersValues } from './config';
 
 export const GitHubPanel = withErrorBoundary(() => {
+  const [drillDownValue, setDrillDownValue] = useState({ field: '', value: '' });
+  const filterDrillDownValue = (value) => {
+    setDrillDownValue(value)
+  }
```
mpRegalado, line 25: Any reason why the `filterDrillDownValue` should only filter by the filter selected when drilling down instead of the global filter state?

mpRegalado (changes requested):
The only comment I think would block this PR is the type for the new prop in useSuggestedValues. The rest are either suggestions or questions for curiosity's sake.

asteriscos (changes requested):
Test: :x: 
Tested the PR and found this strange behavior:
After selecting different values and clicking one of the table rows with a different field of the selected ones, the combo-box does not show any selected items. The selection counter is ok, but in the items list there's none selected.

Peek 2021-08-18 15-37

public/components/overview/github-panel/github-panel.tsx
CPAlejandro: If we use global filter, we can have some problems distinguishing which filter we can pick to the boolFilter. I think that catch them when you click will be a better option.

mpRegalado (approved):
CR: LGTM :heavy_check_mark:

asteriscos (approved):
Tested :heavy_check_mark:

asteriscos (changes requested):
Test ❌

Tested it on fields which have +500 values and it is not possible to select values different ranges.
Seems the last selection replaces the previous one.

Peek 2021-09-03 18-28

frankeros:
> Test x
> 
> Tested it on fields which have +500 values and it is not possible to select values different ranges.
> Seems the last selection replaces the previous one.
> 
> Peek 2021-09-03 18-28

Please create a new issue with this.

PR #3569, opened 2021-08-18
Erasing windows update option in wrong cases

### LINKED ISSUE #3566: Windows updates in Linux agant inventory
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.x   | 7.x     | - |

**Description**
A section with `Windows updates`  is displayed in reports of a Linux agent inventory.

**Steps to reproduce**

1. Go to a Linux agent
2. Click on the inventory
3. Generate report
4. See the Windows updates section 

**Screenshots**
image

CPAlejandro:
Hi guys,

We disabled Windows Update part of Inventory PDF in agents with an operative system different from Windows.

To test it, go to Agents, click an agent without Windows, go to Inventory Section and generate a PDF. Then go to the end of that and see if Windows Update section appers.

If you can, try as well with a Windows agent to check that in this case, this section appears.

Closes #3566


mpRegalado (approved):
CR: LGTM :heavy_check_mark: 
Testing: LGTM :heavy_check_mark:

server/controllers/wazuh-reporting.ts
```diff
@@ -1906,7 +1906,7 @@ export class WazuhReportingCtrl {
             ]
           }
         },
-        {
+        agentOs === 'windows' && {
```
frankeros, line 1909: this is a bad implementation, is not a good practice using a ternary in an array definition 
please use a different approach, maybe inserting after the declaration of the array with a conditional 

also, check if this endpoint is used for another SO but windows

frankeros (changes requested):

server/controllers/wazuh-reporting.ts
CPAlejandro: I change it using your suggestions. Thanks you!

frankeros (approved):
LGTM!

PR #3598, opened 2021-09-09
Made index pattern selector appear without refreshing page

mpRegalado:
### Previous behaviour
The index pattern selector did not show in the menu after it was created from within the app unless you either refreshed the page, exited and re-entered the app or opened the menu.

### Current behaviour
Upon performing a health check, the wz menu updates its data and displays the selector for index patterns if it is needed. To achieve this, the following changes were required.
- Redux store now handles the current index states
- Pattern handler updates the redux store whenever there are changes
- Wz menu reloads it's data whenever there are changes to the index patterns

### To test

- **Given** The app contains a single index pattern
- **And** The browser is on `app/wazuh#/settings?tab=configuration`
- **And** The user has edited the index pattern
- **When** The user executes the health check
- **Then** An index pattern selector will appear in the menu

Closes https://github.com/wazuh/wazuh-kibana-app/issues/3579

### DIFF
--- a/public/components/wz-menu/wz-menu.js
+++ b/public/components/wz-menu/wz-menu.js
@@ -24,14 +24,17 @@ import {
   EuiLoadingSpinner,
   EuiFormRow,
   EuiSelect,
-  EuiSpacer
+  EuiSpacer,
 } from '@elastic/eui';
 import { AppState } from '../../react-services/app-state';
 import { PatternHandler } from '../../react-services/pattern-handler';
 import { WazuhConfig } from '../../react-services/wazuh-config';
 import { connect } from 'react-redux';
 import WzReduxProvider from '../../redux/wz-redux-provider';
-import { updateCurrentAgentData, showExploreAgentModalGlobal } from '../../redux/actions/appStateActions';
+import {
+  updateCurrentAgentData,
+  showExploreAgentModalGlobal,
+} from '../../redux/actions/appStateActions';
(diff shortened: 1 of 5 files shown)


public/components/wz-menu/wz-menu.js
```diff
@@ -47,349 +50,335 @@ import WzTextWithTooltipIfTruncated from '../../components/common/wz-text-with-t
...
-        if (!filtered) AppState.setCurrentPattern(list[0].id);
+        const data = filtered
+          ? filtered
+          : await this.indexPatterns.get(AppState.getCurrentPattern());
+        this.setState({ theresPattern: true, currentPattern: data.title });
+
+        // Getting the list of index patterns
+        if (list) {
+          this.setState({
+            patternList: list,
```
Desvelao, line 164: **nitpick**: If we move the list of valid index patterns to the Redux store, we should not use the component state for this case. Take into account this for renders or conditional, but it should use the values from the store.

mpRegalado: Removed in https://github.com/wazuh/wazuh-kibana-app/pull/3598/commits/bff95d08d148140258ac93ea1255b739e88c5d71

CPAlejandro (approved):
CR: LGTM!
Testing: It works perfectly, i have changed the text of index pattern configuration option and after execute Health-Check, it appears the index pattern selector, so LGTM!

PR #3604, opened 2021-09-15
Removed api selector toggle from the settings menu

mpRegalado:
This PR removes the api selector toggle from the settings menu, since it was not actually affecting the state of the selector in the menu bar.

Closes https://github.com/wazuh/wazuh-kibana-app/issues/3335

### To test

- **Given** the browser is in wazuh app
- **When** the user navigates to `settings/configuration`
- **Then** the configuration panel does not show an api selector toggle

### Screenshots
Before
Before

After
After

### DIFF
--- a/public/react-services/app-state.js
+++ b/public/react-services/app-state.js
@@ -237,12 +237,6 @@ export class AppState {
    * Set a new value to the 'patternSelector' cookie
    * @param {*} value
    */
-  static setAPISelector(value) {
-    const encodedPattern = encodeURI(value);
-    getCookies().set('APISelector', encodedPattern, {
-      path: window.location.pathname
-    });
-  }
 
   /**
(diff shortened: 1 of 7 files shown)


public/react-services/app-state.js
```diff
@@ -237,12 +237,6 @@ export class AppState {
    * Set a new value to the 'patternSelector' cookie
```
asteriscos, line 237: If the method was deleted it's comments should also be deleted.

CPAlejandro (approved):
CR: LGTM!
Testing: LGTM!

asteriscos (changes requested):
Tested: :heavy_check_mark: 
Code looks good to me, except a small comment

asteriscos (approved):
LGTM

Issue #3332, opened 2021-06-04
Group agent management table does not update on error


mpRegalado:
| Wazuh | Elastic | Rev |
| ----- | ------- | --- |
| 4.x   | 7.x     | 444 |

**Description**
When you have a 2 column management interface, such as the "Manage Agents" in the groups section, if there is an error while processing the operation, the state does not update to reflect what operations have been applied at the moment, leaving the possibility of further errors.

This is different from the behavior on a successful operation where it navigates the user out of the panel, forcing it to load the correct state if opened again.

**Steps to reproduce**

1. Go to 'Management -> Groups -> Specific Group'
2. Click on 'Manage Agents'
3. Attempt to add multiple agents to the group, at least one of which is never connected.
4. See that all the agents are still on the right side column, and removing the never connected agent and attempting to apply changes leads to more error messages.

**Screenshots**
![agentTable](https://user-images.githubusercontent.com/80431234/120780569-8d01c800-c528-11eb-9906-6e5993d1be87.gif)

**Additional context**
Same issue stands with any other error, like trying to remove agents with at least one not belonging to any other group.
I reproduced the issue in 4.2-7.12.1-dev and 4.1.5-7.10.0-prod

sortizowlh:
Different errors when trying to add disconnected agents and never connected agents to a group:

- Disconnected
![imagen](https://user-images.githubusercontent.com/47242022/132532217-3ee39a81-9279-4430-9a87-6d4f11143f9b.png)
- Never connected
![imagen](https://user-images.githubusercontent.com/47242022/132532273-99fc0cbe-394d-4ce4-ab1e-f2839de67aec.png)

sortizowlh:
- Save the previous state.
- Check if agents are allocated in their correct previous position
- Move agents to their previous position

PR #3605, opened 2021-09-15
Disable apply button at initial state and move agents to their original position

sortizowlh:
Hi team!

When managing agents into a group generates an error, all the agents move to their original positions.
Also, the `Apply changes` button is disabled when all the agents are in their original position.

**To test it:** 

1. Go to Management/Groups

- **Given** the browser is in wazuh app with more than 1 user available
- **When** the user navigates to Management/Groups
- **Then** the Groups section will be shown

2. Create a Group

- **Given** the Group section
- **When** the user click in add new group
- **And** the user fill the name with "test"
- **And** the user click on "Save new Group"
- **Then** the group will be displayed in Groups table

3. Enter in Manage Agents in Groups section

- **Given** the group section with the group "test" created
- **When** the user click on test
- **And** the user click on "Manage agents" in the Test group section
- **Then** the agents will be displayed in Available agents and current agents in the group will be empty

4. Try to move agents in Manage agents of group test

- **Given** the section to move agent from some groups
- **When** the user can move all agents to the right or left back, save them and move them again
- **Then** the button to save will be enabled if the user move some of the users and there isn't any errors (except when the user will try to move an inactive or never connected agent, that will persist in the left side)

Solves: #3332


asteriscos (changes requested):
Test: ❌

When one of the agents is added and the rest is not, it doesn't end in the right group column. Also, the `See full error` doesn't do anything. In addition, the number of `added` or `removed` items is not reset after the operation.

Peek 2021-09-15 13-42

public/components/management/groups/multiple-agent-selector.tsx
```diff
@@ -9,707 +9,897 @@
...
-      item.forEach(elem => this.moveItem(elem, from, to, type));
-      this.checkLimit();
-    } else {
-      item = JSON.parse(item);
-      const idx = from.findIndex(x => x.key === item.key);
-      if (idx !== -1) {
-        from.splice(idx, 1);
-        item.type = !item.type ? type : '';
-        to.push(item)
+    moveItem = (item, from, to, type) => {
```
mpRegalado, line 405: This functions mutates the state. It doesn't seem to have been changed in this PR but it should be addressed

public/components/management/groups/multiple-agent-selector.tsx
```diff
@@ -9,707 +9,897 @@
...
+        //relaod agent columns
+        this.reload('right');
+        this.reload('left');
+
+        try {
+          //delete nodes selected
+          await WzRequest.apiReq('DELETE', `/agents/group`, {
+            params: {
+              group_id: this.props.currentGroup.name,
+              agents_list: itemsToSave.addedIds.toString(),
```
mpRegalado, line 362: Deleting all agents from the `itemsToSave` list upon error means that the agents that caused the error in the first place are not in the group. So attempting to delete them causes another error. The final consequence is simply a console log, but the `failedIds` list should be updated when there is an error and used to filter what items are to be deleted.

mpRegalado (changes requested):
The code performs the function required. Upon encountering an error, it reverts all the changes to its original state.
However, the way that it does so causes further errors to be logged into the console.

public/components/management/groups/multiple-agent-selector.tsx
```diff
@@ -9,707 +9,897 @@
...
+        if (itemsToSave.addedIds.length) {
+          const addResponse = await WzRequest.apiReq('PUT', `/agents/group`, {
+            params: {
+              group_id: this.props.currentGroup.name,
+              agents_list: itemsToSave.addedIds.toString(),
+            },
+          });
+
+          if (addResponse.data.data.failed_ids) {
+            failedIds.push(...addResponse.data.data.failed_ids);
```
mpRegalado, line 293: This code is never executed. An error in the assignment of `addResponse` raises an exception that skips this section to the `catch` phase, meaning that `failedIds` is never updated.

CPAlejandro (approved):
CR and Testing: LGTM!

PR #3638, opened 2021-09-30
Fix #3578 Logging out of the API should deprecate the token

### LINKED ISSUE #3578: Logging out of the API should deprecate the token
| 4.2.0 | 7.10.2 - 7.11.2 | 4201 | Basic, ODFE, Xpack |

| Browser |
| ------- |

**Description**
When logging out of the API, the token is not always deprecating (given the option to continue its usage after the session is closed)

**Preconditions**
Having an active API token

**Steps to reproduce**
1- Get API token with CURL
2- Logout from Kibana
3- Throw GET request with curl, using the API_TOKEN from step 1

**Expected Result**
The token should be deprecated

**Actual Result**
The token is still active

**Screenshots**

**Additional context**
Add any other context about the problem here. Here you can paste log entries too or any other useful information that may help with the issue.

AlexRuiz7:
The request to deprecate the user's token was cancelled by the browser due to the removal of the DOM element that triggered the event. In this case, the logout button perfomed an inmediate redirect to the `'/logout'` page, and therefore, removing the element from the DOM, causing the failure of the request.

As a solution, the anchor's behaviour (<a href='logout'..>)(part of X-Pack plugin) has been overrided with `event.preventDefault()`, to later on navigate programatically to `'/logout'` once a response from the server has been received.

image

### DIFF
--- a/public/app.js
+++ b/public/app.js
@@ -111,6 +111,8 @@ app.run(function ($rootElement) {
   // Add plugin help links as extension to Kibana help menu
   addHelpMenuToAppChrome();
 
+  const urlToLogout = window.location.origin + '/logout';
+
   // Bind deleteExistentToken on Log out component.
   $('.euiHeaderSectionItem__button').on('mouseleave', function () {
     // opendistro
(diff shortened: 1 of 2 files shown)


public/app.js
```diff
@@ -109,15 +111,22 @@ app.run(function ($rootElement) {
   // Add plugin help links as extension to Kibana help menu
   addHelpMenuToAppChrome();
 
+  const urlToLogout = window.location.origin + '/logout';
+
   // Bind deleteExistentToken on Log out component.
   $('.euiHeaderSectionItem__button').on('mouseleave', function () {
     // opendistro
     $('span:contains(Log out)').on('click', function () {
       WzAuthentication.deleteExistentToken();
```
AlexRuiz7, line 120: May be worth to apply the same behaviour to ODFE

mpRegalado (approved):
CR: LGTM :heavy_check_mark: 
Testing: LGTM :heavy_check_mark: 

To test:

- **Given** any browser with a valid `wz-token` in cookies running any security plugin
- **When** The user logs out
- **Then** Executing `curl -k -X GET "https://$WZMANAGER:55000/" -H "Authorization: Bearer $WZTOKEN"` returns `{"title": "Unauthorized", "detail": "No authorization token provided"}`

Issue #3634, opened 2021-09-28
[Wazuh API] [Optimization] Fix the requests `GET //`


Desvelao:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.x   | 7.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |
| Chrome, Firefox, Safari, etc |

**Description**
The app does requests to the Wazuh API 
```
GET //
```
that returns a response with status `308` and as a result a redirection. The app backend, thank you to the `axios` library, manages to do a request to the redirection endpoint `GET /` and returns the response to the frontend.

We are doing an invalid request to the Wazuh API that could be avoided.

Wazuh API logs (`/var/ossec/logs/api.log`):
```
2021/09/28 08:07:11 INFO: wazuh-wui 172.19.0.3 "GET //" with parameters {} and body {} done in 0.000s: 308
2021/09/28 08:07:11 INFO: wazuh-wui 172.19.0.3 "GET /" with parameters {} and body {} done in 0.008s: 200
```

This occurs when accessing the view for an agent after selecting the `Agents` section.

Backend app request:
```
GET // { headers: {} } { status: 200,
  statusText: 'OK',
  headers:
   { 'content-type': 'application/json; charset=utf-8',
     server: 'Wazuh',
     'strict-transport-security': 'max-age=63072000; includeSubdomains',
     'x-frame-options': 'DENY',
     'x-xss-protection': '1; mode=block',
     'x-content-type-options': 'nosniff',
     'content-security-policy': 'none',
     'referrer-policy': 'no-referrer, strict-origin-when-cross-origin',
     pragma: 'no-cache',
     expires: '0',
     'cache-control': 'no-cache, no-store, must-revalidate, max-age=0',
     'content-length': '277',
     date: 'Tue, 28 Sep 2021 08:40:05 GMT',
     connection: 'close' },
  config:
   { url: 'https://172.17.1.2:55000//',
     method: 'get',
     data: '{}',
     headers:
      { Accept: 'application/json, text/plain, */*',
        Authorization:
[...]

PR #3661, opened 2021-10-22
[Wazuh API] Fix the requests `GET //`

yenienserrano:
**Description**

Replace all the requests to GET // by the valid endpoint to avoid redirection and a useless request.

**Issues Resolved**

[Wazuh API] [Optimization] Fix the requests `GET //`[#3634](https://github.com/wazuh/wazuh-kibana-app/issues/3634)

**Test**

see the logs of the wazuh api and in the UI go to the agents table and see that the redirect "//" is no longer done, it only does the "/"

**Check List**

- [x]  Replace all the requests to GET // by the valid endpoint to avoid redirection and a useless request.

image


Desvelao:
### Changes
- Removed the space in the root Wazuh API endpoint (related to `GET / ` requests in the previous commits of this PR)
- Fixed the requirement for the `path` endpoint that should start with `/`. Now uses `.startsWith` function instead a regular expression that required something more than `/` in the path

gabiwassan (approved):
LGTM!

PR #3670, opened 2021-11-01
Fix remove token on logout kibana 7.13

frankeros:
Hi team this resolves: 

Remove Wazuh API token on logout with Kibana 7.13 and 7.14

To test:
- login and copy the wz-token and make a curl with this token
- logout and repeat the curl
- the second curl should return 401


gabiwassan (approved):
CR: ✅

CPAlejandro (approved):
CR: LGTM!

Testing:

- 7.10.2: ✅ 
- 7.13.4: ✅ 
- 7.14.2: ✅

PR #3681, opened 2021-11-04
[FIX] [Compatibility] Get the valid index patterns to use in the app in Kibana 7.10.2 - 7.13.4

Desvelao:
## Description

This PR adds compatibility to `Kibana version >= 7.11.x` to manage how the index pattern data is retrieved to check if meets the Wazuh requirements to use.

The index pattern of Wazuh alerts should own a minimum set of fields (`agent.id`, `manager.name`, `rule.groups` and `timestamp`) to be considered usable in the app. 

For Kibana `<7.11.0`, the index pattern fields are stored as the saved object's `attribute.fields` property and these are getting with the next request:
```
GET <KIBANA_PROTOCOL>://<KIBANA_HOST>:<KIBANA_PORT>/api/saved_objects/_find?type=index-pattern&fields=title&fields=fields&per_page=9999
```
but on `Kibana 7.11.x or newer` only the scripted fields are in the `attributes.fields` so is required getting the fields with:
```
GET <KIBANA_PROTOCOL>://<KIBANA_HOST>:<KIBANA_PORT>/api/index_patterns/_fields_for_wildcard?pattern=<INDEX_PATTERN_NAME>
```

More information in:
- [Kibana 7.11.0 changelog](https://www.elastic.co/guide/en/kibana/7.11/release-notes-7.11.0.html)
- [PR that modified the cache of index pattern fields](https://github.com/elastic/kibana/pull/82223)
- [New Index pattern API](https://github.com/elastic/kibana/pull/83576). This API is not used in the current app PR to maintain compatibility between Kibana versions

### Changes
  - Added a function to check if the Kibana version (`kibana.version` property of `package.json`) satisfies a specific version.
  - Modified some logic when getting the index patterns or refreshing them to be compatible with 7.10.2 and 7.13.4. A Kibana change was modified as getting the index pattern fields.
  - Added a toast when an unknown index pattern field is detected in
Kibana >7.10.2 in `Modules/<MODULE>/(Dashboard/Events)`
    - set its lifetime to `5 minutes` instead of the default time to the toast for the different versions.

### Notes about the fix
- The fix uses the `kibana.version` property of the `package.json`. Set this property to the Kibana version you are using.
- It was tested with Kibana 7.10.2 and 7.13.4 but should work in Kibana 7.11.x, 7.12.x. For 7.14.x or newer, this should be tested to check if works.

## Tests
- With a Kibana without index patterns, the app should create the index patterns (alerts, monitoring, statistics).
- Go to Settings/Configuration and change the default index pattern to one that matches the name of Wazuh alerts indices, save and push the button to `Execute the health check`. The new index pattern should be created correctly.
- Add some alerts that contain not existent fields, it could use the app sample data to generate alerts of AWS/GC by example, if there is no previous related data and appears new fields, go to `Modules/Amazon Web Services/Events` and explore the new alerts, expand the alert information and a message should be displayed saying you need to refresh the page to apply changes (`7.10.2`) or update de fields (`>7.11.0`)
- Create an index pattern from `Kibana>Stack management>Index patterns` that matches the name of Wazuh alerts indices and go to the app, when opening the menu, the index pattern should be retrieved and these should appear in the index pattern selector of the app. Create some index pattern that doesn't matches with the Wazuh alerts and check these don't appear in the index pattern selector

The next condition should be fulfilled:
  - `<=7.10.2`: the fields should be in the `attributes.fields` of the index pattern
  - `>=7.11.0`: the fields should not be in the `attributes.fields` of the index pattern
  ```
  GET <KIBANA_PROTOCOL>://<KIBANA_HOST>:<KIBANA_PORT>/api/saved_objects/_find?type=index-pattern&fields=title&fields=fields&per_page=9999
  ```
  
## Issues
This PR is related to a bug reported in https://github.com/wazuh/wazuh-kibana-app/issues/3310#issuecomment-954609355.

### DIFF
--- a/public/components/common/modules/events.tsx
+++ b/public/components/common/modules/events.tsx
@@ -171,7 +171,7 @@ export class Events extends Component {
     if (!this.state.hasRefreshedKnownFields) {
       try {
         this.setState({ hasRefreshedKnownFields: true, isRefreshing: true });
-        if(satisfyKibanaVersion('<=7.10.2')){
+        if(satisfyKibanaVersion('<7.11')){
           await PatternHandler.refreshIndexPattern();
         };
         this.setState({ isRefreshing: false });
@@ -236,7 +236,7 @@ export class Events extends Component {
 
   reloadToast = () => {
     const toastLifeTimeMs = 300000;
-    if(satisfyKibanaVersion('<=7.10.2')){
+    if(satisfyKibanaVersion('<7.11')){
       getToasts().add({
         color: 'success',
         title: 'The index pattern was refreshed successfully.',
@@ -251,7 +251,7 @@ export class Events extends Component {
         </EuiFlexGroup>),
         toastLifeTimeMs
       });
-    }else if(satisfyKibanaVersion('>7.10.2')){
+    }else if(satisfyKibanaVersion('>=7.11')){
       getToasts().add({
         color: 'warning',
         title: 'Found unknown fields in the index pattern.',
--- a/public/components/common/modules/events.tsx
+++ b/public/components/common/modules/events.tsx
@@ -20,6 +20,7 @@ import { PatternHandler } from '../../../react-services/pattern-handler';
 
 import { enhanceDiscoverEventsCell } from './events-enhance-discover-fields';
 import { toMountPoint } from '../../../../../../src/plugins/kibana_react/public';
+import { satisfyKibanaVersion } from '../../../../common/semver';
 
 export class Events extends Component {
   intervalCheckExistsDiscoverTableTime: number = 200;
@@ -170,10 +171,11 @@ export class Events extends Component {
     if (!this.state.hasRefreshedKnownFields) {
       try {
         this.setState({ hasRefreshedKnownFields: true, isRefreshing: true });
-        await PatternHandler.refreshIndexPattern();
-
+        if(satisfyKibanaVersion('<=7.10.2')){
+          await PatternHandler.refreshIndexPattern();
+        };
         this.setState({ isRefreshing: false });
-        this.reloadToast()
+        this.reloadToast();
 
       } catch (err) {
         this.setState({ isRefreshing: false });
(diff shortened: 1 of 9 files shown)


Desvelao:
## Changes
- Replaced the references of version from `7.10.2` to 7.11`

public/components/common/modules/events.tsx
```diff
@@ -170,10 +171,11 @@ export class Events extends Component {
     if (!this.state.hasRefreshedKnownFields) {
       try {
         this.setState({ hasRefreshedKnownFields: true, isRefreshing: true });
-        await PatternHandler.refreshIndexPattern();
-
+        if(satisfyKibanaVersion('<7.11')){
```
frankeros, line 174: '<7.11' could be a constant

public/components/common/modules/events.tsx
```diff
@@ -233,19 +235,38 @@ export class Events extends Component {
...
+            There were some unknown fields for the current index pattern.
+            You need to refresh the page to apply the changes.
+          </EuiFlexItem>
+          <EuiFlexItem grow={false}>
+            <EuiButton onClick={() => window.location.reload()} size="s">Reload page</EuiButton>
+          </EuiFlexItem>
+        </EuiFlexGroup>),
+        toastLifeTimeMs
+      });
+    }else if(satisfyKibanaVersion('>=7.11')){
```
frankeros, line 254: 7.11 is included?

public/react-services/saved-objects.js
```diff
@@ -28,11 +29,30 @@ export class SavedObject {
...
+        indexPatternsFields = await Promise.all(indexPatterns.map(async indexPattern => {
+          try{
+            const {data: {fields}} = await GenericRequest.request(
+              'GET',
+              `/api/index_patterns/_fields_for_wildcard?pattern=${indexPattern.attributes.title}`,
+              {}
+            );
+            return fields;
+          }catch(error){
+            return [];
```
frankeros, line 51: it could be a false positive, maybe the better is let it throw ?

Machi3mfl (approved):
CR ✅

public/react-services/saved-objects.js
Desvelao: If the index pattern is created, but there are no indices, then this request fails, and as result, the method fails.

public/react-services/saved-objects.js
```diff
@@ -261,7 +280,10 @@ export class SavedObject {
    */
   static async createWazuhIndexPattern(pattern) {
     try {
-      const fields = await SavedObject.getIndicesFields(pattern, WAZUH_INDEX_TYPE_ALERTS);
+      let fields = '';
```
CPAlejandro, line 283: Could this be something like:
`const fields = satisfyKibanaVersion('<7.11') ? await SavedObject.getIndicesFields(pattern, WAZUH_INDEX_TYPE_ALERTS) : '';`

public/react-services/saved-objects.js
```diff
@@ -183,8 +200,11 @@ export class SavedObject {
         params
       );
 
-      if (type === 'index-pattern')
-        await this.refreshFieldsOfIndexPattern(id, params.attributes.title, fields);
+      if(satisfyKibanaVersion('<7.11')){
```
CPAlejandro, line 203: Could this be:
`if(satisfyKibanaVersion('<7.11') && type === 'index-pattern'){
`

public/components/common/modules/events.tsx
Desvelao: Yes, it is. `7.11.0 or higher`

public/react-services/saved-objects.js
Desvelao: Yes, I prefered to nest the `if`s instead a complex condition.

public/components/common/modules/events.tsx
Desvelao: Yes, it is a good idea, but I think it is not necessary. The variable name could be overcomplicated.

PR #3692, opened 2021-11-24
Fixing error that shows we're using X-Pack when we have Basic

CPAlejandro:
SHOULD_TEST=FALSE
Hi guys,

We are fixing a bad behaviour that's shows to Wazuh app like we are using X-Pack when we have Basic Security.

To test it:

Test with X-Pack, Basic and ODFE and be sure that the app is receiving properly which security plugin you are using.

Close [this comment](https://github.com/wazuh/wazuh-kibana-app/issues/3684#issuecomment-972941510)


server/controllers/wazuh-api.ts
```diff
@@ -1060,7 +1060,7 @@ export class WazuhApiCtrl {
       
       const disabledRoles = ( await getConfiguration() )['disabled_roles'] || [];
       const logoSidebar = ( await getConfiguration() )['customization.logo.sidebar'] || 'icon_blue.png';
-      const wazuhSecurity = SecurityObj(context.wazuh.plugins);
+      const wazuhSecurity = await SecurityObj(context.wazuh.plugins, context);
```
Desvelao, line 1063: **nitpick**: use the `context.wazuh.security` instead of create another security object.

server/lib/security-factory/security-factory.ts
```diff
@@ -1,23 +1,34 @@
...
+
+  try {
+    const responseCurl = await context.core.elasticsearch.client.asInternalUser.transport.request(
+      params
+    );
+  } catch (error) {
+    return !!opendistroSecurityKibana
+      ? new OpendistroFactory(opendistroSecurityKibana)
+      : new DefaultFactory();
   }
```
Desvelao, line 32: **nitpick**: The request to check if X-Pack is enabled should be done only if `security` plugin is installed.

Desvelao (approved):
**review**:

Code Review: LGTM

sortizowlh (approved):
CR: 🟢 
Testing: 🟢

PR #3709, opened 2021-12-07
[FIX] [TELEMETRY] Hide the banner

Desvelao:
## Description
This PR hides the `telemetry` banner from UI and saves that the banner was seen to avoid the banner is displayed on the next times.

The status of if the banner was seen/dismissed of `telemetry` banner is stored in a saved object. This is persistent. For new environments, the attribute that controls this is not stored so the banner is displayed, if use the `Dismiss` button, store the flag in the saved objects for next times.

Moreover, the status is saved as a property i the `telemetry` plugin.

## Changed
  
- Hide the telemetry banner from UI and set a flag in the saved object that controls if the banner was seen and dismissed
- Changed the `wazuh` `public` `start` method to sync instead of `async` function

## Tests
- In new environments, the flag in a saved object of the `telemetry` banner is not set, so this is displayed. The PR, removes the banner from UI and sets the flag in the saved object and this should not be displayed.

Note:
Delete the `telemetry` object so this should be displayed again, or remove the indices `.kibana*` (where the telemetry saved object is stored) so it could test it again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 
 - Support for Kibana 7.13.4
 - Support for Kibana 7.14.2
+- Hide the `telemetry` banner [#3709](https://github.com/wazuh/wazuh-kibana-app/pull/3709)
 
 ### Fixed
 
--- a/kibana.json
+++ b/kibana.json
@@ -17,7 +17,8 @@
     "savedObjects",
     "kibanaReact",
     "kibanaUtils",
-    "securityOss"
+    "securityOss",
+    "telemetry"
   ],
   "optionalPlugins": [
     "security",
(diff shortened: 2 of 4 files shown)


gabiwassan (approved):
LGTM!

eze9252 (approved):
LGTM!

asteriscos (changes requested):
CR: :heavy_check_mark: 
Test: :x: 

Tested it with Open Distro and I'm getting this error before logging in

image

Desvelao:
### Changes
Now the telemetry banner is hidden and set the flag of the saved object when the frontend of the Wazuh app is to be mounted.

This means that the banner is visible if it was not dismissed before.

asteriscos (approved):
Test: :heavy_check_mark: 
CR: :heavy_check_mark:

PR #3722, opened 2021-12-09
[Panel] [Office365] [GitHub] Update data when clearing the query filter

Desvelao:
…r in the search bar and the data is not updated

  - Now uses `useQueryManager` instead of `useQuery`
  
## Description
This PR solves when clearing the query filter in the search bar of `Panel` tab of modules `Office 365` and `GitHub`. See https://github.com/wazuh/wazuh-kibana-app/issues/3713#issue-1075421139

## Changes
- Replace the `hook` used by `useQueryManager` instead of `useQuery`.

## Tests
- Add a query filter in the search bar of `Panel` tab of modules `Office 365` and `GitHub` and apply the searching, after clearing the query string and updating the search, the data should be updated.


eze9252 (approved):
LGTM!

PR #3792, opened 2022-01-07
Enhance agent deployment UI

### LINKED ISSUE #3762: Improve agent deployment UI
Hello,

looking at the current agent deployment UI I see improvements are needed:

- The registration password appears as asteriscs in the deployment one-liner. We should only show the agent deployment UI if the current Wazuh API user has admin privileges, including the full registration password.
- The deployment one-liner should be shown in a text-area type of element.
- The `Copy command` buttons aren't clear. They look like they are pressed and it is not clearly stated what they copy.
- The deployment one-liner for Windows uses the `Invoke-WebRequest` cmdlet. It is meant to be executed in a PowerShell terminal, but it is not stated anywhere.
Regards,
Javier.

asteriscos:
Hi team, 

this PR enhances the user experience when trying to copy an agent install code block. Also warns the user that the Windows script must be run in a PowerShell terminal.

Closes #3762 

Peek 2022-01-07 17-11

### Tests
- This should be tested with different browsers.

### DIFF
--- a/public/controllers/agent/components/register-agent.js
+++ b/public/controllers/agent/components/register-agent.js
@@ -543,11 +543,13 @@ export const RegisterAgent = withErrorBoundary(
                     )}
                   </EuiCopy>
                 </div>
-                <EuiSwitch
-                  label="Show password"
-                  checked={this.state.showPassword}
-                  onChange={(active) => this.setShowPassword(active)}
-                />
+                {this.state.needsPassword && (
+                  <EuiSwitch
+                    label="Show password"
+                    checked={this.state.showPassword}
+                    onChange={(active) => this.setShowPassword(active)}
+                  />
+                )}
                 <EuiSpacer />
                 {windowsAdvice}
               </EuiText>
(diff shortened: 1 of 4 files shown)


Desvelao:
**nitpick** Use `PowerShell` instead of `powershell`

yenienserrano (approved):
CR y Test: :heavy_check_mark: 
LGTM

public/controllers/agent/components/register-agent.js
```diff
@@ -569,16 +588,24 @@ export const RegisterAgent = withErrorBoundary(
...
+                      </div>
+                    )}
+                  </EuiCopy>
+                </div>
+                <EuiSwitch
+                  label="Show password"
+                  checked={this.state.showPassword}
+                  onChange={(active) => this.setShowPassword(active)}
+                />
+                <EuiSpacer />
```
Desvelao, line 608: **thought**: I am not sure why a switch to change the visibility of password is displayed here. The code block displays the code to enable and start the `wazuh-agent` service.

public/controllers/agent/components/register-agent.js
```diff
@@ -523,17 +531,25 @@ export const RegisterAgent = withErrorBoundary(
...
```
Desvelao, line 551: **nitpick**: The switch that manages the visibility of the password should only be displayed when there is one.

```jsx
{this.state.wazuhPassword && (
  <EuiSwitch
    label="Show password"
    checked={this.state.showPassword}
    onChange={(active) => this.setShowPassword(active)}
  />
                <EuiSpacer />
)}
```

asteriscos: great suggestion

matiasmoreno876 (approved):
CR LGTM ✅

Issue #3786, opened 2022-01-05
The height of the evolution card is smaller than the other cards.


Mayons:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.3  | 7.16.1     | 4.3 RC2 | Xpack |

| Browser |
| ------- |
| Chrome, Firefox, Safari, etc|

**Description**
The height of the evolution card is smaller than the other cards when it has or no data.

**Preconditions**
1. Login with valid credentials

**Steps to reproduce**
1. Navigate to the agent's page

**Expected Result**
1. The evolution card should have the same height as the other cards

**Actual Result**
1. The height of the evolution card is smaller than the other cards.

**Screenshots**
![Evolution card](https://user-images.githubusercontent.com/33849347/148206263-fc5aa0f3-80ae-4944-921a-cba3a533e647.png)

**Additional context**
Add any other context about the problem here. Here you can paste log entries too or any other useful information that may help with the issue.

yenienserrano:
**Description**

Change of styles so that it has the same height as the other cards and the same length when it has this one the screen is small

**PR**

Fixed the styles on the evolution card [#3796](https://github.com/wazuh/wazuh-kibana-app/pull/3796)

**Test**

1. Navigate to the agent's page
2. Look at the evolution card

**Check List**

[x] The evolution card is the same size as the other cards

![image](https://user-images.githubusercontent.com/63758389/149157642-a3790d6f-902a-4d79-b979-0b5abd11753a.png)

PR #3796, opened 2022-01-12
Fixed the styles on the evolution card

yenienserrano:
**Description**

Change of styles so that it has the same height as the other cards and the same length when it has this one the screen is small

**Issues Resolved**

The height of the evolution card is smaller than the other cards. [#3786](https://github.com/wazuh/wazuh-kibana-app/issues/3786)

**Test**

1. Navigate to the agent's page
2. Look at the evolution card

**Check List**

[x] The evolution card is the same size as the other cards

image


Desvelao (approved):
**review**
Code ✅ 
Test ✅ 
  > Environment: 
  > - Kibana 7.16.1 (dev)

PR #3877, opened 2022-03-03
[BUG][Tools/Api Console] Parse error when using  character on the line start

### LINKED ISSUE #3810: [Tools/APIConsole] The `#` character provokes a `Parse error` when not used as comment
| 4.x   | 7.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |

**Description**
The `#` character provokes a `parse error` in the API Console when is not used as the first character of the line. This bug is avoiding to create a user with a password that contains that character. It could be affecting other endpoints when using this character.

## Tasks
- [x] Research the reason for the problem
- [x] Apply the fix to solve it

**Preconditions**
1. 

**Steps to reproduce**
1. Navigate to `Tools/API Console`.
2. Add the next code:
```
POST /security/users
{
  "username": "custom",
  "password": "1234abcD#"
}
```
3. A parse error is displayed in the UI.

**Expected Result**
1. There should not be an error when parsing the API request.

**Actual Result**
1. A parse error is displayed in the UI.

**Screenshots**
image

Desvelao:
## Description
This PR fixes when the `#` character is used not at the beginning of the line. This solves a parse error when using the character in the `password` field to create a new user.

## Tests
- Go to `Tools/API Console` and set the next query:
```
POST /security/users
{
  "username": "wazuh_test_user",
  "password": "Abcdef1#"
}
```

The query shouldn't display the parse error.

Closes #3810

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to the Wazuh app project will be documented in this file.
 
+## Wazuh v4.3.0 - Kibana 7.10.2 , 7.16.3, 7.17.0 - Revision 4302
+
+### Fixed
+- Parse error when using `#` character not at the beginning of the line [#3877](https://github.com/wazuh/wazuh-kibana-app/pull/3877)
+
 ## Wazuh v4.3.0 - Kibana 7.10.2 , 7.16.3, 7.17.0 - Revision 4301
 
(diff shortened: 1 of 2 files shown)


Machi3mfl (approved):
CR ✅
Test ✅

PR #3893, opened 2022-03-10
Vulnerability inventory rework

asteriscos:
Hi team,

this PR removes Vulnerability dashboard, adds visualizations to Vulnerability inventory and enhances Details Flyout fields. 
### UI
- Removed `Dashboard` tab of `Vulnerabilities` module.
  - Removed definitions of visualizations
  - Removed the dashboard layouts: `agent` and `overview` (no selected agent)
- Added 3 new panels to `Vulnerabilities/Inventory`:
  - `Severity`: visualization
  - `Details`: stats
  - `Summary`: visualization with selector by field
- Added new fields to the flyout with the details of vulnerability:
  - `condition`
  - `external_references`
  - `published`
  - `title`
  - `updated`
- Added a new field `data.vulnerability.status` to the table of `Vulnerabilites/Events`
  
### Logic
- Refactorized the methods to get the data.

### React components
- Refactorized the donut visualization of `Compliance` panel in the agent summary dashbaord to reusable components.
- Added `VisualizationBasic`, `VisualizationBasicWidget` and `VisualizationBasicSelector` components.

### React hooks
- Added `useAsyncActionRunOnStart` hook.

TODO:
 - [ ] Add screenshots
 
Closes #3883 and #3890 

## Modules/Vulnerabilites
- The initial tab should be `Inventory`.
- The `Dashboard` tab was removed.
### Modules/Vulnerabilities/Inventory
- With no vulnerabilities in an agent:
  - `Severity` panel:
    - display `No results` message 
  - Details` panel:
    - severity levels sorted by `Critical`, `High`, `Medium` and `Low` from left to right
    - `last full scan` and `last partial scan` stats display as `-`.
  - `Summary` panel:
    - display `No results` message
- With vulnerabilities in an agent:
    - `Severity` panel:
      - severity levels sorted by `Critical`, `High`, `Medium` and `Low`
    - Details` panel:
      - severity levels sorted by `Critical`, `High`, `Medium` and `Low` from left to right
      - `last full scan` and `last partial scan` stats' format depends on the plugin platform date format.
    - `Summary` panel:
      - display selector
      - changing the selected field should fetch data, displaying a loading status, and renders the visualization with the new data
  - Severity levels in the visualization of `Severity` panel and `Details` should match their colors.
- Select another agent, should fetch the data corresponding to the selected agent, displaying the loading status, and renders the result depending on if there is data or not.
- Visualization behavior:
  - Hover the donut chart should display a tooltip. The tooltip has the label of data and its color.
[...]

PR #3908, opened 2022-03-22
Enhance rework vulnerabilities inventory flyout

asteriscos:
Hi team,

this PR show all available fields in Discover Details Flyout table. Also fixes the open row icon width in the first column when the table has few columns.
Related PR #3893 

To test it:
- Have some vulnerability alerts
- Go to Vulnerabilities / Inventory
- Select a row in the table which has events
- Open the row to see the details and verify it show all available fields for that event

image

vulnerabilities


Desvelao:
**thought**: There are some files that use the `Discover` component. Do we want these components to display all the fields or the default subset? If we want to display all fields in all components, maybe we could remove the default subset and the component property to manage when displaying all fields or the subset of them.

Desvelao:
**Testing**

I found a possible problem with the date of scans. See `Last partial scan`.
image

Maybe we should display an `-` as in `Details` panel:
image

yenienserrano (approved):
Test: :heavy_check_mark:

Issue #3918, opened 2022-03-25
[Modules][Dashboard] The toast to reload the page due to the index pattern fields were refreshed appear constatly despite to refresh the page previously.


Desvelao:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.x   | 7.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |
| Chrome, Firefox, Safari, etc|

**Description**
When accessing some dashboards of modules, a toast appears saying the index pattern fields were refreshed and it is required to reload the page. After reloading the page, the toast appears again.

**Steps to reproduce**
1. Navigate to `Modules/Google Cloud Platform/Dashboard`
2. See the toast
3. Refresh the page by clicking on the `Reload page` button or refreshing the page with the browser button or shortcut.
4. The toast appears again.

**Expected Result**
1. The toast should appear constantly.

**Screenshots**
![image](https://user-images.githubusercontent.com/34042064/160106845-96e10e71-cb30-42b6-9aa0-565f7cb5acff.png)

**Additional context**
Add any other context about the problem here. Here you can paste log entries too or any other useful information that may help with the issue.

Desvelao:
# Research
When building a visualization requires a non-existent field in the index pattern, [an error is caught](https://github.com/wazuh/wazuh-kibana-app/blob/4.3-7.10/public/kibana-integrations/kibana-vis.js#L310-L336). At the current source code status, the field name is extracted from the error message and is checked if this exists in the [known fields](https://github.com/wazuh/wazuh-kibana-app/blob/4.3-7.10/public/utils/known-fields.js) and used to refresh the index pattern fields as an extra field with builtin metadata and the existent fields in the indices.

For some modules such as Office 365 or GitHub, the fields used in the visualizations are not included in the builtin known fields and this causes the field can't be found there and can't be added to the index pattern fields despite the field really doesn't exist in the indices. This causes when there is no data related to alerts of these modules, the toast message of `The index pattern was refreshed successfully` to be displayed every time despite refreshing the browser tab.

## Possible solutions
- Sure the fields used by the visualization are available in the indices before creating the visualizations. This approach changes the current logic.
[...]

PR #3924, opened 2022-03-29
[Known fields] Add fields to the know fields for the index pattern of alerts

Desvelao:
# Description

This PR adds some fields to the known fields for the index pattern of alerts. These fields are used as the default fields when creating the index pattern of alerts and there are no related indices.

We extract the fields used in the visualizations and check if they were defined in the know fields file.

Related to #3918

# Test
- With an environment without indices related to Wazuh alerts, access to the Wazuh plugin for Kibana. The health check should run and create the index pattern for the Wazuh alerts with the fields used in the visualizations of all current modules. Access to each module in the two modes (with and without the selected agent) and should appear the toast that says `The index pattern was refreshed successfully.


asteriscos (approved):
LGTM  :heavy_check_mark:

PR #3942, opened 2022-04-01
Map visualization render problem

### LINKED ISSUE #3919: [UI] [Visualization] Problems with the map visualizations
| 4.x   | 7.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |

**Description**
The map visualization has some problems:
- it is not sometimes completely rendered
- the legend could appear duplicated

The problems were seen in Wazuh dashboard, we should review if they appear in Kibana.

**Tasks**
- [x] Research the causes of the problems
- [x] Review if we can apply some fix taking into account that the visualizations are rendered using a method exposed by the core of plugin platform. If not, we should consider removing the visualizations if the problems are persistent.

**Screenshots**
image

asteriscos:
Hi team,

this PR fixes a miscalculation when the map gets its position relative to the map wrapper and the visualization is in a loading state. 

Closes #3919 

To test it, check all visualizations in all dashboards render properly.


Desvelao (approved):
**review**:
Code review :heavy_check_mark: 
Test :heavy_check_mark:

PR #3949, opened 2022-04-05
Change google group icon

### LINKED ISSUE #3943: Add icon "Google group"
we need to replace the mail icon with the Google group icon

image

asteriscos:
Hi team,

this PR replaces old Google Group Icon by the new one. 

Closes #3943 

image

image


yenienserrano:
starting the project with --no-base-path the icons are seen but if we start it without the --no-base-path the icon is not seen

Desvelao:
> starting the project with --no-base-path the icons are seen but if we start it without the --no-base-path the icon is not seen

Good catch!

Fixed in cf4d7c99a8020875b5f877ce75e4f7c24c8cb49b.

PR #3991, opened 2022-04-12
[Fix] Unhandled promise error when Wazuh API

### LINKED ISSUE #3887: Unhandled promise error when unable to reach Wazuh API
| 4.3   | 7.16+     |  | Basic, ODFE, Xpack |

| Browser |
| ------- |

**Description**

We have discovered thanks to  https://github.com/elastic/kibana/issues/118536 that we have an unhandled promise when a request to the Wazuh API fails because of a connection error. This can make the process crash depending on the configuration of the node.config as stated here https://github.com/elastic/kibana/blob/main/config/node.options#L9  

That configuration is disabled by default in opensearch fork.

**Steps to reproduce**
1. Navigate to a dashboard with an API call
2. Check it works
3. Take down Wazuh API
4. Try to call again to the API
5. check whether the message about an unhandled promise appears in kibana logs.

**Expected Result**
1.  The kibana server does not crash, and do not have an unhandled promise

**Actual Result**
1.  Opensearch dashboard crashes with the default config, kibana does not, but the eror about the unhandled promise is shown.

We need to fix all unhandled promises we find.

asteriscos:
### Description

This PR prevents Wazuh throwing an unhandled promise rejection error, which caused Kibana to crash node.options was not set up to only throw a warning after an unhandled promise rejection.

Closes #3887 

To test it:
- Start up / login in Kibana 
- Take down Wazuh API
- Wait for Wazuh to start cron-jobs
- Kibana should not exit on fatal exception and the promise rejection is handled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -155,6 +155,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 - Fixed the `rule.mitre.id` cell enhancement that doesn't support values with sub techniques [#3944](https://github.com/wazuh/wazuh-kibana-app/pull/3944)
 - Fixed error not working the alerts displayed when changin the selected time in some flyouts [#3947](https://github.com/wazuh/wazuh-kibana-app/pull/3947)
 - Fixed the user can not logout when the Kibana server has a basepath configurated [#3957](https://github.com/wazuh/wazuh-kibana-app/pull/3957)
+- Fixed fatal cron-job error when Wazuh API is down [#3991](https://github.com/wazuh/wazuh-kibana-app/pull/3991)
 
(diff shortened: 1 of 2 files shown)


asteriscos:
Test switching to one Wazuh API down entry :heavy_check_mark: 
Running cron-jobs with Wazuh API idown :heavy_check_mark: 


Peek 2022-04-13 11-35

Desvelao (approved):
**review**:
Code review :heavy_check_mark: 
Test :heavy_check_mark: 
- API host with a problem and enabled and ran the `cron-scheduler` (statistics) job
- Channing the selected API host in the API host selector when the API has some problem:
image

PR #4098, opened 2022-04-25
[Initialize] Remove the migrations tasks related to deprecated `.wazuh` indices (`.wazuh` and `.wazuh-version`)

Desvelao:
# Description

This PR removes the migration tasks in the deprecated `.wazuh` indices (`.wazuh` and `.wazuh-version`). Moreover, an additional side-effect of these changes is the some logs related to these indices won't appear in the plugin logs.

For example, the error in the plugin:
```
ERROR  Could not check if the index .wazuh exists due to no permissions for create, delete or check
```
could be interpreted as an error while this could be ignored for new installations.

# Changes
- Removed the migration tasks in the `.wazuh` indices ( `.wazuh` and `.wazuh-version`).

# Test
- With an environment with the internal user with no permisions in the `.wazuh` index, the above log should not appear.
- The `wazuh-registry.json` file should be created in new environments if it doesn't exist when the plugin starts ( located in `<PLUGIN_PLATFORM_PATH>/data/wazuh/config/wazuh-registry.json`)

Closes https://github.com/wazuh/wazuh-kibana-app/issues/3941

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to the Wazuh app project will be documented in this file.
 
-## Wazuh v4.3.0 - Kibana 7.10.2 , 7.16.3, 7.17.0 - Revision 4301
+## Wazuh v4.3.0 - Kibana 7.10.2 , 7.16.x, 7.17.x - Revision 4301
 
 ### Added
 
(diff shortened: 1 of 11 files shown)


AlexRuiz7 (approved):
CR: passed
Test: passed

yenienserrano (approved):
test: ✔️ 
CR: ✔️

Issue #4045, opened 2022-04-19
Can't remove cluster filters


mauromalara:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.3   | 7.x     | 4301-1 | Basic, ODFE, Xpack |

| Browser |
| ------- |
| Firefox |

**Description**
In the Cluster section (within Management) when an existing but negated filter is added, it cannot be removed.
It has a workaround (refresh the page and enter the Cluster section again).

**Preconditions**
1. A cluster must exist

**Steps to reproduce**
1. Navigate to Management -> Cluster
2. Click on 'Add filter'
3. Add the same filter as the existing one but negated it.

**Expected Result**
1. The App should allow removing the added filter.

**Actual Result**
1. Is not possible to remove the added filter.

**Screenshots**

https://user-images.githubusercontent.com/39094716/164027500-ecdfc38f-de08-402a-a879-ad90132dc9d2.mp4

gdiazlo:
These mechanics come from the ability to select different endpoints. This filter affects the whole application. 

https://user-images.githubusercontent.com/2181918/164730503-541f1efa-cc62-47b8-ab11-991a3f0f47d3.mp4

Interestingly, in the opensearch-dashboard discover, the filters using the same field are rewritten appropriately.

https://user-images.githubusercontent.com/2181918/164731547-7b10a715-dcd2-4b2d-bd97-369bbe7fd1d9.mp4

But not so in our discover section.

![image](https://user-images.githubusercontent.com/2181918/164732126-9cb0e541-96b1-4e33-9d90-de44255f8291.png)

 This could mean there is code in the opensearch-dashboard discover native plugin to solve this.

yenienserrano:
**solved:** 

- #4112

matiasmoreno876:
@yenienserrano This error also appears for kibana versions greater than 7.16.x, we need a PR pointing to the branch [4.3-7.16](https://github.com/wazuh/wazuh-kibana-app/tree/4.3-7.16)

mauceballos:
Testing on kibana v7.10.2 OK

![4112](https://user-images.githubusercontent.com/76791841/165830448-b93b4d78-5c07-4023-bc30-3e385dcaff6b.gif)

Desvelao:
Solved in https://github.com/wazuh/wazuh-kibana-app/pull/4112.

PR #4112, opened 2022-04-28
Can remove custom filter

yenienserrano:
**Description:**

Changes in the function where it checks if the filter is from the modules view or if it is an added filter, since currently the filters that can be added, such as the negation of the filters of each module, cannot be removed, and ideally they are have to be able to remove all custom filters.

**issue:**

- #4045

**Test:**

You should always be able to remove the filter

1. Navigate to any module, for example, Security Events
2. Add filter `NOT cluster.name: wazuh` 
3. Pin filter
4. Unpin filter

**Screenshot:**

Filtros

### DIFF
--- a/public/components/common/modules/modules-helper.js
+++ b/public/components/common/modules/modules-helper.js
@@ -51,16 +51,18 @@ export class ModulesHelper {
     }
     const filters = $(`.globalFilterItem .euiBadge__childButton`);
     for (let i = 0; i < filters.length; i++) {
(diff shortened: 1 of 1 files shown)


AlexRuiz7 (approved):
- Code review: ✔️ 
- Test: ✔️

matiasmoreno876 (approved):
CR OK ✅ 
TEST OK ✅

mauceballos:
Testing on kibana v7.10.2 OK

PR #4119, opened 2022-04-28
Fixed "See full error" on error toast

### LINKED ISSUE #4030: `See full error` toast button exception
| 4.2   | 7.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |

**Description**
When an error toast pops up, sometimes the `See full error` button does nothing and throws an exception in the console.

**Steps to reproduce**
1. Force an error, like killing _wazuh-modulesd_ process in the manager
image

**Expected Result**
1. Clicking the See full error button should open a modal with a full description of the error
2. In case there's no additional information, the button should not appear

**Actual Result**
1. Clicking the button does nothing and throws an exception

**Screenshots**
Peek 2022-04-18 19-10

Screenshot from 2022-04-18 19-11-14

Machi3mfl:
Hi Team,
This PR resolves and error when user click on "See full error" button on error toast.
Closes #4030


public/react-services/wz-request.ts
```diff
@@ -156,4 +152,18 @@ export class WzRequest {
...
-}
\ No newline at end of file
+
+  /**
+   * Customize message and return an error object
+   * @param error 
+   * @param message 
+   * @returns error
+   */
+  static customErroMessage(error, message){
```
Desvelao, line 162: **issue**: if the method is `static` to the class, this is not accessible by the class instance.

**suggestion**: Remove the `static` keyword of the method or replace `this` with the class name when using it.

Explanation:
```
// Define a class
class MyClass{
  // Constructor of the class instances
  constructor(){
    // Initialize the class instance. `this` keyword does reference to the class instance
  }
  // Define a static method for the class
  static staticMethod(){
    // This method is static of the class. This can't be accesed by the class instance.
    // Use:
    // MyClass.staticMethod();
  }
  
  // Define a method accesible by the class instance throught the prototype chain
  methodAccesibleByTheClassInstance(){
    // This method is in the prototype chain of the class instance, so this is accesible.
    // Use:
    // this.methodAccesibleByTheClassInstance();
  }
}
```

**nitpick**: The method name has a typo `Erro` instead of `Error`.

Desvelao:
**thought**: I am not sure that the proposed solution is good. It modified the thrown error in the `WzRequest` service, so the `See full error` button in theory, should work. But, the problem with the button could appear if the error has no the expected interface.

As @Machi3mfl mentioned in this comment https://github.com/wazuh/wazuh-kibana-app/issues/4030#issuecomment-1112490919 for the long term solution, we should adopt the same interface for the errors. This will need an analisys of the solution and the changes could break some things.

My proposed solution is sure the parameter used by the service that display the toast, has all the expected interface, regardless if the error is a string or an object valid to use in the toast. With this solution, we sure the button should work with different thrown errors.

Mayons:
Test OK ! 
Error Toast 1
Error toast2

mauceballos:
Test kibana 7.10 OK

toast-error

mauceballos (approved):

Issue #4105, opened 2022-04-26
TypeError in Management/Rules


juliamagan:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.3   | 7.x     | 4301-1 | Basic, ODFE, Xpack |

| Browser |
| ------- |
| Firefox|

**Description**
If we search a rule file, we get:
```
Uncaught TypeError: this.inputRef is undefined
```

**Preconditions**
1. Demo environment

**Steps to reproduce**
1. Navigate to Management/Rules
2. Click on Manage rules
3. Search a file

gdiazlo:
Happens 4.2.6.

AlexRuiz7:
Partially fixed on #4125 (workaround)

This issue will stay open until we find a better solution.

AlexRuiz7:
Work continues in #4174

PR #4125, opened 2022-04-29
Fixed searchbar suggest handler undefined input ref

asteriscos:
**This is a workaround for the issue #4105** 

The ruleset view and the rule files view share the same input, so the reference of the input does not trigger the useEffect and is lost. To fix this I added a possible pre-existing reference to the constructor of the suggestion component. 

This base issue also causes another problem, because changing the view doesn't reset the input and keeps the previous search. When a previous field has a query it fails on the new view. 

E.g.
Peek 2022-04-29 20-26


matiasmoreno876 (approved):
CR OK ✅ 
TEST OK ✅

Machi3mfl (approved):

PR #4126, opened 2022-04-29
management -> groups -> agent: selectors appear when there are more t…

chantal-kelm:
**In order to test this issue, click 'manage' -> 'groups', choose a group, and click the 'export PDF' button, you will see that the 'select all and deselect all' selectors no longer appear**

**In the image you can see that when the options are 3 or less, the select all and unselect all buttons are not visible.**
Screenshot from 2022-04-29 15-15-01

**In this image you can see that when the options are more than 3 the select all and unselect all buttons are seen**
Screenshot from 2022-04-29 15-16-18


package.json
```diff
@@ -43,7 +43,6 @@
     "angular-animate": "1.7.8",
     "angular-chart.js": "1.1.1",
     "angular-material": "1.1.18",
-    "axios": "^0.21.1",
```
matiasmoreno876, line 46: These dependencies should not be removed.

matiasmoreno876 (approved):
CR OK ✅ 
TEST OK ✅ 

image

Machi3mfl (approved):

PR #4157, opened 2022-05-18
[FIX] Maintain the filters when click on the `Visualize` button of a document field from `<Module>/Events` and redirect to `lens` plugin

Desvelao:
# Description

This PRs avoids removing the filters in the `filterManager` when clicking on the `Visualize` button from `<Module>/Events` and redirects to `lens` plugin. So the visualization maintains the same filters.

It is important to note that the data displayed in the Wazuh plugin, could be less than in `lens` plugin due to the application of some filters, related to permissions about agents or hide the alerts related to the Wazuh managers.

# Test

Using a Kibana 7.16.x or 7.17.x with X-Pack, go to `<Module>/Events`, click on some field of the sidebar and click on the `Visualize` button. The filters displayed in the search bar in the Wazuh plugin and the `lens` plugin should be equal.

Try to add more custom filters before clicking on the `Visualize` button and these should maintain.

Closes https://github.com/wazuh/wazuh-kibana-app/issues/4143


AlexRuiz7 (commented):
LGTM!

Issue #4145, opened 2022-05-13
Improve error management


Machi3mfl:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.x   | 7.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |
| Chrome, Firefox, Safari, etc|

### Description
On this issue, we will analyze all the error treatments.
We need to find all the cases where we catch errors and decide how to throw them to other application methods.

In this issue, we want to improve the error management in our app.
We need to normalize and improve the treatment of errors.

### Steps to find the issue scope:

- [x] Check all the current error treatment
   1. Tracking all error throw errors that don't return an error instance
   2. Searching those methods usage (previous step)
   3. Analyzing possible solutions, best error treatment
**The analysis was documented in the following issue comments:** 
- https://github.com/wazuh/wazuh-kibana-app/issues/4145#issuecomment-1127751874
- https://github.com/wazuh/wazuh-kibana-app/issues/4145#issuecomment-1129001143
- https://github.com/wazuh/wazuh-kibana-app/issues/4145#issuecomment-1129001589
- https://github.com/wazuh/wazuh-kibana-app/issues/4145#issuecomment-1129002088
- https://github.com/wazuh/wazuh-kibana-app/issues/4145#issuecomment-1129002538
- https://github.com/wazuh/wazuh-kibana-app/issues/4145#issuecomment-1129002659

- [x] Create a solution that solves all the current error treatment problems.
- [x] Implement a `first version of Error Handler` and document the solution. https://github.com/wazuh/wazuh-kibana-app/pull/4163
- [ ] Replace the use of error orchestrator in error management solution
- [ ] Apply the new error management solution to a known issue #5210 
- [ ] Attack the error sources and use typescript to make interfaces and create better type error management.
  1. On server-side controllers #5227
  2. On client-side services #5262

Related issue 
- #5227
- #4297

Machi3mfl:
## wz-api-check:
- [public/react-services/wz-api-check.js](https://github.com/wazuh/wazuh-kibana-app/blob/4.3-7.10/public/react-services/wz-api-check.js)

<details>
  <summary>checkStored</summary>
https://github.com/wazuh/wazuh-kibana-app/blob/aa1a8e41a4337610889dd0332c63a1214ce91582/public/react-services/wz-api-check.js#L50-L61
</details>

<details>
  <summary>checkApi</summary>
[...]

PR #4163, opened 2022-05-23
Added new global error treatment (client-side)

Machi3mfl:
This PR,
We need to centralize and normalize the way that we treat the errors thrown in execution time.
With this solution, we will centralize error management and use different artifacts that make easy the error treatment for the developer.
This is the first version of the error management for the client side using react and javascript/typescript patterns.

Solves partially the epic issue: #4145 

For more details about error management checks de `error-management docs`:
- [Main README.md](https://github.com/wazuh/wazuh-kibana-app/blob/feat/4145-request-errors-management/public/react-services/error-management/docs/README.md)
- [Examples README.md](https://github.com/wazuh/wazuh-kibana-app/blob/feat/4145-request-errors-management/public/react-services/error-management/docs/examples/README.md)


public/react-services/error-handler/error-handler.ts
```diff
@@ -0,0 +1,138 @@
...
+// 201 = Created
+// 202 = Accepted
+// 204 = No Content
+// 400 = Bad Request
+// 401 = unauthorized
+// 403 = forbidden
+// 404 = not found
+// 405 = method not allowed
+// 500 = internal server error
+// 501 = not implemented
```
gabiwassan, line 26: Remove this commented code 🤓

Machi3mfl:
## Error sources

- Wazuh API errors
- Elasticsearch Error
- Operational errors (development)
- Axios errors

# Error management components

## Error factory

The `error factory` is responsible to create different instances of error depending on the parameters received.

**The error factory can receive:**
- A `string`
- An `error instance`
- An `error type`: this param defines the error type returned

 The errors returned are defined as the `error type` received.

- ElasticError
- WazuhApiError
- UIError
- etc

## Error handler

The `error handler` is responsible to receive the errors (or strings) and define what type of error will be returned.
After identifying and classifying the parameters received the error factory returns a new error instance.
Always will return an error instance.

# React patterns

The error handler can be implemented using react patterns:

- ### HOC (Higher order component)
- ### Hook

Machi3mfl:
# Errors Classes

The next diagram shows how is the relationship between the different types of errors created.

```mermaid
classDiagram

class iWazuhError {
    <<interface>>
    +Error error
    +string message
    +integer code    
    +handle(amount)
}

iWazuhError <|-- WazuhError : implements
WazuhError <|-- WazuhApiError : extends
WazuhError <|-- WazuhReportingError : extends
WazuhError <|-- ElasticApiError : extends
WazuhError <|-- ElasticError : extends

```

Machi3mfl:
## Unit tests

### Error factory

image

### Error handler

image

### Error handler HOC

image

### Error handler Hook

image

Machi3mfl:
## Tasks

- [ ] Implement `ErrorHandler` in `plugins/wazuh-kibana-app/public/components/common/error-boundary`

public/react-services/error-management/error-handler/error-handler.ts
```diff
@@ -0,0 +1,140 @@
...
+          message: error.message,
+          title: error.message,
+          error: error,
+        },
+        level: 'ERROR',
+        severity: 'UI',
+        display: false,
+        store: false,
+      };
+      ErrorOrchestratorService.handleError(errorLog);
```
Desvelao, line 137: **nitpick**: in both situations `if/else` the `ErrorOrchestratorService.handlerError` is called. You could call it outside `if/else` block and build the parameter used by `ErrorOrchestratorService.handlerError` in the `if/else` blocks instead.

public/react-services/error-management/error-handler/hooks/useErrorHandler.ts
```diff
@@ -0,0 +1,18 @@
...
+  try {
+    let res = await callback();
+    return [res, null];
+  } catch (error) {
+    if (error instanceof Error) {
+      error = ErrorHandler.handleError(error);
+    }
+    return [null, error];
+  }
+};
```
Desvelao, line 18: **thought**: If this is a ReactJS hook, there is something that is missing, the hook can't return asynchronously. For another hand, and ignoring the implementatin problem, it only covers a use case. Could you clarify what is the intention of this hook?

Desvelao:
**thought**: Personally, I don't like the usage of the `ErrorOrchestrator` service. It looks overcomplicated to me.

Desvelao:
**praise**: Good job adding some documentation about error management.

Desvelao:
**thought**: I feel we should discuss how we want to manage the errors and the required implementation.

public/react-services/error-management/error-handler/hooks/useErrorHandler.ts
Machi3mfl: That's right. I've changed this implementation to avoid repeating the error orchestrator call.

public/react-services/error-management/error-handler/error-handler.ts
Machi3mfl: That's right. I've changed this implementation to avoid repeating the error orchestrator call.

public/react-services/error-management/error-handler/hooks/useErrorHandler.ts
Machi3mfl: The use case for this hook is using it to wrap the async calls and avoid the try-catch block for every place. Also, clarify the error catching.
**For instance:** If we wrap a method that calls to the API, we can treat the error for this specific case because the hook returns the error. Right now, we are wrapping many methods within a try-catch block and we don't know which of these methods is throwing the error and can't treat it as separate.

Machi3mfl:
> **thought**: Personally, I don't like the usage of the `ErrorOrchestrator` service. It looks overcomplicated to me.

I think is better to use the existing ErrorOrchestrator to avoid rewriting a new Error Logger. But, we can prepare the implementation to change the ErrorOrchestrator after with the lowest possible cost.

Machi3mfl:
## Next steps

- [ ] Add uses cases and document uses
- [ ] See how to improve the error categorization - change the categorization by error code
- [ ] Add documentation about how the error is shown in the UI

AlexRuiz7:
Branch updated with rebase.

public/controllers/agent/components/agents-table.js
```diff
@@ -459,7 +459,7 @@ export const AgentsTable = withErrorBoundary(
         field: 'id',
         name: 'ID',
         sortable: true,
-        show: true,
+        width: '4%',
```
Desvelao, line 462: **nitpick**: This change is not present in the current `4.5` branch and I don't know if there is a reason of this change.

Maybe this is a problem resolving conflicts, because that values are present before this more recent change: https://github.com/wazuh/wazuh-kibana-app/pull/5143/files#diff-c115a3bedabddfdf963d2ea58acb86e21d3d125de731b17d93622f94c70846bb.

This is happening for all the modifications of `width` of this files. Check the another properties with the branch 4.5 if this was not intentionally changed.

It seems this was changed

public/react-services/error-management/docs/README.md
```diff
@@ -0,0 +1,292 @@
...
+# Error sources
+
+- Operational errors (development) - Native javascript errors
+- Wazuh API errors
+- Indexer Error
+- Http errors
+- Etc
+
+
+Our frontend server-side have a intermedial layer between the frontend and the backend APIs like Indexer and wazuh.
```
Desvelao, line 38: **nitpick**: put `Wazuh` in capital case.

public/controllers/agent/components/agents-table.js
Machi3mfl: Yes, this was added when I made a rebase from 4.4-7.10.2 to 4.5. I've reverted the changes.

AlexRuiz7 (approved):

PR #4196, opened 2022-06-03
[FIX][Kibana][7.16.x-7.17.x] Fix error when building the visualizations

Desvelao:
# Description

This PR fixes an error when building the core visualizations for the plugins for Kibana 7.16.x and 7.17.x.

# Changes
- fix an error when building the core visualizations
- fix error in the `Agents` section caused by a not defined class.

# Test
- Navigate to any section which has a core visualization. This should be displayed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


asteriscos (approved):
LGTM :heavy_check_mark:

yenienserrano (approved):
Test: ✔️

PR #4200, opened 2022-06-03
[FIX] Visualizations are not built due a Types was not set error in a production package

Desvelao:
# Description

This PR fixes an error that appears in the production package when building the visualizations. The problem doesn't appear in development mode as they are rendered correctly.

# Changes
- Revert how the visualization is built
- Add logic to apply the params and uiStateJSON properties of the visualization that due a bung in KIbana 7.10.2 these are not working.

# Screenshots
image

# Test

- The visualizations should be visible in development mode and in a production package without the mentioned error.
- Moreover, the visualizations should renders as expected related to the UI properties.
  - Use case: with data related to the monitoring job of the plugin (agent status), review that the `Evolution` visualization of `Agents` section is displayed with the expected colors for the status labels.
- Review each visualization is visible without problems to see if there is not any problem with the new logic to set some properties of the visualization

Closes https://github.com/wazuh/wazuh-kibana-app/issues/4199

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 
 - Replaced the visualization of `Status` panel in `Agents` [#4166](https://github.com/wazuh/wazuh-kibana-app/pull/4166)
 - Replaced the visualization of policy in `Modules/Security configuration assessment/Inventory` [#4166](https://github.com/wazuh/wazuh-kibana-app/pull/4166)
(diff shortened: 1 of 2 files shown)


asteriscos (approved):
CR: :heavy_check_mark: 
Test: :heavy_check_mark: 
image

Machi3mfl (approved):
CR ✅
Test ✅

AlexRuiz7 (approved):
LGTM!

Issue #4218, opened 2022-06-06
TypeError: e.meta.value is not a function 


yenienserrano:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.3.x   | 7.16.x 7.17.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |
| Chrome, Firefox, Safari, etc|

**Description**

In the Regulatory Compliance or Mitre Attacks modules, when you enter, the error message appears `(TypeError: e.meta.value is not a function)` in the console and allows you to remove the filters from the view.

**Steps to reproduce**
1. Navigate to 'Regulatory Compliance or Mitre Attacks modules'
2. Look at the module filters 
3. See the console for the error

**Expected Result**
1. 

![image](https://user-images.githubusercontent.com/63758389/172180488-49b3851c-5168-436b-a07e-bace4f338eff.png)

**Actual Result**
1. 

![image](https://user-images.githubusercontent.com/63758389/172180819-b8fe15d1-e92b-488f-9783-e27cf79d5a76.png)

**Screenshots**

**Additional context**
Add any other context about the problem here. Here you can paste log entries too or any other useful information that may help with the issue.

yenienserrano:
What could be found out is that in this block of code
```
const objValue = moduleFilter.query?.match_phrase
            ? moduleFilter.query.match_phrase[objKey].query
            : moduleFilter.meta.value();
```

`moduleFilter.meta.value` sometimes it is a function and sometimes a string.
What will be proposed is to handle both occasions even if we are not sure that it will ever come into when it is a function.
Because when this is executed it is because there is no match in the match_phrase and in the cases we saw it only does not match when it is a string.

yenienserrano:
Solve: 

- #4219

PR #4219, opened 2022-06-06
String or function validation

yenienserrano:
**Description:**

A validation is added to a section of the filter code, since sometimes `moduleFilter.meta.value` is a function and sometimes a string.
When it was a string it lost the check to be able to eliminate or not the filter.

For example in the Mitre Attack module

**Issue:**

- #4218 

**Test:**

1. Navigate  to 'Mitre Attack module'
2. Look at the module filters can''t be removed.
3. View the console without error

**Screenshot:**

image


Desvelao (approved):
**review**
code :heavy_check_mark: 
test :heavy_check_mark: 

screenshots
image

asteriscos (approved):
CR: :heavy_check_mark: 
Test: :heavy_check_mark: 

image

PR #4240, opened 2022-06-10
[BUG][Modules][Reports] Error when generating a report after changing the selected agent

Desvelao:
# Description

This PR fixed an error when generating a module report after changing the selected agent.

Due to some side effects, it could not get the applied filters in the visualizations because the visualization handlers are removed of the `VisHanders` singleton class and the request done to a plugin endpoint sent properties values with no expected value.

Closes https://github.com/wazuh/wazuh-kibana-app/issues/4230

# Test

- Go to `Modules/<Module>` and generate a report
- Select an agent with the `Explore agent` button and generate a module report clicking on the `Generate report`
- Change the selected agent to another one, and generate a module report. This should be done without problem. Review the request that is sending the expected values and the generated PDF.

### DIFF
--- a/public/components/visualize/wz-visualize.js
+++ b/public/components/visualize/wz-visualize.js
@@ -77,12 +77,21 @@ export const WzVisualize = compose(
     }
 
     async componentDidUpdate(prevProps) {
-      if (prevProps.isAgent !== this.props.isAgent) {
+      // When in a module changing from without selected agent to selected agent or viceversa, set the visualizations
+      // and remove the visualizations handlers of the VisHandlers singleton.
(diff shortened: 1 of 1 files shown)


Mayons95:
TEST OK
https://user-images.githubusercontent.com/104914131/173120342-53965faf-a3dc-4991-a3f2-f95aa9859a31.mp4

Mayons95 (approved):

Mayons95:
I found a related bug, it's not possible to delete the generated reports.
As you could see in the video, a request is responding with a 400 error code.

AlexRuiz7 (approved):
- Code review: ✔️ 
- Test: ✔️

PR #4280, opened 2022-06-17
Display a message related to missing user permissions when finding index patterns in a route resolver

### LINKED ISSUE #4216: Login attemp with user without any role
| 4.3.x   |  -     | - |  - |

**Description**
When logging in with an user with no role, the UI crashes

image

**Preconditions**
1. Create a new user using wazuh-dashboard security plugin
2. Do not assign any role to the user

**Steps to reproduce**
1. Log in with the created user

**Expected Result**
1. The user logs in

**Actual Result**
1. The UI crashes

Desvelao:
# Description

This PR adds a more descriptive message when finding the index patterns and the user has no permission to do it.

Closes  #4216

## Screenshot

image

## Notes

The proposed fix goes against the last work in frontend error management. We should consider if we want to include it or we should redefine how the errors are being managed to let this case.

# Test

Using OpenDistro for Elasticsearch or Wazuh dashboard:
1. create a new user without mapping any permissions
2. enable the `run_as` in the Wazuh API host entry of the Wazuh plugin. Restart Kibana/Wazuh dashboard to be sure the configuration is taken.
3. login using the created user, using `security_tenant=global` in the URL. For example: `https://localhost/5601/app/wazuh?security_tenant=global`
4. after the plugin loaded, it should redirect to the `blank-screen` and it should display a message related to the user permissions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 
 - Added to the interface API messages in the Ruleset test module [#4244](https://github.com/wazuh/wazuh-kibana-app/pull/4244)
 - Added authorization prompt in Mitre > Intelligence [#4261](https://github.com/wazuh/wazuh-kibana-app/pull/4261)
+- Added a more descriptive message when there is an error related to the user permissions when getting the list of index patterns in a route resolver [#4280](https://github.com/wazuh/wazuh-kibana-app/pull/4280)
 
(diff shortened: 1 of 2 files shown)


yenienserrano (approved):
CR: ✔️ 
Test: ✔️ 

image

chantal-kelm (approved):
CR: ✔️ 
Test: ✔️ 

Screenshot from 2022-07-04 11-39-33

PR #4350, opened 2022-07-21
Fixed errors in group reports

### LINKED ISSUE #4348: Group report generation fails randomly
| 4.3.5-1 and 4.3.6-1   | 7.14.4     | 4xxx | Basick |

| Browser |
| ------- |
| Firefox |

**Description**
When trying to generate a group report, sometimes it fails to get generate the report when the configurations component is selected, with the following error:
```json
{"statusCode":500,"error":"Internal Server Error","message":"5029 - Reporting was aborted (Cannot read property 'length' of undefined)"}
```

**Preconditions**
1. Create a group with 1+ agents

**Steps to reproduce**
1. Navigate to groups tab at `https://ip/app/wazuh#/manager/?tab=groups`
2. Click on the Group you want to generate
3. Click on `Generate reports`
Note: The error occurs randomly, and it can be triggered by selecting and deselecting the `configurations` component checkbox. It is more clearly reproduced on Postman.

**Expected Result**
1. Group report fails to generate

**Actual Result**
1. Group report is generated successfully

**Screenshots**
imagen
imagen
imagen
- Postman exampe
imagen

**Additional context**
The error was found in `kibana-wazuh-app` and `wazuh-dashboard version 4.3.5-1 (installed from documentation) and 4.3.6-1 (installed from packages), during testing for Wazuh-qa Issue  [#3100](https://github.com/wazuh/wazuh-qa/issues/3100)

Desvelao:
# Description

This PRs fixes some problems when generating a group reports:
- Error due to accessing to a property of undefined
- The agent in group table doesn't display all the agents in the group.

Closes #4348

# Changes
- Fixed an error due to accessing to a property of undefined when the API request to get the group configuration failed due to token expiration or another error.
- Changed styles of `Agents in groups` title
- Fixed a problem with missing agents in the table for a group report.
  Now it gets all the agents.
- Removed unnecessary API request
- Optimized API requests to get the agent information required to print
  the agents in group's table
- Removed `manager_host` deprecated field ([more information](https://github.com/wazuh/wazuh-kibana-app/commit/320825de99791fc76fb826712f4bb37d557672db#diff-980834c867463b40f86ef51bd0a723ef2063b9c9ef7cead6cd0c8649dd9bc1ddR647-R648)):
  - Removed of monitoring template mapping
  - Removed field meaning
  - Removed displaying its value reports if the field `manager` doesn't exist

# Test

## Manual tests

Scenario: Error due to accessing to a property of undefined
Given: Logged user in Management/Groups/<Group>/Information, wait the time to the Wazuh API token expires.
When: The user click on "Export PDF" button.
Then: The PDF should be generated without errors. Check the generated PDF.

Scenario: Missing agents in the agent in group table of group report.
Given: group with more than 500 agents.
When: The user clicks on "Export PDF" button.
Then: The PDF should be genrated without errors and it should contains all the agents in the group displayed in a table.

### DIFF
--- a/server/controllers/wazuh-reporting.ts
+++ b/server/controllers/wazuh-reporting.ts
@@ -146,7 +146,7 @@ export class WazuhReportingCtrl {
           printer,
           isAgents,
           apiId,
-          section === 'groupConfig' ? tab : null
+          section === 'groupConfig' ? tab : ''
         );
       }
 
@@ -191,7 +191,7 @@ export class WazuhReportingCtrl {
    * @param {Array<Strings>} ids ids of agents
    * @param {String} apiId API id
    */
-  private async buildAgentsTable(context, printer: ReportPrinter, agentIDs: string[], apiId: string, groupID: string | null = null) {
+  private async buildAgentsTable(context, printer: ReportPrinter, agentIDs: string[], apiId: string, groupID: string = '') {
     const dateFormat = await context.core.uiSettings.client.get('dateFormat');
     if ((!agentIDs || !agentIDs.length) && !groupID) return;
     log('reporting:buildAgentsTable', `${agentIDs.length} agents for API ${apiId}`, 'info');
--- a/server/controllers/wazuh-reporting.ts
+++ b/server/controllers/wazuh-reporting.ts
@@ -134,16 +134,8 @@ export class WazuhReportingCtrl {
         } else if (section === 'groupConfig') {
           printer.addContent({
             text: 'Agents in group',
-            style: { fontSize: 14, color: '#000' },
-            margin: [0, 20, 0, 0],
+            style: 'h1',
           });
-          if (section === 'groupConfig' && !Object.keys(isAgents).length) {
-            printer.addContent({
-              text: 'There are still no agents in this group.',
-              style: { fontSize: 12, color: '#000' },
-              margin: [0, 10, 0, 0],
-            });
-          }
         }
         printer.addNewLine();
       }
@@ -154,7 +146,7 @@ export class WazuhReportingCtrl {
           printer,
           isAgents,
           apiId,
-          section === 'groupConfig' ? tab : false
+          section === 'groupConfig' ? tab : null
         );
       }
 
@@ -188,8 +180,6 @@ export class WazuhReportingCtrl {
           style: 'standard',
         });
       }
-
-      return;
     } catch (error) {
       log('reporting:renderHeader', error.message || error);
       return Promise.reject(error);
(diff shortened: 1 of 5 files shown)


server/controllers/wazuh-reporting.ts
```diff
@@ -261,19 +238,37 @@ export class WazuhReportingCtrl {
...
+              os: (agent.os && agent.os.name && agent.os.version) ? `${agent.os.name} ${agent.os.version}` : '',
+              lastKeepAlive: moment(agent.lastKeepAlive).format(dateFormat),
+              dateAdd: moment(agent.dateAdd).format(dateFormat)
+            }
+          }),
+        });
+      }else if(!agentsData.length && groupID){
+        // For group reports when there is no agents in the group
+        printer.addContent({
+          text: 'There are still no agents in this group.',
```
AlexRuiz7, line 267: A more appropriate sentence would be: `There are no agents in this group yet` or simply `There are no agents in this group`

server/controllers/wazuh-reporting.ts
```diff
@@ -201,57 +191,44 @@ export class WazuhReportingCtrl {
    * @param {Array<Strings>} ids ids of agents
    * @param {String} apiId API id
    */
-  private async buildAgentsTable(context, printer: ReportPrinter, agentIDs: string[], apiId: string, multi = false) {
+  private async buildAgentsTable(context, printer: ReportPrinter, agentIDs: string[], apiId: string, groupID: string | null = null) {
```
AlexRuiz7, line 194: Be careful using `null`

That `null = null` does not look very well.

server/controllers/wazuh-reporting.ts
```diff
@@ -154,7 +146,7 @@ export class WazuhReportingCtrl {
           printer,
           isAgents,
           apiId,
-          section === 'groupConfig' ? tab : false
+          section === 'groupConfig' ? tab : null
```
AlexRuiz7, line 149: Be careful using `null`

`false` would be more appropriate, unless there is justified reason to use `null` instead.

AlexRuiz7 (changes requested):

server/controllers/wazuh-reporting.ts
Desvelao: Code:
```
groupID: string | null = null
```

- Parameter definition schema
```
parameter name : TS type definition = default value
```

This means:
- parameter name: `groupID`
- type definition is: `string | null`
- default value: `null`

server/controllers/wazuh-reporting.ts
Desvelao: That parameter can be a group ID ( `string` ) so using a boolean could be confusing, but if its value is managed correctly we could use any falsy values. I considered changing it to `null`.

server/controllers/wazuh-reporting.ts
AlexRuiz7: And wouldn't it be better to simply have a single type, which is string, and set its default value to an empty string?

Using combined types for a simple parameter as that one seems overcomplicated.

Desvelao: Yes, it is another option. I will change it.

Desvelao: Changed. Thank you for the suggestion!

server/controllers/wazuh-reporting.ts
Desvelao: This message was included in the PDF report. The message was not changed. https://github.com/wazuh/wazuh-kibana-app/pull/4350/files#diff-e344f646f1844581c2ba5159d750a8f075c476e56350e3bcc6018f87136dd973L142

But I will take your suggestion. Thank you!

Desvelao: Changed.

AlexRuiz7 (approved):
LGTM!

yenienserrano (approved):
LGTM!

PR #4358, opened 2022-07-25
Fix creating PDF reports in x-pack without authentication context 

asteriscos:
# Description

This PR fixes when x-pack is detected with no `authContext` available and no `hashUsername` is provided. Therefore in this kind of context, creating PDF reports fails.


Desvelao (approved):
**review**
Code :heavy_check_mark: 
Test :heavy_check_mark:

Machi3mfl (approved):
CR ✅
Test ✅

PR #4359, opened 2022-07-26
Fix modules settings persistence between updates

Desvelao:
# Description

This PR fixes that the host extension configuration was lost when changing the installed plugin (changed version or revision) because the registry file was rebuilt with an empty configuration for the `hosts` field. Now, the `hosts` configuration is kept and migrates the extension configuration when changing the installed plugin.

Extension configuration migration:
- Keep the supported and existing extension with the configured value.
- Add the supported and missing extensions in the previous configuration
- Remove not supported extensions by the installed plugin

# Changes
- Migrated the existing host configuration in the registry file when changing the
  plugin version or revision, instead of removing it when it was rebuilt.
- Removed not necessary return statements
- Removed not necessary try/catch block
- Improved logging
- Created tests for the migration registry file

# Test

## Manual
- Same plugin version and revision, restart the plugin platform. The registry file should not be changed.
- Different plugin version or revision, restart the plugin platform. The registry file should be rebuilt and keeps the host configuration.
- Different plugin version or revision, removes some extension of any host in the registry file, restart the plugin platform. The registry file should be rebuilt and keeps the host configuration. The removed extension should be included in the host and have the default value.

## Automatic

A new test file was created. Run the test with:

```
# in the plugin root directory
yarn test:jest server/start/initialize/index.test.ts
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to the Wazuh app project will be documented in this file.
 
+## Wazuh v4.3.7 - Kibana 7.10.2, 7.16.x, 7.17.x - Revision 4308
+
+### Fixed
+
+- Fixed the `hosts` data stored in the registry file was lost when changing the installed plugin. Migrate the `extensions` data [#4359](https://github.com/wazuh/wazuh-kibana-app/pull/4359)
+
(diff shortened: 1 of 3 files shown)


yenienserrano (approved):
LGTM! ✅ 

image

Machi3mfl (approved):
Test ✅
CR ✅

PR #4362, opened 2022-07-28
TypeError resolved

chantal-kelm:
## Research
In `public/plugin.ts` a Get request is done with a Kibana resource `core.http.get(/api/check-wazuh)`.

That resource fails in Chrome and Firefox browsers.

The Kibana core get resource is set up to catch the error with an `HttpFetchError` class that is a child of the Node Error class. Which is compatible only with browsers with V8 engine.
For this reason in Chrome it shows the error in the console as a debug and in Firefox it shows it in the console as an unrecognized error.

In the `http_fetch_errors.ts` and `globals.d.ts` files, kibana mentions V8 compatibility and that it will only work in browsers that use it. Attached screenshots
https://github.com/v8/v8/wiki/Stack%20Trace%20API#customizing-stack-traces

Screenshot from 2022-08-03 07-10-11
Screenshot from 2022-08-03 07-11-00

## Proposed solution

Change the Get request that was made with a Kibana core.http.get(/api/check-wazuh) resource to the WzRequest.genericReq resource and it no longer fails, also add a test capture to public/plugin.ts that wraps the request and in case of failure, the error is detected when the browser does not work with the V8 engine.

## Test

- Click on the hamburger menu
- Click on dashboard
- Click on home
- Click on Wazuh
- Check that the console does not show: `TypeError: NetworkError when attempting to fetch resource`

https://user-images.githubusercontent.com/99441266/181495074-7394bc97-b81b-43ae-ac15-00dc1e134231.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 ### Fixed
 
 - Fixed path in logo customization section [#4352](https://github.com/wazuh/wazuh-kibana-app/pull/4352)
-- Fixed a TypeError in Firefox. Change the Get request that was made with a Kibana core.http.get(/api/check-wazuh) resource to the WzRequest.genericReq resource and add a test capture to public/plugin.ts wrapping the request so that the error is detected when the navigator does not work with the V8 engine [#4362](https://github.com/wazuh/wazuh-kibana-app/pull/4362)
+- Fixed a TypeError in Firefox. Change the Get request that was made with a Kibana core.http.get(/api/check-wazuh) resource to the WzRequest.genericReq resource and it no longer fails, also add a test capture to public/plugin.ts that wraps the request and in case of failure, the error is detected when the browser does not work with the V8 engine. [#4362](https://github.com/wazuh/wazuh-kibana-app/pull/4362)
 - Fixed an error of an undefined username hash related to reporting when using Kibana with X-Pack and security was disabled [#4358](https://github.com/wazuh/wazuh-kibana-app/pull/4358)
 
 ## Wazuh v4.3.6 - Kibana 7.10.2, 7.16.x, 7.17.x - Revision 4307
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 ### Fixed
 
 - Fixed path in logo customization section [#4352](https://github.com/wazuh/wazuh-kibana-app/pull/4352)
-- Fixed a TypeError in Firefox [#4362](https://github.com/wazuh/wazuh-kibana-app/pull/4362)
+- Fixed a TypeError in Firefox. Change the Get request that was made with a Kibana core.http.get(/api/check-wazuh) resource to the WzRequest.genericReq resource and add a test capture to public/plugin.ts wrapping the request so that the error is detected when the navigator does not work with the V8 engine [#4362](https://github.com/wazuh/wazuh-kibana-app/pull/4362)
 - Fixed an error of an undefined username hash related to reporting when using Kibana with X-Pack and security was disabled [#4358](https://github.com/wazuh/wazuh-kibana-app/pull/4358)
 
 ## Wazuh v4.3.6 - Kibana 7.10.2, 7.16.x, 7.17.x - Revision 4307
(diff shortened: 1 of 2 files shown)


AlexRuiz7 (changes requested):
This PR does not solve the problem. Take into account that is error might be a symptom of something else, like the App not being properly mounted / unmounted, not an isolated error.

Also, be thorough on the explanation on the cause of the problems and the work done to solve it. 

In this case, incompatibilities with the V8 engine on Chrome and Firefox browsers are pointed as the cause of the error, but not a single reference supporting this theory is provided.

public/plugin.ts
```diff
@@ -28,70 +28,67 @@ import { AppState } from './react-services/app-state';
...
+          //Check is user has Wazuh disabled
+          const response = await WzRequest.genericReq(
+            'GET',
+            `/api/check-wazuh`,
+          )
+          params.element.classList.add('dscAppWrapper', 'wz-app');
+          const unmount = await renderApp(innerAngularName, params.element);
+          //Update if user has Wazuh disabled
+          this.stateUpdater.next(() => {
+            if (response.isWazuhDisabled) {
```
Desvelao, line 74: **issue**: the `response` variable doesn't have the endpoint response directly, so it will not work as expected.

Desvelao, line 74: **issue**: the `response` variable doesn't have the endpoint response directly when using `WzRequest.genericReq` service, so it will not work as expected.

Desvelao (changes requested):
**issue**: the `response` variable doesn't have the endpoint response directly when using `WzRequest.genericReq` service, so it will not work as expected.

chantal-kelm:
Fixed access to the isWazuhDisabled property and tested that it works by creating a user with a disabled role.

Screenshot from 2022-08-05 10-16-55
Screenshot from 2022-08-05 10-23-50

public/plugin.ts
```diff
@@ -46,14 +46,17 @@ export class WazuhPlugin implements Plugin<WazuhSetup, WazuhStart, WazuhSetupPlu
       title: 'Wazuh',
       icon: core.http.basePath.prepend(getThemeAssetURL('icon.svg', UI_THEME)),
       mount: async (params: AppMountParameters) => {
+        try {
+          await core.http.get(`/api/check-wazuh`);
```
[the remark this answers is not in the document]
chantal-kelm: This comment was resolved with this commit [a1ca4f5](https://github.com/wazuh/wazuh-kibana-app/pull/4362/commits/a1ca4f5e4d8de540d41454fdd9b617fa25dfa39c)

public/plugin.ts
```diff
@@ -92,6 +95,9 @@ export class WazuhPlugin implements Plugin<WazuhSetup, WazuhStart, WazuhSetupPlu
         return () => {
           unmount();
         };
+      }catch(error){
+        console.debug(error);
+      }
```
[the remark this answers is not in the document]
chantal-kelm: This comment was resolved with this commit [a1ca4f5](https://github.com/wazuh/wazuh-kibana-app/pull/4362/commits/a1ca4f5e4d8de540d41454fdd9b617fa25dfa39c)

public/plugin.ts
chantal-kelm: This comment was resolved with this commit [8be9602](https://github.com/wazuh/wazuh-kibana-app/pull/4362/commits/8be960288a82421ccff10942a6e56de52ab5a956)

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 ### Fixed
 
 - Fixed path in logo customization section [#4352](https://github.com/wazuh/wazuh-kibana-app/pull/4352)
+- Fixed a TypeError in Firefox [#4362](https://github.com/wazuh/wazuh-kibana-app/pull/4362)
```
AlexRuiz7, line 10: This is too generic.

My suggestion is: 
> A compatibility problem with browsers using a different engine than V8 has been fixed. This problem affected Firefox and Chrome browsers.

AlexRuiz7 (approved):
- Code review: ✔️ 
- Test: ✔️ 

**Note**: check the changelog!

Machi3mfl (approved):
CR ✅
Test ✅

CHANGELOG.md
chantal-kelm: This comment was resolved with this commit [f944eeb](https://github.com/wazuh/wazuh-kibana-app/pull/4362/commits/f944eebb60244be1957eccefc17e665a412e19b1)

PR #4459, opened 2022-09-07
Updating XML selector for decoder file

Mayons95:
After executing a complete execution of the automated test suite, it was detected a few errors on Kibana 7.16.3 with the Xpack plugin, so this PR is to solve the detected problem.

Closes: https://github.com/wazuh/wazuh-kibana-app/issues/4460

Executed Test:
image
image

This PR covers the solution to the edit-decoder-without-saving test.
After checking the existing selector, I saw that it was two web elements with the same selector, which was causing the problem.
Modify the selector and the problem is solved, test both xpack and wzd
Changed selector
image


AlexRuiz7 (approved):
LTGM!

Machi3mfl (approved):
LGTM!

PR #4500, opened 2022-09-13
Add all API endpoints to the new Docker environments configuration

yenienserrano:
**Description:**

All possible endpoints that appear in the API reference are added to have the initial base and only have to add the necessary responses.


AlexRuiz7 (approved):
LGMT!

PR #4507, opened 2022-09-13
Add switch to toggle the use of custom branding

asteriscos:
# Description

This PR adds a new configuration setting to enable or disable the whole branding customization. This makes it easier to switch on or off all the white-labeling configurations without changing or deleting all the custom branding configurations. 
Changing `customization.enabled` configuration to `true` or `false` is enough to deactivate all the branding customizations.

- Add a new setting:
  - `customization.enabled`: Enable or disable the application of others `customization.*` settings.
  
  
| key                                   | description                                                                                              | values             | requirements on change |
| ---                                     | ---                                                                                                            | ---                    | --- |
| `customization.enabled`  | Enable or disable the application of others `customization.*` settings    | `true`/`false`    | reload browser tab |
  
- Add tests related to the setting

Closes https://github.com/wazuh/wazuh-kibana-app/issues/4506

## Screenshot
localhost_5601_app_wazuh (1)

# Test

## Manual tests

**Scenario**: Modify the value of `customization.enabled` setting through `Settings/Configuration`
  **When** the user modifies the `custmization.enabled` setting and saves the configuration.
  **Then** the configuration file should be updated and the frontend
  
**Scenario**: Modify the value of `customization.enabled` setting through an API request with a valid value.
  **When** the user does an API request to update the setting `customization.enabled` with a valid value: `true` or `false`
  **Then** the response should be successful
  **And** the configuration file should be updated
  
**Scenario**: Modify the value of `customization.enabled` setting through an API request with an invalid value.
  **When** the user does an API request to update the setting `customization.enabled` with an invalid value.
    **Then** the response should be a bad request
    
**Scenario**: Customization is applied when the `customization.enabled` setting is enabled.
  **Given** the `customization.enabled` setting value is `true` and the `customization.*` settings have custom values
  **When** the user goes to the plugin,
  **Then** the menu displays the custom image
  **And** the plugin category displays the custom image
  **And** the plugin health check displays the custom image
  **And** the PDF report image displays the custom image
  **And** the PDF header displays the custom text
  **And** the PDF footer displays the custom text
  
**Scenario**: Customization is not applied when the `customization.enabled` setting is disabled.
  **Given** the `customization.enabled` setting value is `false` and the `customization.*` settings have custom values
  **When** the user goes to the plugin,
  **Then** the menu displays the default image
  **And** the plugin category displays the default image
  **And** the plugin health check displays the default image
  **And** the PDF report image displays the default image
  **And** the PDF header displays the default text
  **And** the PDF footer displays the default text
  
  ***
  To manually test the configuration endpoint here are 2 scripts to update different configuration values.
  
[switch-toggle.zip](https://github.com/wazuh/wazuh-kibana-app/files/9735686/switch-toggle.zip)

To use it, first unzip the files in the same folder and allow both _executing the file as a program_. Afterwards you can run the `manual-test-switch-toggle.sh` script.

You can also use **wz-api-conf.sh** script with custom configuration parameters, like:
```
\- s setting_name
\- v setting_value
```

eg.
`./wz-api-conf.sh -s disabled_roles -v [\"disabled\"]`

#### Output

```

./manual-test-switch-toggle.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   398  100   398    0     0    418      0 --:--:-- --:--:-- --:--:--   417
{"customization.enabled":25000}
{"statusCode":400,"error":"Bad Request","message":"[request body.customization.enabled]: expected value of type [boolean] but got [number]"}  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   398  100   398    0     0    884      0 --:--:-- --:--:-- --:--:--   884
{"customization.enabled":true}
{"data":{"requiresRunningHealthCheck":false,"requiresReloadingBrowserTab":false,"requiresRestartingPluginPlatform":false,"updatedConfiguration":{"customization.enabled":true}}}  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   398  100   398    0     0    852      0 --:--:-- --:--:-- --:--:--   852
{"customization.enabled":false}
{"data":{"requiresRunningHealthCheck":false,"requiresReloadingBrowserTab":false,"requiresRestartingPluginPlatform":false,"updatedConfiguration":{"customization.enabled":false}}}

```
  
## Automatic tests
    
- Tests
  - `common/services/settings.test.ts`
  - `server/routes/wazuh-utils/wazuh-utils.test.ts`
  - `common/plugin-settings.test.ts`
  
```shell
yarn test:jest common/services/settings.test.ts server/routes/wazuh-utils/wazuh-utils.test.ts common/plugin-settings.test.ts
```


Desvelao:
I added some manual tests to the initial comment.

Desvelao:
# Update

- Update the development branch with the last changes of the base branch.

Desvelao:
# Update
- Add tests
- Sync with the integration branch

Desvelao:
# Update
- Update the development branch with the base branch
- Add a test for the `customization.enabled` setting in the form input

Desvelao:
# Update
- Update the development branch with the base branch

yenienserrano (approved):
LGTM

AlexRuiz7 (dismissed):
LGTM

Desvelao:
# Update
I updated the development branch with the last changes of the base branch and I fonund a problem.

The settings related to the customization of reports don't work, the PDF is printed using the default values. This is related to the `getConfiguration` backend service, that only gets the configuration defined in the file. When the `customization.enabled` setting is not defined, the [service used to get the configuration](https://github.com/wazuh/wazuh-kibana-app/pull/4507/files#diff-86349d2dafbe7649ca8eb6737bb94a0aa046b3300aa40b5a34e07de4f6ebd234R139-R147) depending on if the customization is enabled or not.

I was in a meeting with @asteriscos reviewing the code and we consider refactoring it. The service should return the mixed configuration: default and read in the configuration file. With this change, we can remove some code management done in some sections of the backend.

Desvelao:
# Update

I am analyzing the problem mentioned above. https://github.com/wazuh/wazuh-kibana-app/pull/4507#issuecomment-1297115591

Due to the plugin having some settings with hidden default values, it seems that is not possible to refactor and simplify the `getConfiguration` service to return the plugin configuration. We have 2 different requirements:
- In `Settings/Configuration, the settings don't display the hidden default values.
- Using the settings, the hidden default values are required.

For example, when the user has not defined the `customization.reports.header` setting, in `Settings/Configuration` it is displayed as an empty string while when it is used in the backend to build the PDF reports, for this case, this will have the `info@wazuh.com\nhttps://wazuh.com` value.

The hidden default values exist because the approach of the recent plugins is that if the user configures these settings, then it will use these values. If these settings are not defined and have a hidden default value, then it will use them instead.

Desvelao:
# Update

The `customization.logo.sidebar` is changed without requiring reloading the browser tab.

> This solves an indirect problem when the `customization.enabled` setting is changed from the UI, which should take into account the rest of the `customization.*` settings to display some action.
>
> This change enhances the UX because the change is instant and doesn't require reloading the browser tab.

asteriscos:
> There's an issue in `public/plugin.ts` when the app loads and makes a request to obtain any possible custom logos, it does not evaluate if the customization switch is enabled. The endpoint `api/logos` has to include the `customization.enable` conditional to centralize the source of truth.

Fixed in https://github.com/wazuh/wazuh-kibana-app/pull/4507/commits/6ffbdb6aeef9bfd331d3059fd8b64316ee1ff624

public/plugin.ts
```diff
@@ -57,9 +58,9 @@ export class WazuhPlugin implements Plugin<WazuhSetup, WazuhStart, WazuhSetupPlu
       id: `wazuh`,
       title: 'Wazuh',
       icon: core.http.basePath.prepend(
-        logosInitialState?.logos?.[SIDEBAR_LOGO] ?
+        logosInitialState?.logos?.[SIDEBAR_LOGO] !== getSettingDefaultValue(SIDEBAR_LOGO) ?
           getAssetURL(logosInitialState?.logos?.[SIDEBAR_LOGO]) :
-          getThemeAssetURL('icon.svg', UI_THEME)),
+          getThemeAssetURL(logosInitialState?.logos?.[SIDEBAR_LOGO], UI_THEME)),
```
Desvelao, line 70: **nitpick**: This `icon` is the plugin icon, that currently is not displayed on the UI because it only displays the category icon.

Desvelao: I will revert this change.

Desvelao:
# Update
- This change was reverted https://github.com/wazuh/wazuh-kibana-app/pull/4507#issuecomment-1297354125. The `customization.logo.sidebear` requires reloading the browser tab when this is changed.
- Add the reloading browser tab requirement when changing the `customization.enabled` setting.

I was researching how to avoid the user having to reload the browser tab when changing some of these settings, but the changes in the `customization.enabled` requires to check the value of the `customization.logo.sidebar` and changing or resetting to default the logo through the plugin state updater. The applied solution reduces the complexity of the logic.

common/services/settings.ts
```diff
@@ -130,3 +130,21 @@ export function formatLabelValuePair(label, value){
...
+  const isCustomizationEnabled = typeof configuration['customization.enabled'] === 'undefined'
+    ? getSettingDefaultValue('customization.enabled')
+    : configuration['customization.enabled'];
+  const defaultValue = typeof overwriteDefaultValue !== 'undefined'
+    ? overwriteDefaultValue
+    : getSettingDefaultValue(settingKey);
+
+	return ( isCustomizationEnabled && settingKey.startsWith('customization') && settingKey !== 'customization.enabled')
+		? (configuration[settingKey] ?? defaultValue)
+		: defaultValue;
```
AlexRuiz7: I have several suggestions for improvement for this function:

1. The name. It is too long because it gives information that doesn't matter to the user. The goal (as I understand it) is to get from the `configuration` "object" the option given by `settingKey` (as simple as a getter). The `dependOnCustomizationIsEnabled` is an implementation detail of the function. I, as a programmer, want to get that setting, how the function gets it doesn't matter to me. It breaks the abstraction a bit in my opinion and leaves a rather unattractive name.

2. The types. I understand that `configuration` should be at best a type defined by us, or at least an object. On the other hand, the `overwriteDefaultValue` sounds like a boolean to me, but... then, looking at the code, I see that it seems to be the value that I want the setting to have (?). Even looking at the code, I don't see what it does. Finally, it is not defined what the function returns (the type of the return value is missing)

3. Ternary operator. It's ok to use it in simple situations, but not when it makes the code more messy and less understandable. It is part of our [STYLEGUIDE](https://github.com/wazuh/wazuh-kibana-app/blob/master/STYLEGUIDE.md#only-use-ternary-operators-for-small-simple-code). From my point of view, less code is not always better, it is preferable to have a few more lines if we gain readability.

4. Documentation. It doesn't do much good to add a documentation header to the function if it is not properly filled in, or if it is not updated if we modify the code. In the end, we have documentation that says the code does one thing when it is really doing something else. It just adds more mess and more wasted time to understand it when we go back through this block of code.

AlexRuiz7 (commented):
- Code Review: 🟠 (ok, with comments)
- Test: 🟢

AlexRuiz7 (dismissed):

common/services/settings.ts
Desvelao: Solved in https://github.com/wazuh/wazuh-kibana-app/pull/4507/commits/7ad0476f4ff038fe37195536ecdb4416d29b6c60

AlexRuiz7 (approved):

AlexRuiz7 (approved):

Mayons95 (approved):
LGTM
TR: :heavy_check_mark: 

<details>

[Screencast from 02-11-22 12:08:55.webm](https://user-images.githubusercontent.com/104914131/199527975-3f18b847-4f89-45d9-9a9c-f5aaa2225779.webm)

[Screencast from 02-11-22 12:11:15.webm](https://user-images.githubusercontent.com/104914131/199528014-20402aed-6101-4fb1-b9cc-2523d2e1c65a.webm)


</details>

Issue #4299, opened 2022-06-30
Enhance the message displayed when there is a mismatch in API and plugin versions


Desvelao:
| Wazuh | Elastic | Rev  | Security |
| ----- | ------- | ---- | -------- |
| 4.x   | 7.x     | 4xxx | Basic, ODFE, Xpack |

| Browser |
| ------- |
| Chrome, Firefox, Safari, etc|

**Description**
Some users don't know what is the problem and how to solve it when they see the error related to mismatching API and plugin versions.

Tasks:
- [x] Find a better way to inform the problem and how to solve it to the user
- [x] Remove the `<a>` tag of the message

**Screenshots**
![image](https://user-images.githubusercontent.com/34042064/176664257-4377e3eb-c49d-4324-a136-9b8c041cc7c5.png)

Tostti:
Currently working on this issue. 

Analyzing it, I think that the best approach would be to make a shorter message with the important information and a link to the Troubleshooting documentation, in which we should specify in detail how to solve the issue.

Example of the new message:
![image](https://user-images.githubusercontent.com/42900763/190483635-bf0b9216-d0f4-4885-b624-1746442a0240.png)

Tostti:
Blocked by https://github.com/wazuh/wazuh-documentation/pull/5594

Tostti:
Merged corresponding PRs

PR #4529, opened 2022-09-15
Change message for API version mismatch in Kibana

Tostti:
Solves #4299 
# Important information
On PR https://github.com/wazuh/wazuh-kibana-app/pull/4103, a method was changed to request the API endpoint. Only for testing purposes of this particular issue, it is possible to replace line 50 of `server/lib/api-interceptor.ts` with the following: `method: !!authContext ? 'POST' : 'GET',` to avoid getting an error due to a not allowed method and trigger the desired error.

# Changes made:

- Modified the message displayed when there is a version mismatch between the Wazuh API and the Wazuh App, to make it easier to understand.
- Removed the \<a\> tags on the message. 
- Added `PLUGIN_APP_NAME` constant to make the message match the same structure as in Wazuh Dashboard, for easier modifications in the future. 
- Added `addTagsToUrl` function on `health-check-container.tsx` and modified the displayed error message to use that function. 

## `addTagsToUrl`
This function receives a string and returns it with the next modifications:

- If the string contains a structure `<something> guide: <URL>`, it will modify that part to return `<something> guide` wrapped in \<a\> tags with the href pointing to the URL. 
- If the string contains a URL without `guide:` before it, it will wrap the URL in \<a\> tags.
- If the string doesn't contain any URL, it will return it without modifications.

## Before the changes: 
image

## After the changes:
image

# Tests
**Scenario**: have an environment with different versions of Wazuh API and Wazuh app, in which the major and minor are not the same.
**When** the user opens the Wazuh plugin
**Then** the health-check confirmation should show an error on `Check Wazuh API Version`
**And** the log message should contain a URL to the Elasticsearch troubleshooting section of Wazuh Documentation.
**And** the log message should not contain any \<a\> tags or HTML structure
**And** the message shown at the bottom should contain a clickable link to the Elasticsearch troubleshooting section of Wazuh Documentation.

**Scenario**: have an environment with different versions of Wazuh API and Wazuh app, in which the major and minor are not the same.
**When** the user forces a health check in debug mode
**Then** the health-check confirmation should show an error on `Check Wazuh API Version`
**And** the log message should contain a URL to the Elasticsearch troubleshooting section of Wazuh Documentation.
**And** the log message should not contain any \<a\> tags or HTML structure
**And** the message shown at the bottom should contain a clickable link to the Elasticsearch troubleshooting section of Wazuh Documentation.

### DIFF
--- a/public/components/health-check/services/check-setup.service.ts
+++ b/public/components/health-check/services/check-setup.service.ts
@@ -14,7 +14,7 @@
 
 import { AppState, GenericRequest, WzRequest } from '../../../react-services';
 import { CheckLogger } from '../types/check_logger';
-import { PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_TROUBLESHOOTING } from '../../../../common/constants';
+import { PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_TROUBLESHOOTING, PLUGIN_APP_NAME } from '../../../../common/constants';
 import { webDocumentationLink } from '../../../../common/services/web_documentation';
 
 export const checkSetupService = appInfo => async (checkLogger: CheckLogger) => {
@@ -44,7 +44,7 @@ export const checkSetupService = appInfo => async (checkLogger: CheckLogger) => 
         api.groups.version !== appSplit[0] ||
         api.groups.minor !== appSplit[1]
       ) {
-        checkLogger.error(`Wazuh API and Wazuh App version mismatch. API version: ${apiVersion}. App version: ${setupData.data.data['app-version']}. Check more info about this error on ${webDocumentationLink(PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_TROUBLESHOOTING)}.`);
+        checkLogger.error(`Wazuh API and ${PLUGIN_APP_NAME} version mismatch. API version: ${apiVersion}. App version: ${setupData.data.data['app-version']}. Check more info about this error on our troubleshooting guide: ${webDocumentationLink(PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_TROUBLESHOOTING)}.`);
       }
     }
   }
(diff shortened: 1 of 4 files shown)


Tostti:
# Changes done

- Modified the message to be clearer and removed the <a> tags on it. 
- Modified the URL shown in the message to the troubleshooting URL. 
- Added a function on `health-check.container.tsx` to add the corresponding tags to URLs in error messages, thus there are shown with a working link. 

Before the changes:
image

After the changes:
image

public/components/health-check/services/check-setup.service.ts
```diff
@@ -44,7 +44,7 @@ export const checkSetupService = appInfo => async (checkLogger: CheckLogger) =>
         api.groups.version !== appSplit[0] ||
         api.groups.minor !== appSplit[1]
       ) {
-        checkLogger.error(`Wazuh API and Wazuh App version mismatch. API version: ${apiVersion}. App version: ${setupData.data.data['app-version']}. At least, major and minor should match. Check more info about upgrading Wazuh App <a target='_blank' href='${webDocumentationLink(PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_UPGRADE_PLATFORM)}'>here</a>.`);
+        checkLogger.error(`Wazuh API and ${PLUGIN_APP_NAME} version mismatch. API version: ${apiVersion}. App version: ${setupData.data.data['app-version']}. Check more info about this error on our troubleshooting guide: ${webDocumentationLink(PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_TROUBLESHOOTING)}.`);
```
AlexRuiz7, line 47: I think that: `Read more about this error in our troubleshooting guide:` sounds better

Tostti: Talked in daily, modified in 6eebd86221204bfe87af16d7ebaa8beae2b403ba

Mayons95 (approved):
LGTM

PR #4531, opened 2022-09-15
Fix/configuration read

Mayons95:
The goal of this PR is to fix the read configuration test for Wazuh Dashboard

Screenshot from 2022-09-15 12-22-06


Mayons95:
Close because @mauceballos was working on a fix for the same test https://github.com/wazuh/wazuh-kibana-app/issues/4530

PR #4544, opened 2022-09-16
Change message for API version mismatch in WzD

Tostti:
Solves #4299 

# Important information
On PR https://github.com/wazuh/wazuh-kibana-app/pull/4103, a method was changed to request the API endpoint. Only for testing purposes of this particular issue, it is possible to replace line 50 of `server/lib/api-interceptor.ts` with the following: `method: !!authContext ? 'POST' : 'GET',` to avoid getting an error due to a not allowed method and trigger the desired error.

# Changes made:

- Modified the message displayed with a version mismatch between Wazuh API and Wazuh App occurs, to make it more clear. 
- Removed the \<a\> tags on the message. 
- Added `PLUGIN_APP_NAME` constant to make the message match the same structure as in Wazuh Dashboard, for easier modifications in the future. 
- Added `addTagsToUrl` function on `health-check-container.tsx` and modified the displayed error message to use that function. 

## `addTagsToUrl`
This function receives a string and returns it with the next modifications:

- If the string contains a structure `<something> guide: <URL>`, it will modify that part to return `<something> guide` wrapped in \<a\> tags with the href pointing to the URL. 
- If the string contains a URL without `guide:` before it, it will wrap the URL in \<a\> tags.
- If the string doesn't contain any URL, it will return it without modifications.

## Before the changes: 
image

## After the changes:
image

# Tests
**Scenario**: have an environment with different versions of Wazuh API and Wazuh Dashboard, in which the major and minor are not the same.
**When** the user opens the Wazuh section
**Then** the health-check confirmation should show an error on `Check Wazuh API Version`
**And** the log message should contain a URL to the Wazuh dashboard troubleshooting section of Wazuh Documentation.
**And** the log message should not contain any \<a\> tags or HTML structure
**And** the message shown at the bottom should contain a clickable link to the Wazuh dashboard troubleshooting section of Wazuh Documentation.

**Scenario**: have an environment with different versions of Wazuh API and Wazuh Dashboard, in which the major and minor are not the same.
**When** the user forces a health check in debug mode
**Then** the health-check confirmation should show an error on `Check Wazuh API Version`
**And** the log message should contain a URL to the Wazuh Dashboard troubleshooting section of Wazuh Documentation.
**And** the log message should not contain any \<a\> tags or HTML structure
**And** the message shown at the bottom should contain a clickable link to the Wazuh Dashboard troubleshooting section of Wazuh Documentation.


public/components/health-check/services/check-setup.service.ts
```diff
@@ -44,7 +44,7 @@ export const checkSetupService = appInfo => async (checkLogger: CheckLogger) =>
         api.groups.version !== appSplit[0] ||
         api.groups.minor !== appSplit[1]
       ) {
-        checkLogger.error(`Wazuh API and ${PLUGIN_APP_NAME} version mismatch. API version: ${apiVersion}. App version: ${setupData.data.data['app-version']}. At least, major and minor should match. Check more info about upgrading ${PLUGIN_APP_NAME} <a target='_blank' href='${webDocumentationLink(PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_UPGRADE_PLATFORM)}'>here</a>.`);
+        checkLogger.error(`Wazuh API and ${PLUGIN_APP_NAME} version mismatch. API version: ${apiVersion}. App version: ${setupData.data.data['app-version']}. Check more info about this error on our troubleshooting guide: ${webDocumentationLink(PLUGIN_PLATFORM_WAZUH_DOCUMENTATION_URL_PATH_TROUBLESHOOTING)}.`);
```
AlexRuiz7, line 47: I think that: `Read more about this error in our troubleshooting guide:` sounds better

Mayons95 (approved):
TR: :heavy_check_mark: 
LGTM

<details>

Screenshot from 2022-09-26 10-06-52

Screenshot from 2022-09-26 10-07-02

</details>

AlexRuiz7 (approved):

PR #4729, opened 2022-10-20
Adding sample data call on test definition

Mayons95:
Hello team.

This PR includes a fixed for a problem that was related to generate report test. The problem was related to cases when sample data was not added, so after the work of @mauceballos we can made a function to call and add sample data with API calls.
So I added the call for the function that @mauceballos implemented and with this the problem was solved.

<details>

image

The error displayed it's related to a reported bug: https://github.com/wazuh/wazuh-kibana-app/issues/4621
</details>


AlexRuiz7 (approved):
LGTM

PR #4758, opened 2022-10-24
Add centralized request service

Tostti:
### Description
This PR creates a centralized request service for the plugin
 
### Issues Resolved
https://github.com/wazuh/wazuh-kibana-app/issues/4757

### Service description
The created service has an async function named request, that can be called from different places in the code. It also contains a function named initializeInterceptor, which must be called at the plugin's start in order to define the interceptor.

The request function receives an object with the next parameters:

Method: GET, PUT, POST, DELETE, PATCH
Headers: an object with the desired headers. Plugin platform and content type headers are not needed.
Path: the path to which the request should be made. It can include queries, and should not include the basePath.
Data: the payload of the request.
Timeout: desired timeout in ms. If it's missing or the value is 0, no timeout will be set.
Once executed, it returns a promise. If there are no errors, it will be accepted and contain the result of the request. If there are errors, it will be rejected and contain the error.

This function makes the request with the core.http.fetch service. If there is a 401 unauthorized response from the server, due to the session expired, it will prevent doing any other request, abort the current ones and reload the page to trigger the login process. The prevention and abortion of new requests can also be done externally with the `disableRequests` function.

This PR also adds a verification on the server definition to prevent the component from making requests if the user is not authenticated. This is because on the reload/login process sometimes the plugin was initialized, a request was done and that caused the login process to be broken. There is detailed information about that on https://github.com/wazuh/wazuh-kibana-app/issues/4757

### Check List
- [x] All tests pass
  - [x] `yarn test:jest`

## Manual testing
To test the integration of the service, an environment of OSD with SAML is required. It is recommended to add the following line to `opensearch_dashboards.yml`: `opensearch_security.session.ttl: 10000`. That will make the session to be invalid after 10 seconds of inactivity (default is 1h) 

Scenario: Log in with SAML
Given the idle time after logging in is less than the session validity (10s in the above example)
When the Wazuh app is opened
Then it should open without crashing

Scenario: Log in with SAML
Given the idle time after logging in is more than the session validity (10s in the above example)
When the Wazuh app is opened
Then the system should revalidate the SAML session and redirect to the home page (not Wazuh)

Scenario: Log in with SAML and immediately after that open the Wazuh App
Given the idle time after opening the app is less than the session validity (10s in the above example)
When an interaction happens
Then it should work normally, with the expected behavior of that interaction.

Scenario: Log in with SAML and immediately after that open the Wazuh App
Given the idle time after opening the app is more than the session validity (10s in the above example)
When an interaction happens
Then the system should revalidate the SAML session and redirect to the home page (not Wazuh)


AlexRuiz7 (dismissed):
LGTM!

AlexRuiz7 (approved):

Mayons95:
With @Tostti we found an error with the expire session on Open search Logout. The logout service return a 401 error when te session expires.

https://user-images.githubusercontent.com/104914131/200359123-884b4d35-67f9-4d99-b562-ebccd2b02953.MOV

PR #4830, opened 2022-11-08
Fix styles in deploy new agent section

chantal-kelm:
### Description

In Wazuh info, in the 'Deploy new agent' section, the OS and versions were not displayed correctly.
 
### Evidence

This is what Wazuh Info used to look like:

image (10)

This is what it looks like now: 

image

### Steps to test:
*Go to the Agents tab
*Click on the 'Deploy new agent' button.
*Verify that the styles are displayed correctly.

### DIFF
--- a/public/styles/component.scss
+++ b/public/styles/component.scss
@@ -120,20 +120,25 @@ kbn-dis doc-table .kbnDocViewer__warning {
     flex-grow: 0;
 }
 
-.osButtonsStyle {
-    display: grid;
-    grid-template-columns: repeat(5, 1fr);
-    grid-gap: 10px;
-}
-
-.osButtonsStyleMac {
-    display: grid;
-    grid-template-columns: repeat(3, 1fr);
-    grid-gap: 10px;
-}
-
 .message {
     margin-top: 10px;
     display: flex;
     flex-direction: row;
+}
+
+.euiButtonGroup__buttons {
+    display: grid;
+    grid-template-columns: repeat(5, 1fr);
+    grid-gap: 10px;
+    padding: 0 3px;
+}
+
+.flex {
+    display: flex;
+}
+
+.euiButtonGroup--small {
+    &.euiButtonGroupButton:not(:first-child) {
+    margin-left: 0;
+    }
 }
--- a/public/styles/component.scss
+++ b/public/styles/component.scss
@@ -131,14 +131,19 @@ kbn-dis doc-table .kbnDocViewer__warning {
     grid-template-columns: repeat(5, 1fr);
     grid-gap: 10px;
     padding: 0 3px;
+    box-shadow: none;
 }
 
 .flex {
     display: flex;
 }
 
-.euiButtonGroup--small {
-    &.euiButtonGroupButton:not(:first-child) {
-    margin-left: 0;
-    }
+
+.euiButtonGroup--medium .euiButtonGroupButton:not(:first-child), .euiButtonGroup--small .euiButtonGroupButton:not(:first-child) {
+    margin-left: 0 !important;
+
 }
+
+// .euiButtonGroupButton:not(:first-child) {
+//     margin-left: 0;
+// }
--- a/public/styles/component.scss
+++ b/public/styles/component.scss
@@ -138,12 +138,7 @@ kbn-dis doc-table .kbnDocViewer__warning {
     display: flex;
 }
 
-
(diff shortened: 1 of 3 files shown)


Mayons95 (commented):
image

We saw witch @chantal-kelm that it's better to disable box-shadow option to have a clear UI.
Also the option for Suse 12 has the same problem as before.

yenienserrano (approved):
LGTM

image

Mayons95 (approved):
TR: :heavy_check_mark: 

image

public/styles/component.scss
```diff
@@ -120,20 +120,25 @@ kbn-dis doc-table .kbnDocViewer__warning {
...
 
-.osButtonsStyleMac {
+.euiButtonGroup__buttons {
     display: grid;
-    grid-template-columns: repeat(3, 1fr);
+    grid-template-columns: repeat(5, 1fr);
     grid-gap: 10px;
+    padding: 0 3px;
+    box-shadow: none;
 }
```
Desvelao, line 135: **question**: is this a class of the @elastic/eui? If it does, we should ensure only to affect the UI of the Wazuh plugin. This could be modifying the styles of plugins of the platform.

public/styles/component.scss
```diff
@@ -120,20 +120,25 @@ kbn-dis doc-table .kbnDocViewer__warning {
...
-    margin-top: 10px;
+.flex {
     display: flex;
-    flex-direction: row;
+}
+
+.euiButtonGroup--medium .euiButtonGroupButton:not(:first-child), .euiButtonGroup--small .euiButtonGroupButton:not(:first-child) {
+    margin-left: 0 !important;
+
 }
```
Desvelao, line 144: **question**: is this a class of the @elastic/eui? If it does, we should ensure only to affect the UI of the Wazuh plugin. This could be modifying the styles of plugins of the platform.

public/styles/component.scss
```diff
@@ -120,20 +120,25 @@ kbn-dis doc-table .kbnDocViewer__warning {
     flex-grow: 0;
 }
 
-.osButtonsStyle {
-    display: grid;
-    grid-template-columns: repeat(5, 1fr);
-    grid-gap: 10px;
+.message {
```
Desvelao, line 123: **nitpick**: this class name is generic and could have conflicts with the styles of other plugins, we should use scoped names, for example prefixing it with `wz`, `wazuh`.

public/styles/component.scss
Desvelao: **thought**: For another hand, this pull request has as its base the branch `4.4-2.3-wzd`, which is a plugin compatible with Wazuh dashboard based on OpenSearch Dashboards 2.3.0. This version of OpenSearch is using a forked UI library, https://github.com/opensearch-project/OpenSearch-Dashboards/blob/2.3.0/package.json#L111, and it seems the class names are prefixed by `oui` instead of `eui` https://github.com/opensearch-project/oui/blob/1.0.0/src/components/button/button_group/_button_group_button.scss#L122. Is these styles applying?

Desvelao: > **thought**: For another hand, this pull request has as its base the branch `4.4-2.3-wzd`, which is a plugin compatible with Wazuh dashboard based on OpenSearch Dashboards 2.3.0. This version of OpenSearch is using a forked UI library, https://github.com/opensearch-project/OpenSearch-Dashboards/blob/2.3.0/package.json#L111, and it seems the class names are prefixed by `oui` instead of `eui` https://github.com/opensearch-project/oui/blob/1.0.0/src/components/button/button_group/_button_group_button.scss#L122. Is these styles applying?

I was testing the branch and it is working. It seems that the UI library contains components from @elastic/eui and oui components.

Anyways, we should ensure that the modified styles only happen in the Wazuh plugin.

public/styles/component.scss
chantal-kelm: Fixed in this [PR](https://github.com/wazuh/wazuh-kibana-app/pull/4837)

public/styles/component.scss
chantal-kelm: Fixed in this [PR](https://github.com/wazuh/wazuh-kibana-app/pull/4837)

public/styles/component.scss
chantal-kelm: Fixed in this [PR](https://github.com/wazuh/wazuh-kibana-app/pull/4837)

PR #4831, opened 2022-11-08
Add centralized request service

Tostti:
### Description
This PR creates a centralized request service for the plugin

 
### Issues Resolved
https://github.com/wazuh/wazuh-kibana-app/issues/4757

### Service description
The created service has validators that disable the possibility to make new requests if a `401` was received due to an expired session. It contains an async function named request, that can be called from different places in the code, and also a function named initializeInterceptor, which must be called at the plugin's start in order to define the interceptor.

The request function validates if it is possible to make requests, and in that case make the requests with `axios`. It receives an object with all the options that can be sent to axios (method, headers, url, etc). Once executed, it returns a promise. If there are no errors, it will be accepted and contain the result of the request. If there are errors, or the requests are disabled due to a previously received 401, it will be rejected and contain the a message.

This PR also modifies the lifecycle of the initialization of the plugin: 

- Previously, after mounting the component a verification was made to check if the user has Wazuh disabled, and in that case the component was unmonuted. That caused issues when SAML was enabled, the session has expired and the user tried to open the app, resulting in a request interfering with the login process and breaking it. 
- With the new changes, that validation is done at the setup of the component, and in case that the user has Wazuh disabled, the component is not even registered.

# Manual testing
### XPACK
Scenario: Log in to an environment with Xpack
When the Wazuh app is opened
Then it should open without crashing

Scenario: Log in to an environment with Xpack
When the user navigates in the Wazuh APP
It should not crash

Scenario: Log in to an environment with Xpack, and navigate to the Wazuh API Console and manually delete the `sid` cookie to emulate a session expired
When the user makes a request to `GET /`
Then it should be redirected to the login page

Scenario: Log in to an environment with Xpack with a user without Wazuh enabled
When the sidebar is opened
Then the Wazuh app should not be shown

### Without XPACK
Before these tests, is recommended to modify the `kibana.yml` file and add the following line: 
`opendistro_security.session.ttl: 15000`
That will make the session to be closed after 15 seconds of inactivity automatically

Scenario: Log in to an environment and open the Wazuh APP before the session expires
When the Wazuh app is opened
Then it should open without crashing

Scenario: Log in to an environment, open the Wazuh APP before the session expires, and navigate inside the app
When the user navigates in the Wazuh APP
It should not crash

Scenario: Log in to an environment, open the Wazuh APP after the session expires
When the Wazuh app is opened
It should redirect to the login

Scenario: Log in to an environment, open the Wazuh APP before the session expires, wait until the session expires and navigate inside the app
When the user navigates in the Wazuh APP
It should redirect to the login page

Scenario: Log in to an environment with a user without Wazuh enabled
When the sidebar is opened
Then the Wazuh app should not be shown

## Backport 7.16 

Same as main PR

## Backport 2.3-wzd
Before these tests, is recommended to modify the `opensearch_dashboards.yml` file and add the following line: 
`opensearch_security.session.ttl: 15000`
That will make the session to be closed after 15 seconds of inactivity automatically

### Without SAML
Scenario: Log in to an environment and open the Wazuh APP before the session expires
When the Wazuh app is opened
Then it should open without crashing

Scenario: Log in to an environment, open the Wazuh APP before the session expires, and navigate inside the app
When the user navigates in the Wazuh APP
It should not crash

Scenario: Log in to an environment, open the Wazuh APP after the session expires
When the Wazuh app is opened
It should redirect to the login

Scenario: Log in to an environment, open the Wazuh APP before the session expires, wait until the session expires, and navigate inside the app
When the user navigates in the Wazuh APP
It should redirect to the login page

Scenario: Log in to an environment with a user without Wazuh enabled
When the sidebar is opened
Then the Wazuh app should not be shown

### With SAML

Scenario: Log in to an environment with SAML and open the Wazuh APP before the session expires
When the Wazuh app is opened
Then it should open without crashing

Scenario: Log in to an environment with SAML, open the Wazuh APP before the session expires, and navigate inside the app
When the user navigates in the Wazuh APP
It should not crash

Scenario: Log in to an environment with SAML, open the Wazuh APP after the session expires
When the Wazuh app is opened
It should redirect to the IDP login

Scenario: Log in to an environment with SAML, open the Wazuh APP before the session expires, wait until the session expires, and navigate inside the app
When the user navigates in the Wazuh APP
It should redirect to the IDP login

Scenario: Log in to an environment with SAML with a user without Wazuh enabled
When the sidebar is opened
Then the Wazuh app should not be shown

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`


server/plugin.ts
```diff
@@ -73,6 +74,9 @@ export class WazuhPlugin implements Plugin<WazuhPluginSetup, WazuhPluginStart> {
     let wazuhSecurity;
     core.http.registerRouteHandlerContext('wazuh', async(context, request) => {
       !wazuhSecurity && (wazuhSecurity = await SecurityObj(plugins, context));
+      if(!request.auth.isAuthenticated){
+        disableRequests();
+      }
```
Desvelao, line 79: **thought**: I'm not sure if this does what you expect. This code is executed on the backend, meanwhile, the request service is executed in the frontend, I don't understand how from the backend you want to communicate to the frontend to disable the requests done from there.

common/services/request-handler.js
```diff
@@ -0,0 +1,60 @@
...
+    if (!options.method | !options.url) {
+        return Promise.reject("Missing parameters")
+    }
+    const requestId = currentid;
+    currentid++;
+    const abort = new AbortController();
+    if (allow) {
+        try {
+            aborts.push({ id: requestId, controller: abort });
+            const requestData = await axios(options);
```
Desvelao, line 52: **nitpick**: the request requires a method to be canceled. This documentation link could help: https://axios-http.com/docs/cancellation

common/services/request-handler.js
```diff
@@ -0,0 +1,60 @@
...
+            removeController(requestId);
+            return Promise.resolve(requestData);
+        }
+        catch (e) {
+            return Promise.reject(e);
```
Desvelao, line 57: **nitpick**: the requests done via axios are not managing when there is a request failing due to the session expired.

The security plugin, implement a request interceptor in the `core.http` service that reloads the page when a request with status code 401 is detected.

Tostti: The service is intended to manage the requests and prevent doing more of them with the interceptor configured before. 

The problem using `core.http` is that it will reload the page with every 401 code received. That include some codes that we receive from the Wazuh API, leading to a healthcheck loop. 

This approach intends to allow keep making requests if the 401 is received from the Wazuh API, but disable them if there's a 401 received on Kibana/Opensearch Dashbords (i.e. due to session expired), because those uses `core.http`

common/services/request-handler.js
Tostti: Issue fixed

public/services/request-handler.js
```diff
@@ -0,0 +1,52 @@
...
+        },
+    }
+    if (allow) {
+        try {
+            const requestData = await axios(options);
+            if (requestData.status === HTTP_STATUS_CODES.UNAUTHORIZED) {
+                if (requestData.data.message === 'Unauthorized' || requestData.data.message === 'Authentication required') {
+                    disableRequests();
+                }
+                throw new Error(requestData.data.message)
```
Desvelao, line 44: **nitpick**: This modification of the error is breaking the functionality to refresh the page when the session expires.

The valid requests through axios and the returned interface should match with other services or adapt them to the new management.

For the expired session case, the WzRequest and GenericRequest services manage the response and reload the page if the error matches with some conditions. These conditions are being modified with these changes. We should ensure the browser tab is reloaded when the ODFE session expired.

References:
- WzRequest error management related to the ODFE session: https://github.com/wazuh/wazuh-kibana-app/blob/v4.3.9-7.10.2/public/react-services/wz-request.ts#L67
- GenericRequest error management related to the ODFE session: https://github.com/wazuh/wazuh-kibana-app/blob/v4.3.9-7.10.2/public/react-services/generic-request.js#L93
- Reload the browser tab depending on the error message from the request: https://github.com/wazuh/wazuh-kibana-app/blob/v4.3.9-7.10.2/public/utils/odfe-utils.ts#L27-L32

Tostti: Possibly fixed in https://github.com/wazuh/wazuh-kibana-app/pull/4831/commits/030c1e4be75bf725416108c34ff330ed223e9c3e

Desvelao:
# Update
- Fix a condition problem
- Replace a hardcoded http status code value by constant
- Remove unused import

Desvelao (approved):
**review**:

:heavy_check_mark: code review
:heavy_check_mark: tests for Kibana 7.10.2
- :heavy_check_mark: X-Pack
- :heavy_check_mark: ODFE

yenienserrano (approved):
CR: ✔️ 
tests for Kibana 7.10.2: :heavy_check_mark:

PR #4909, opened 2022-11-18
Fix IPV6 visualizations

Tostti:
### Description
As IPv6 addresses are longer than IPv4, the places where the IPs are displayed may be too small, and cause visualization issues like overlapping of the address with text from another place.
This PR fixes that issue by: 
- Compressing the addresses following the [IPv6 compression rules](https://www.ciscopress.com/articles/article.asp?p=2803866).  
- Modifying the way in which the addresses are displayed, regardless if there can be compressed or not.

### Issues Resolved
https://github.com/wazuh/wazuh-kibana-app/issues/4566

### Evidence
**Before the changes:**
image
image

**After the changes:**
image
image

### New functions
`compressIPv6()` receives a string. If it is an uncompressed IPv6, it returns the compressed version in a string. If it is not an uncompressed IPv6, or it is not a string, it returns the same that it received. 

### Test
It is recommended to modify the Imposter's `agent.json` and `agents.json` to do these tests, and change the IP addresses for different IPv6 examples.

Scenario: have a registered agent with an incompressible IPV6 (i.e.: FE80:1244:2223:A000:2202:B3FF:FE1E:8329)
When the user navigates to the Agents tab
Then the IP should be displayed without visual errors.

When the user navigates inside the agent from the Agents tab
Then the IP should be displayed without visual errors.

Scenario: have a registered agent with a compressible IPV6 (i.e.: FE80:0000:0000:0000:2202:00FF:0E1E:8329)
When the user navigates to the Agents tab
Then the IP should be displayed without visual errors and correctly compressed.

When the user navigates inside the agent from the Agents tab
Then the IP should be displayed without visual errors and correctly compressed.

Scenario: have a registered agent with an IPV4  address 
When the user navigates to the Agents tab
Then the IP should be displayed without visual errors.

When the user navigates inside the agent from the Agents tab
Then the IP should be displayed without visual errors.

When the user navigates inside the agent from the Agents tab
Then the IP should be displayed without visual errors.
### Check List
- [x] All tests pass
  - [x] `yarn test:jest`


public/services/ipv6-services.js
```diff
@@ -0,0 +1,21 @@
+export const compressIPv6 = (ip) => {
+  if (typeof (ip) !== 'string') {
+    return ip;
+  }
+  if (ip?.split(':').length !== 8) {
+    return ip;
+  }
+
+  let output = ip.split(':').map(terms => terms.replace(/\b0+/g, '') || '0').join(":");
+  let zeros = [...output.matchAll(/\b:?(?:0+:?){2,}/g)];
```
AlexRuiz7, line 10: Probably can be switched to `const`.

Also, would be great to migrate this service to TypeScript

AlexRuiz7 (approved):
Check the suggestions

public/services/ipv6-services.js
Tostti: Changes implemented on c0f53f5b2a93db93599009d321f53c5d1b244699

chantal-kelm (approved):
✅ ✅

PR #4964, opened 2022-12-06
Add mismatch error message when method not allowed

Tostti:
### Description
This PR adds a message when the API returns a Method Not Allowed to let the user know that the error may be related to a version mismatch between the App and the Server.
 
### Issues Resolved
https://github.com/wazuh/wazuh-kibana-app/issues/4915

### Evidence
image

### Test
**Scenario:** have an environment with Wazuh server <=4.3 and Wazuh App >=4.4
**When** the user navigates to Wazuh and the healthcheck is done
**Then** an error should appear in the first check, and a message telling that may be related to a version mismatch should appear at the bottom.

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`


public/components/health-check/services/check-api.service.ts
```diff
@@ -33,11 +33,24 @@ const trySetDefault = async (checkLogger: CheckLogger) => {
...
+            `Could not connect to API id [${hosts[i].id}]: ${
+              err.message || err
+            }`,
+          );
         }
       }
       if (errors.length) {
+        if (errors[0].includes('ERROR3099 - 405: Method Not Allowed')) {
+          return Promise.reject(
+            'No API available to connect. This may be related to a version mismatch between Wazuh Server and Wazuh Kibana plugin. Please check the versions and try again.',
```
Desvelao, line 51: **nitpick**: We should avoid using hardcoded values for the platform name.

public/components/health-check/services/check-api.service.ts
```diff
@@ -33,11 +33,24 @@ const trySetDefault = async (checkLogger: CheckLogger) => {
...
+          );
+          errors.push(
+            `Could not connect to API id [${hosts[i].id}]: ${
+              err.message || err
+            }`,
+          );
         }
       }
       if (errors.length) {
+        if (errors[0].includes('ERROR3099 - 405: Method Not Allowed')) {
```
Desvelao, line 49: **nitpick**: this is only checking the first element of the array, In the case, the user has more Wazuh API host entries set, if the API with the mismatched version related to the change of HTTP method to authenticate is not placed the first one, it could not be caught.

I think the error should be managed for each Wazuh API host. We could add more information related to the error instead of the `Could not connect to API id [host ID]` message.

Desvelao, line 49: **nitpick**: this is only checking the first element of the array, In the case, the user has more Wazuh API host entries set, if the API with the mismatched version related to the change of HTTP method to authenticate is not placed the first one, it could not be caught.

I think the error should be managed for each Wazuh API host. We could add more information related to the error instead of the `Could not connect to API id [host ID]` message.

Desvelao (changes requested):
**review**:

Solve the commented problems.

Changelog is missing.

public/components/health-check/services/check-api.service.ts
Tostti: Resolved in https://github.com/wazuh/wazuh-kibana-app/pull/4964/commits/dc4300c581cdccb125bbed6b0922dca7a1a605ee

public/components/health-check/services/check-api.service.ts
Tostti: Resolved in https://github.com/wazuh/wazuh-kibana-app/pull/4964/commits/dc4300c581cdccb125bbed6b0922dca7a1a605ee

Machi3mfl (dismissed):
CR: ✔️ 
Test: ✔️

Desvelao (approved):
**review**

Code :green_circle: 

Test

| Scenario | Result |
| --- | --- |
| 1 | :green_circle: |

<details>
  <summary>Scenario 1 :green_circle: </summary>
  
An API in 4.3.x  
image

An API down and an API up 4.3.x
image

<details>

asteriscos (approved):
CR: :heavy_check_mark: 
Test: :heavy_check_mark: 

image

Issue #5052, opened 2022-12-29
Implement new versioning methodology


AlexRuiz7:
## Description

We currently use 3 properties in the `package.json` file to handle the versioning of the app. These are:

- `version`: the Wazuh version. Follows the `major.minor.patch` pattern.
- `revision`: used to be the version without dots. Two digits were appended to specify the build number. This was only changed when the version had changed. Example: 4400.
- `code`: unused at the moment. Usually set to the same value as the revision.

https://github.com/wazuh/wazuh-kibana-app/blob/130cd255ba7a0f37ffaa4e5c11efefca46edea76/package.json#L3-L5

Instead, we would like to simplify and standardize the versioning method we use, to follow this approach:

- `version`: untouched. Follows the `major.minor.patch` pattern
- `revision`: represents the build number. It is a number of 2 digits, starting at 00 and increased by one **each time** a package for the same version is built. This value is reset **only** when the version changes.
- `stage`: replaces the code property which has no actual use. With this property, we'll represent the [stage of development](https://en.wikipedia.org/wiki/Software_release_life_cycle#Stages_of_development) at which the package has been built at: `pre-alpha` (development), `alpha`, `beta`, `release-candidate` or `stable` (production).
- `commit`: we'll also include the short SHA of the latest commit included in the package (can also be the SHA of the commit used to tag). With this, we'll be able to have a clear reference of what was included in the package.

![Software_dev2 svg (1)](https://user-images.githubusercontent.com/15186973/209981087-04191391-8cc7-4e1b-9bb9-f5f4f2fcae3c.png)

## Tasks

- [x] Check the `code` property is actually unused.
- [x] Replace the `code` property with `stage`.
- [x] Add the `commit` property.
- [x] Add validation. Validate that `version`, `revision` and `stage` contain expected values.
- [ ] Update the [build workflow](https://github.com/wazuh/wazuh-kibana-app/blob/v4.4.0-2.4.1-alpha2/.github/workflows/build.yml), **if required**.
- [ ] (Optional) Automate the process with a build script.

## Related issues 

- https://github.com/wazuh/wazuh-packages/issues/2012

AlexRuiz7:
I've set up 2 development environments (7.10 and 7.17) on which I have change `code` with `stage` and no errors were raised.

AlexRuiz7:
Useful resource: https://www.shellhacks.com/git-get-short-hash-sha-1-from-long-hash-head-log/

[...]

PR #5060, opened 2023-01-02
Implement new versioning methodology

yenienserrano:
### Description

Some of the `package.json` settings were changed, 
for example: 
- `version`: untouched. Follows the major.minor.patch pattern
- `revision`: represents the build number. It is a number of 2 digits, starting at 00 and increased by one each time a package for the same version is built. This value is reset only when the version changes.
- `stage`: replaces the code property which has no actual use. With this property, we'll represent the stage of development at which the package has been built at: pre-alpha (development), alpha, beta, release-candidate or stable (production).
- `commit`: we'll also include the short SHA of the latest commit included in the package (can also be the SHA of the commit used to tag). With this, we'll be able to have a clear reference of what was included in the package.

Tests were added to corroborate the correctness of these settings.

Also, a script to generate the tags automatically has been added.
 
### Issues Resolved

- #5052 

### Evidence
image
image

### Test

Check that the `package.json` contains the settings:
- `version`
- `revision`
- `stage`
- `commit`

Verify that the tests are working and if you change the value of these settings to invalid values, the tests will not work.

`package.test.ts`

### Check List
- [x] All tests pass
  - [x] `yarn test:jest`


chantal-kelm (dismissed):
CR: ✅ 
Test: ✅ 

Evidence:

image

image

chantal-kelm (approved):
CR: ✅ 
Test: ✅

Issue #4810, opened 2022-11-03
Modularization


gdiazlo:
## UI modules

We want to decompose our application into modules following the single responsibility design principle, among others. Before starting, we want to plan, design, and clarify the following topics:

- [x] https://github.com/wazuh/wazuh-kibana-app/issues/5113

- [x] https://github.com/wazuh/wazuh-kibana-app/issues/5114

- [x] https://github.com/wazuh/wazuh-kibana-app/issues/5228

 - [x] Decide between multi-plugin modules or single-plugin modules. 
     What's more convenient for us: developing different modules as separate plugins, or building different modules within the current one. Our preference would be to build them using our current one, but we need to ensure the different modules do not interact between them in unsuspected ways.

 - [x] Designs to communicate a module with others.
    - How a module can search for others
    - How a module can export services and views
    - How a module can prevent others from using certain services and views

 - [x] Define module lifecycle
   - login and logout 
   - wazuh API auth 
   - session renewal

 - [ ]  Define how to implement and use RBAC cross-module

 - [ ] Define a way to make them portable between our platforms
 
 - [ ] https://github.com/wazuh/wazuh-kibana-app/issues/5220
 
 - [ ] Define a way to build and install multiple plugins

We might revisit this list and add or remove items as the work progress.

Desvelao:
Multi-plugin module (one repository)
> Register multiple applications

Shared services:
- requests related to Wazuh API
- RBAC
- Wazuh API session
- Configuration
- Reporting
- Filters
- Visualizations
- agnostic UI components

Define the structure of modules:
  - views
  - user flow

asteriscos:
The following principles will be applied to improve scalability and maintainability:

- Encapsulate each app module files in the app module folder
- Apps modules should only consume services and components from the _/common_ folder or their own folder.
- Apps modules should not depend on other modules
- Services, components, stores, etc contained in a specific app module folder should be shared with other modules. If the need arises, it should be migrated to the _/common_ folder.

Modules as registered apps

![image](https://user-images.githubusercontent.com/9343732/203764136-ffe4f06d-3e44-415f-8668-cd8f73b6af7f.png)

gdiazlo:
I'll close this as our new roadmap has evolved. The research effort will continue on other issues when we see fit. For know, we have the knowledge we wanted.

PR #5229, opened 2023-02-24
Add custom formatter

### LINKED ISSUE #5228: Create POC custom field formatter
# Description

We need to create a custom field formatter proof of concept to extend the capabilities of a field in the Discover plugin. The goal is to have the same functionality we currently have in our custom implementation of the Discover plugin on the native Discover plugin.

image

asteriscos:
### Description
Hi team,
thi PR is an example of a custom field formatter. It has the basic feature of changing the styles of a field. The custom field formatter has to return a string, which can be in HTML format.
 
### Issues Resolved
Closes #5228 
Epic #4810 

image

## How to test it

The custom formatter will register automatically, so what we have to do is enable it in the index pattern configuration.

- Go to Stack Management -> Index patterns

image

- Select the desired index-pattern and look for the field which has to be formatted.

image

- Click on the Edit (_pencil_) button and select the custom field formatter. In this particular case _Wazuh custom format_

image

Finally, you should be able to see the formatted field in any of the table views available, such as the Discover plugin or a regular table.

image

image

### DIFF
--- a/public/apps/field-format-test.ts
+++ b/public/apps/field-format-test.ts
@@ -1,0 +1,21 @@
+import { FieldFormat } from '../../../../src/plugins/data/common/'
+import { HtmlContextTypeConvert } from '../../../../src/plugins/data/common/';
+import { OSD_FIELD_TYPES } from '../../../../src/plugins/data/common/';
+import { template } from 'lodash';
+
+const convertTemplate = template('<span style="<%- style %>"><%- val %></span>');
+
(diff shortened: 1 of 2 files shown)


asteriscos:
# Proposed solution

This code comparison to the base branch shows the POC implementation:
https://github.com/wazuh/wazuh-kibana-app/compare/modules/base-module-architecture-4810...modules/custom-field-formatter-5228

Issue #5415, opened 2023-05-02
Agent deploy page - optional steps


gdiazlo:
## Description
We need to clearly indicate the options in step 3 are optional. We don't want to hide them, but we want to clearly mark them as optional.

We want to change the step 3 title to:

_Optionally, assign a name and a group to the agent_

Also, we need to add the following phrase on top of the agent name input control:

_The agent name is set as the name of the host where it is deployed. You can set your own name in the field below. Note that the agent name cannot be changed once registered._

<img width="1486" alt="Captura de pantalla 2023-05-02 a las 17 10 59" src="https://user-images.githubusercontent.com/2181918/235708498-7a8b9027-309b-4077-9ed4-0e6775a159ff.png">

We don't want to hide the options inside the step, just to mark them as optional.

chantal-kelm:
It was decided to change the sentence: _The agent name is set as the name of the host where it is deployed. You can set your own name in the field below. Please note that the agent name cannot be changed once registered._

We are waiting for the new sentence

chantal-kelm:
We will change the sentence: "The agent name is set as the name of the host where it is deployed. You can set your own name in the field below. Note that the agent name cannot be changed once registered."

By: "The deployment sets the endpoint hostname as the agent name by default. Optionally, you can set the agent name below.".

And we will add this callout below the agent name entry:

<img width="481" alt="Captura de pantalla 2023-05-10 a la(s) 10 07 07" src="https://github.com/wazuh/wazuh-kibana-app/assets/99441266/57e7c2ba-a044-48e1-82c0-71a2b5924327">

PR #5515, opened 2023-05-30
fix macos command

chantal-kelm:
### Description

Fix macos command. This problem arose when merging changes

### Related issue

#5415


Desvelao:
Should we add the pull request entry to the changelog?

PR #5592, opened 2023-06-22
Remove unused embedded jquery-ui

Desvelao:
### Description
This pull request removes the unused embedded `jquery-ui`.
 
### Issues Resolved

Solve part of #5040

### Check List
- [x] All tests pass
  - [x] `yarn test:jest`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [x] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

PR #5674, opened 2023-07-12
[Backport 4.6.0-7.16] Remove unused embedded jquery-ui

Desvelao:
Backport 249eb8b5325234950ff91752d0ec404ca7ff8ebd from #5592.


plugins/main/public/app.js
```diff
@@ -105,23 +105,29 @@ app.run(['$rootElement', function ($rootElement) {
...
+              err,
+            );
+          })
+          .finally(() => {
+            window.location = event.currentTarget.href;
+          });
+      });
+    },
+  );
+});
```
asteriscos, line 133: Something is wrong at the end of the function, it's missing a bracket. Please check there wasn't an error in the merge.

asteriscos (changes requested):

plugins/main/public/app.js
Desvelao: Good catch. I did not see that the closing bracket was missing due to code formatting. Good catch. Fixed in 6e9e4b7cbddd81b316e3ff004262516653efdf82.

asteriscos (approved):
Test: :heavy_check_mark: 
CR: :heavy_check_mark: 

image

PR #6827, opened 2024-07-16
Fixed error when pinned agent in vulnerabilities

jbiset:
## Description
This PR solves filter error when pinning an agent when an index pattern is used for wazuh-alerts that is not `wazuh-alerts-*`.
 
### Before

Before

### After

After

## Test

1. It is necessary to remove both the index and the index pattern from wazuh-alerts-*.
2. To emulate the problem, the indexes generated with sample data will be used, so sample data must be added.
3. The default index pattern name in AppSettings must be changed. In this case I changed to wazuh-alerts-4.x-sample-security to take advantage of the sample data.
4. An index pattern named wazuh-alerts-4.x-sample-security must be created to match the wazuh-alerts-4.x-sample-security index of sample data.
5. Go to the vulnerabilities module, pin an agent in Dashboard/Inventory and then go to the Events tab.
6. Check that the pinned agent filter appears correctly without error.

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/plugins/main/public/components/overview/vulnerabilities/common/vulnerability_detector_adapters.tsx
+++ b/plugins/main/public/components/overview/vulnerabilities/common/vulnerability_detector_adapters.tsx
@@ -1,7 +1,6 @@
 import {
   AUTHORIZED_AGENTS,
   VULNERABILITY_IMPLICIT_CLUSTER_MODE_FILTER,
-  WAZUH_ALERTS_PATTERN,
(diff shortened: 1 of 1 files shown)


yenienserrano (dismissed):
CR: LGTM
TEST: LGTM

image

asteriscos (approved):
CR: :green_circle: 
Test: :green_circle: 

image

Issue #6866, opened 2024-07-23
Agent version shows red status when up to date


lucianogorza:
**Description**
The agent version status is incorrectly displayed as red (outdated) even when it is actually up to date. 

**Steps to reproduce**
1. Go to the Endpoints Summary.
2. Verify the agent version status.
3. Observe that the status is red despite the agent being up to date.

**Expected Result**
![image](https://github.com/user-attachments/assets/65e81e5d-cf24-480e-a9f3-8f6e72d20002)

**Actual Result**
The agent version shows a red status when it is actually up to date
![image](https://github.com/user-attachments/assets/4c598f85-a594-48b6-a806-3b1ede0cc302)

lucianogorza:
## Reproduce the bug

I can reproduce the bug with a CentOS agent.

![image](https://github.com/user-attachments/assets/1240f4d8-7273-4ff5-8f1f-6131bc6ee5ed)

lucianogorza:
Apparently it's an API issue, so it will be resolved at https://github.com/wazuh/wazuh/issues/24914

lucianogorza:
Test before update manager with changes from https://github.com/wazuh/wazuh/pull/24946
![image](https://github.com/user-attachments/assets/c4f0c504-3b31-45ac-bdd1-fe434bcd22ce)

After apply changes

asteriscos:
This problem was solved by a pull request made by @wazuh/devel-pyserver in Wazuh API. No work is needed on the Wazuh Dashboard side. I'm closing this as not planned.

PR #6891, opened 2024-07-31
Fix incorrect red status for up to date agent versions

lucianogorza:
### Description
This PR addresses the issue where the agent version status is incorrectly displayed as red (outdated) even when the agent is actually up to date. 
 
### Issues Resolved
https://github.com/wazuh/wazuh-dashboard-plugins/issues/6866

### Evidence
image

### Test

1. Add up to date agents, outdated agents and never connected agents.
2. Go to the Endpoints Summary.
3. Verify the agent version status.

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`


plugins/main/public/components/endpoints-summary/services/get-outdated-agents.ts
```diff
@@ -21,7 +21,7 @@ export const getOutdatedAgentsService = async ({
     params: {
       ...(agentIds?.length || q
         ? {
-            q: finalQ,
+            q: finalQ ? `(${finalQ})` : '',
```
Desvelao, line 24: **nitpick**: the usage of the parenthesis should not be necessary because the query should mean the same. Is this a bug on the Wazuh server API? If this is the case, then the problem should be solved on the server API instead of forcing to the Wazuh plugin to do a query that should return the expected results.

Desvelao:
# Test

I tested the proposed solution and this seems to work despite I understand that could have a bug on the server API.

image

## Comment
I created an environment with a Wazuh server 4.9.0 and deployed the following agents:
- `4.7.0`
- `4.9.0`

I got the following results testing the API endpoint to get the outdated agents:
- `GET /agents/outdated`: returns the expected agent 4.7.0
- `GET /agents/outdated?q=id=001,id=002`: returns both including an unexpected outdated agent 4.9.0 (it is up to date)
- `GET /agents/outdated?q=id=(001,id=002)`: returns the expected agent 4.7.0

asteriscos:
> [!WARNING]
> I tested the 4.9.0 branch without this fix using a Wazuh Manager 4.9.0 built from the source code (includes the fix made by @wazuh/devel-pyserver) and works properly. 

I will close this pull request because it's not needed.

## Evidence

image

PR #7484, opened 2025-05-29
Reduce the number of API calls to access agents summary information

### LINKED ISSUE #7426: Reduce the number of API calls to access agents' summary information
## Description

We are currently making three requests to the API to obtain the agent's summary information:

1. `/agents/summary/os`
2. `/agents/summary/status`
3. `/agents/stats/distinct?fields=group`

This is detrimental to the server since each of the requests requires authentication and permissions validation, which translates into more queries to wazuh-db and an overall worse performance.

In https://github.com/wazuh/wazuh/issues/29563, we are adding the endpoint `GET /agents/summary` to expose the OS, groups and status of all agents, or a set of them, in a single call.

We should replace the three requests outlined above with a single one to the new endpoint.

### Endpoint details

The endpoint takes a parameter `agents_list` for the user to select the summary of a specific subset of agents, if the parameter is unused, the summary of all agents is returned.

Regarding the endpoint response, it will contain three main fields `status`, `groups` and `os`.

```json
{
  "data": {
    "status": {
  	"active": 24,
  	"disconnected": 0,
  	"never_connected": 11,
  	"pending": 15
    },
    "os": {
  	"debian": 50
    },
    "groups": {
  	"default": 1
    }
  },
  "error": 0
}
```

The `status` object will always have a key/value for each one of the four statuses, regardless of whether the count is zero or non-zero.

On the other hand, `os` and `groups` objects are dynamic, and depend on the agent OSs and groups present in the system. The key contains the name of the OS/groups and the value the number of agents under that specific subset.

## Tasks
The steps that have to be completed in order to close the issue.

- [ ] Update the agent summary information retrieval to use a single API request.
- [ ] Validate that the agents dashboard widgets show the summary information successfully.

Tostti:
### Description

This PR implements a service to use the new `GET /agents/summary` endpoint, reducing the number of requests made to the API when entering the Agents Summary section. 

### Issues Resolved

#7426

### Evidence
image
image

### Check List

### DIFF
--- a/.github/workflows/4_bumper_repository.yml
+++ b/.github/workflows/4_bumper_repository.yml
@@ -133,7 +133,7 @@ jobs:
       - name: Merge pull request
         run: |
           # Any checks for the PR are bypassed since the branch is expected to be functional (i.e. the bump process does not introduce any bugs)
-          gh pr merge "${{ steps.create_pr.outputs.pull_request_url }}" --merge
+          gh pr merge "${{ steps.create_pr.outputs.pull_request_url }}" --merge --admin
 
       - name: Show logs
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,8 +13,7 @@ jobs:
     name: Build app package (auto)
     uses: ./.github/workflows/dev-environment.yml
     permissions:
-      actions: read
-      contents: read
+      pull-requests: write
     with:
       reference: ${{ github.ref_name }}
       command: 'yarn build'
--- a/.github/workflows/manual-build.yml
+++ b/.github/workflows/manual-build.yml
@@ -26,7 +26,8 @@ jobs:
   build:
     name: Build app package
     uses: ./.github/workflows/dev-environment.yml
-    permissions: {}
+    permissions:
+      pull-requests: write
     with:
       reference: ${{ inputs.reference }}
       command: 'yarn build'


gonzaarancibia:
🟢 LGTM
🟢 TEST:
Screenshot 2025-06-05 at 12 58 19 PM
Screenshot 2025-06-05 at 12 58 31 PM

gonzaarancibia (dismissed):

asteriscos (approved):
CR: :green_circle: 
Test: :green_circle: 

image

image

guidomodarelli:
CR: 🟢
Test: 🟢

image
image

guidomodarelli (dismissed):
LGTM! 🟢

plugins/main/public/components/endpoints-summary/services/get-agents-info.ts
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable */
```
Firefly0, line 1: I m an outsider - please feel free to dismiss my review 

I m here to practice me review skills and in the same time bring some value - if needed 

why disable lint for this part of the file? if is to overwrite some rule for imports order or import paths it is probably better to disable that specific rule - if there is something else that is not correct - eslint won't catch that either.
Most likely is not - but on the log run you may overlook that and miss something along the way

plugins/main/public/components/endpoints-summary/services/get-agents-info.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const statusData = AGENT_STATUS.map(({ label, status, color }) => ({
+      status,
+      label,
+      value: agentSummaryData.status?.[status] ?? 0,
+      color,
+    }));
+
+    return {
+      osData: osData.sort((a, b) => b.value - a.value).slice(0, 5),
```
Firefly0, line 97: This works just fine for this implementation - what if the requirement changes in the future for the limit?
Would not be easier to read and maintain if you put that in a constant and use it here
e.g.: 
```
// top 5 smth asked by stakeholders
const TOP_N = 5
slice(0, TOP_N)
```
this may be used on the following line as well
Easier to follow
Easier to test

plugins/main/public/components/endpoints-summary/services/get-agents-info.test.ts
```diff
@@ -0,0 +1,150 @@
+/* eslint-disable */
```
Firefly0, line 1: same as bellow

plugins/main/public/components/endpoints-summary/services/get-agents-info.ts
```diff
@@ -0,0 +1,116 @@
...
+      groupsData: groupsData.slice(0, 5),
+      statusData,
+    };
+  } catch (error) {
+    const options = {
+      context: 'EndpointsSummary.getSummary',
```
Firefly0, line 103: I see that this context is used in multiple places across the code-base 
If this are the requirements - this should be perfect
If not - maybe you can add more context, like the name of the function
e.g.: ` 'EndpointsSummary.getSummary getAgentsInfo func' ` - it will be easier to pinpoint exactly to the place where this took place

asteriscos (approved):

asteriscos:
@Firefly0 thank you for your interest in Wazuh source code. Your review is very welcome. Feel free to check other pull requests as well.

PR #7559, opened 2025-06-27
Merge 4.14.0 into main

Desvelao:
Merge 4.14.0 into main
> [!CAUTION]
> Merge PR strategy: Create a merge commit


Desvelao:
The eslint check will be ignored due to this was using the approach for the old 5.0.0. A new issue should be created to solve this problem.

The unit test check will be ignored due to it is a problem related with the elastic/eui library version used for the Docker image that is using 6.0.0 instead of the expected 5.0.0. This should be solved in another issue.

Issue #7794, opened 2025-10-02
Deprecate the use of `/agents/outdated` endpoint


asteriscos:
# Description

Due to performance issues, we need to avoid the use of the `/agents/outdated` of the Wazuh API endpoint, while keeping the current features.

## Tasks

- [ ] Replace or deprecate the use of the `/agents/outdated` endpoint from Wazuh dashboard

<img width="3952" height="1848" alt="Image" src="https://github.com/user-attachments/assets/fa732079-7047-4c40-802a-9ba7430aa29d" />

gonzaarancibia:
## Update

We’ve analyzed the current behavior and performance issues related to the `/agents/outdated` endpoint.

### Context

The dashboard currently:
- Calls `/agents/outdated` to retrieve outdated agents.
- Calls `/agents` to retrieve the complete list of agents.

The `/agents/outdated` endpoint is known to be heavy, as it executes complex database queries that can become slow or unstable when handling large environments with many agents or filters.
To improve performance and stability, we evaluated several alternatives to remove or simplify the logic behind outdated agents.

### Alternatives considered
**Option A – Deprecate `/agents/outdated` and handle outdated status in the frontend**

This approach determines outdated agents directly on the dashboard by comparing each **agent version vs. the manager version (from API info)**, instead of making an additional request to `/agents/outdated`.
Although this adds minimal logic to the frontend, it avoids the costly backend queries and simplifies request handling.

**Option B – Add an outdated flag in /agents API response**

The server could include an outdated boolean field per agent.
This would centralize the logic on the API side, but it requires backend changes that are currently out of scope.

**Option C – Keep /agents/outdated, but optimize it**

This option would involve optimizing the existing endpoint by adding indexes, caching, and pagination, but it still introduces significant complexity and backend work.

### Decision

After reviewing the impact and feasibility of each option, we’ll proceed with `Option A`, applying the following changes:

- Deprecate the `/agents/outdated` endpoint.
- Keep only the “Outdated” indicator column in the agents table, calculated by comparing the agent version with the manager version (API info).
- Remove filters and counters related to outdated agents to avoid unnecessary computations and requests.
- No pagination or total count for outdated agents will be calculated.

This change is considered a temporary patch, aimed at improving performance and stability until a more complete server solution is implemented.

PR #7796, opened 2025-10-07
Deprecate the use of /agents/outdated endpoint

gonzaarancibia:
### Description
Deprecate the `/agents/outdated` endpoint used to determine whether an agent is outdated.
Now, the agent version is compared directly against the manager version (retrieved from the API info endpoint).
If the agent version is older than the manager version, the agent will be considered outdated.
 
### Issues Resolved
https://github.com/wazuh/wazuh-dashboard-plugins/issues/7794

### Evidence

Screenshot 2025-10-07 at 10 49 44 AM

### Test
- Go to Endpoints summary. 
- Verify that the dashboard no longer calls `/agents/outdated` and instead uses the API info endpoint.
- Confirm that outdated filters are no longer available, and only the red outdated icon is shown in the agents table.

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/plugins/main/public/components/endpoints-summary/services/get-api-version.ts
+++ b/plugins/main/public/components/endpoints-summary/services/get-api-version.ts
@@ -1,8 +1,11 @@
-import { UIErrorLog, UILogLevel } from './../../../react-services/error-orchestrator/types';
-import { UI_LOGGER_LEVELS } from "../../../../common/constants";
-import { WzRequest } from "../../../react-services";
-import { getErrorOrchestrator } from "../../../react-services/common-services";
-import { UI_ERROR_SEVERITIES } from "../../../react-services/error-orchestrator/types";
+import {
+  UIErrorLog,
+  UILogLevel,
+} from './../../../react-services/error-orchestrator/types';
+import { UI_LOGGER_LEVELS } from '../../../../common/constants';
+import { WzRequest } from '../../../react-services';
+import { getErrorOrchestrator } from '../../../react-services/common-services';
+import { UI_ERROR_SEVERITIES } from '../../../react-services/error-orchestrator/types';
 
 /**
  * Retrieves the Wazuh API version by making a GET request to the root endpoint.


plugins/main/public/components/endpoints-summary/table/utils/index.ts
```diff
@@ -0,0 +1,21 @@
...
+
+  // Normalize lengths, in case any version has fewer parts
+  const maxLength = Math.max(aParts.length, bParts.length);
+  while (aParts.length < maxLength) aParts.push(0);
+  while (bParts.length < maxLength) bParts.push(0);
+
+  // Compare part by part
+  for (let i = 0; i < maxLength; i++) {
+    if (aParts[i] < bParts[i]) return true;
+    if (aParts[i] > bParts[i]) return false;
```
guidomodarelli: ### Strip non-numeric prefixes before comparing versions

When `apiReq('GET', '/')` returns the API version string it may come back with a leading `v` (we already handle this possibility elsewhere, e.g. in `check-setup.service.ts` via `/v?(?<version>...)`), but `isVersionLower` feeds that raw string directly into `Number`. For a value like `v4.14.0` the first element becomes `Number('v4')`, i.e. `NaN`, so every comparison falls through and the function always returns `false`. This means the new logic will never mark any agent as outdated and the upgrade action remains disabled. We need to normalise the version strings (strip the optional v or any other non-numeric prefixes) before the numeric comparison.

guidomodarelli: See these lines

https://github.com/wazuh/wazuh-dashboard-plugins/blob/2fcfede0245948818198f076a5ff98d0ff74fb93/plugins/main/public/components/health-check/services/check-setup.service.ts#L45-L48

gonzaarancibia: [resolved here](https://github.com/wazuh/wazuh-dashboard-plugins/pull/7796/commits/73366c05389fbd3dc21ce1fa29a130ab0aac3267)

Machi3mfl (dismissed):
CR: ✅ 
Test: ✅ 

### The api info is retrieved from / endpoint

Image

Image

Machi3mfl (approved):

guidomodarelli:
The file `plugins/wazuh-core/common/api-info/endpoints.json` was not modified to remove the deprecated endpoint, which will no longer be used.

guidomodarelli:
There are several files that still reference the `/agents/outdated` endpoint, which should be updated to remove it, since it will no longer be used. Specifically, according to the search, the files that need to be modified are:

```console
$ rg /agents/outdated
docker/imposter/wazuh-config.yml
127:    path: /agents/outdated

docker/imposter/security/security-actions.json
40:        "GET /agents/outdated",

plugins/wazuh-core/common/api-info/security-actions.json
39:      "GET /agents/outdated",

plugins/wazuh-core/common/api-info/endpoints.json
575:        "name": "/agents/outdated",

plugins/main/server/lib/api-request-list.json
515:        "name": "/agents/outdated",

plugins/main/common/api-info/security-actions.json
38:      "GET /agents/outdated",

plugins/main/common/api-info/endpoints.json
584:        "name": "/agents/outdated",
```

PR #8068, opened 2026-01-23
Merge 4.14.4 into main

Desvelao:
Merge 4.14.4 into main
> [!CAUTION]
> Merge PR strategy: Create a merge commit


Desvelao:
# Changes

- Implement fix for `vulnerability.under_evaluation` when using filters different to `is` such as `exist` or `does not exist`

image

- Fix a problem when using the `is not` filter that did not highlight the expected filter button

Issue #8094, opened 2026-02-06
Excessive decimal precision in SCA score on Agent Overview


yenienserrano:
**Description**

In the Agent Overview tab, the Security Configuration Assessment (SCA) widget (located at the bottom right) displays the Score value with an excessive number of decimal places. This negatively impacts the UI readability and visual consistency of the dashboard.

**Steps to Reproduce**
1. Open the dashboard.
2. Navigate to the Agents section.
3. Select a specific agent to view its Overview.
4. Scroll down to the Security Configuration Assessment (SCA) widget.
5. Observe the Score column in the policies table.

**Expected Behavior**
The score should be formatted to a maximum of 2 decimal places (e.g., 44.90%) or rounded to the nearest integer for better user experience.

**Actual Behavior**
The system displays a raw floating-point number with more than 10 decimals (e.g., 44.89795918367347%), which causes visual clutter in the table row.

<img width="422" height="194" alt="Image" src="https://github.com/user-attachments/assets/9e091a20-663c-484c-b1ae-1c272d4bd900" />

PR #8099, opened 2026-02-09
Excessive decimal precision in SCA score on Agent Overview

FernandoCastilla:
### Description

Fixed the number formatter using `format:percent:defaultPattern` setting in the the Security Configuration Assessment table of the agent overview.

Side changes:
- Removed som unused files.
 
### Issues Resolved
Issue: https://github.com/wazuh/wazuh-dashboard-plugins/issues/8094

### Evidence

Evidence_issue_8094

### Test

    1. Open the dashboard and select a specific agent.
    2. Scroll down to the Security Configuration Assessment (SCA) widget(bottom right corner).
    3. Observe the Score column in the policies table.

### Check List
- [x] All tests pass
  - [x] `yarn test:jest`


FernandoCastilla:
The issue stems from the decimalFormat function. It retrieves Kibana's default Numeral.js pattern (typically 0,0.[000]%), but the convertNumeralToD3Format utility transforms this into ',.3~f'. The tilde, originally intended to trim trailing zeros, causes the D3 render to fail. Adjusting the conversion to output ',.3f' resolves the bug.

plugins/main/public/components/overview/sca/components/dashboard/utils/visualization-helpers.ts
```diff
@@ -20,23 +23,24 @@ function convertNumeralToD3Format(numeralFormat: string): string {
...
-  format += '~f';
+  format += '%';
 
   return format;
 }
 
-export const decimalFormat = () => {
+export const decimalFormat = (maxPrecision?: number) => {
   let decimalFormat;
   const pattern = core.uiSettings.get('format:percent:defaultPattern');
```
rodrigofez: Looking at this part, this setting defines the pattern of how percetanges should be displayed, by default it is this `0,0.[000]%` (you can check it in advanced settings), so it's supposed to show max 3 decimals, this is a [numeralJs](http://numeraljs.com/) pattern, I wonder why it was decided before to convert this pattern to a d3 pattern, and use the d3 formatter instead of just importing and using directly numeral (`import numeral from '@elastic/numeral';`), and use it to format it.

This fix works since the conversion of this numeral pattern to d3 is not working as expected, but the `format:percent:defaultPattern` setting won't be able to control the percentage format for this as it's supposed to(I guess).

rodrigofez (approved):
🟢 Score column no longer has an excessive decimal precission in the policies table

Screenshot 2026-02-10 at 13 17 40

It works, however I left a comment below that would be good to consider if we want to keep the formatting to be dynamic through that setting.

plugins/main/public/components/overview/sca/components/dashboard/utils/visualization-helpers.ts
Desvelao: **suggestion**: review if wen can use the `PercentFormat` (`data` plugin, `common` directory) to format the value instead of the custom formatter in our code.

In the case this can be used instead, review if the replaced function or derivated are used, else remove them.

Adman23: Resolved in commit: [23fd5f6](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8099/changes/23fd5f6a7394eed11fe9ef84f8ea829f2bbb051d)

Adman23:
While investigating the logic in visualization-helpers.ts, we discovered redundant code in plugins/main/common/dashboards/lib/helpers.ts that duplicates this logic but appears unused.

Additionally, dashboard-kpis.ts (which references decimalPattern) seems to be an orphaned file with no active imports in the project. The callback to decimalPattern has been updated. We'll perform a final search to confirm no dynamic references exist.

Machi3mfl:
@FernandoCastilla, you must run the prettier because the prettier workflow is failing. 

Screenshot 2026-02-12 at 6 30 18 PM

plugins/main/public/components/common/welcome/components/sca_scan/sca_scan.tsx
```diff
@@ -122,8 +122,8 @@ const ScaScanTable = ({ dataSourceAction }) => {
       width: '10%',
       sortable: true,
       render: score => {
-        const scoreFormat = decimalFormat();
-        return `${d3.format(scoreFormat)(score * 100)}%`;
+        const scoreFormatter = decimalFormat(2);
```
Desvelao: **nitpick**: this seems to define the decimals to render, but this should use the value of the setting. We could consider defining the format if that setting is not defined or has no value as fallback.

Adman23: Fixed in: [9d36a1e](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8099/changes/9d36a1eaf747a7985a2431284005f0fc705b56ac)

FernandoCastilla:
> While investigating the logic in visualization-helpers.ts, we discovered redundant code in plugins/main/common/dashboards/lib/helpers.ts that duplicates this logic but appears unused.
> 
> Additionally, dashboard-kpis.ts (which references decimalPattern) seems to be an orphaned file with no active imports in the project. The callback to decimalPattern has been updated. We'll perform a final search to confirm no dynamic references exist.

After further investigation we found some files that weren't being used at all. [fc94bd0](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8099/commits/fc94bd09e721775d58a311da4e3ad6f805184fb8)


Note: These files weren't necessarily related to this issue, but we found them while researching the usage of the now-updated decimalFormat() and agreed with other members that they aren't needed anymore.

rodrigofez (approved):
CR: ✅
Test: ✅

https://github.com/user-attachments/assets/73d898a1-b9e5-4d27-b3c9-e2cf69abc03c

PR #8196, opened 2026-03-18
Buttons tooltips asking for administrator role when they shouldn't 

### LINKED ISSUE #8191: Some buttons tooltips ask for administrator role when they shouldn't
## Description

Users without administrator privileges may have buttons blocked that do not require such role to be used. This is only a problem with the button tooltip, when the user has the correct role (even if its not administrator) the button is enabled.

## Steps to reproduce
1. Launch the dashboard using `./dev.sh --server-local 5.x`.
2. inside osd-dev container edit file `config/opensearch_dashboards.yml` with `run-as: true` for the manager-local.
4. Launch dashboard with yarn start and log in as admin.
5. Navigate to `Indexer management > Security > Internal users` and create a new internal user with a strong password (you can use a password generator).
6. Go to `Overview` and verify the disabled `deploy new agent` button and the toolip. It will be: 

Image

And it should be: 

Image

Adman23:
### Description
If the user had insufficient permissions some buttons could be disabled and indicate the need for administrator role, even if the button didn't need it, this PR fixes that tooltip.
 
### Issues Resolved
#8191 

### Evidence

issue_evidence

### Test
1. Launch the dashboard using `./dev.sh --server-local 5.x`.
2. inside osd-dev container edit file `config/opensearch_dashboards.yml` with `run-as: true` for the manager-local.
4. Launch dashboard with yarn start and log in as admin.
5. Navigate to `Indexer management > Security > Internal users` and create a new internal user with a strong password (you can use a password generator).
6. Log in as the new user and navigate to `Agents management > summary` and remove the agents (if any).
7. Go to `Overview` and verify the disabled `deploy new agent` button and the toolip appear like the evidence and no `administrator` role is required.

### Check List

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,6 +59,7 @@ All notable changes to the Wazuh app project will be documented in this file.
 - Fixed version hardcoded value in the deploy agent wizard [#7923](https://github.com/wazuh/wazuh-dashboard-plugins/pull/7923)
 - Fixed a visual bug in SCA score decimal precision on the Agent Overview [#8099](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8099)
 - Fixed the agent stats view was innaccesible for some version combinations [#8150](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8150)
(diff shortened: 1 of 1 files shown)


asteriscos (approved):
CR: 🟢 
Test: 🟢 

Image

PR #8409, opened 2026-04-29
Adds Mailpit email testing service for development

### LINKED ISSUE #110: [BUG] Email reports cannot be opened from a different instance
## Description

While testing the reporting plugin by sending reports through an email channel (following the use case stated in https://github.com/wazuh/wazuh-dashboard/issues/1180#issuecomment-4200389910), we have seen that the report cannot be opened from the instance where the email is received because the IP address is set to the indexer address.

If the instance from which the report is opened is different from the instance where the indexer is installed, the report cannot be opened. This means that the report must be opened from the same instance where the indexer is installed, which may not be the expected behavior.

yenienserrano:
### Description

Add the option to use Mailpit for testing email delivery in development environments
 
### Issues Resolved

- https://github.com/wazuh/wazuh-dashboard-reporting/issues/110

### Test

run the development environments with the `—mailpit` flag and without the flag

1. With this flag, you should be able to start a Mailpit container and configure the email notification channel. 
2. Without the flag, the mailpit container should not start

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`

### DIFF
--- a/docker/osd-dev/README.md
+++ b/docker/osd-dev/README.md
@@ -8,6 +8,7 @@
   `sudo sysctl -w vm.max_map_count=262144`
 
   To make the change permanent in host machine:
+
   - In host machine:
     `sudo nano /etc/sysctl.conf`
   - Add at the end of the file: vm.max_map_count=262144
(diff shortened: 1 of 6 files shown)


rodrigofez (approved):
🟢 CR
🟢 Mailpit container gets up with --mailpit flag
Image

Ripdiegozz (approved):
🟢 CR & Tests:

Image

PR #8755, opened 2026-07-10
Automate content update on plan change

Ripdiegozz:
### Description

Currently, the Wazuh cloud register component detects any change in the subscription plan, but the CTI content is never refreshed to reflect it. This PR adds a silent, server-side trigger that updates CTI content automatically whenever the CTI registration status or subscription plan changes, plus three follow-up additions requested during review (discriminated trigger reason, a configurable steady-state poll interval, and a Consumers accordion in the CTI registration modal).

### Proposed Changes

- Added a per-environment subscription snapshot and an in-flight update lock to `CtiRegistrationStore`, used to detect real transitions between polls.
- Fixed a pre-existing bug where the "CTI" button and the status badge ("Wazuh Cloud" + health indicator) could render **at the same time** when registration failed (both `showStartCtiBar` and `showStatusBar` evaluated to `true`).
- Added `triggerContentUpdateOnChange`: server-side, edge-triggered detection that fires the existing (previously unused) Content Manager `contentUpdate` action whenever the subscription transitions **into** the registered state (registration completes, or the plan name changes while already registered). The first observation after a server restart only stores a baseline and never fires.
- Wired the trigger into the existing CTI registration status route (`GET .../cti-registration/status`), reusing the `environmentUuid` the route already resolves.
- Made the trigger fully silent per product decision: the route fires it fire-and-forget (never awaited) and does not surface any outcome to the client - no response field, no toast, on success or failure.
- Hardened the trigger against two review findings:
  - The subscription snapshot is now persisted synchronously, before any `await`, so an overlapping poll can no longer have its fresher snapshot clobbered by a slower call's stale write once its request resolves.
  - The fire-and-forget call site now catches and logs (via the plugin logger) any unexpected error outside the trigger's own internal error handling, instead of allowing it to become an unhandled promise rejection.
- Renamed the header nav button next to the healthcheck/user controls from "Register" to "CTI" (icon and tooltip unchanged), since "Register" no longer matched its purpose once already registered.
- **Discriminated trigger reason**: added a `reason` (`registration-completed` | `plan-name-changed`) surfaced in the server log for observability. The fire logic itself is unchanged.
- **Configurable steady-state poll interval**: new setting `ctiRegistrationStatusPollIntervalSec` (default 30s, min 5s), mirroring the existing `ctiRegistrationUiEnabled` pattern. Drives a new post-registration re-poll, independent of the OAuth device-flow polling, with a freshness guard so a remount or tab-focus event doesn't re-fetch if already polled within the interval.
- **Consumers accordion**: shows `.wazuh-cti-consumers` details (2-column grid, Resource as a truncated link) inside the CTI registration success modal (opened via "Wazuh Cloud" once registered). Uses `asCurrentUser` (not `asInternalUser`) after hitting a real permission gap with the internal `kibanaserver` role in testing.

### Artifacts Affected

- `plugins/wazuh-check-updates` (server routes, server services, common types, public nav/modal components).

### Configuration Changes

New setting: `wazuh_check_updates.ctiRegistrationStatusPollIntervalSec` (default `30`, minimum `5`).

### Documentation Updates

None required.

### Tests Introduced

- `cti-registration-store.test.ts`
- `trigger-content-update-on-change.test.ts`
- `status.test.ts`
- `start-cti-registration.test.tsx`: updated to assert the new "CTI" label.
- `useCtiStatus.test.ts`: poll-interval clamp and freshness guard (remount + visibility-focus races).
- `consumers.test.ts` / `cti-consumers-accordion.test.tsx`: Consumers route and modal rendering.

## Check List
- [x] All tests pass
  - [x] `yarn test:jest`


asteriscos (dismissed):
CR: 🟢 
Test: 🟢 

### 30 registration status polling

Image

### Change the default value to 5 seconds

Image

### Consumers

Image

### Update content on plan change

Image

asteriscos (approved):

PR #8788, opened 2026-07-26
Add Wazuh AI Assistant plugin

MiguelCasaresRobles:
## Description

Adds the Wazuh AI Assistant plugin for Wazuh 5.0: an AI chat embedded in the dashboard that answers security questions by running read-only queries against the Wazuh Indexer and Server API and returning a short grounded answer together with the result table. It works with OpenAI-compatible providers (OpenAI, Groq, Ollama, vLLM, LiteLLM), Anthropic, and the Wazuh hosted provider.

## Proposed Changes

- New plugin `plugins/wazuh-ai-assistant/` (structure mirrors the sibling plugins):
  - 29 read-only declarative tools (`server/tools/catalog/`) plus a guarded free-search tool limited to `wazuh-events-v5-*`, `wazuh-findings-v5-*` and `wazuh-states-*`. No mutating tools.
  - Every Indexer/Manager query runs `asCurrentUser`, so the user's own RBAC is the enforcement boundary; there is no service-account path.
  - Server-side guardrails on every query: injected timeout, size clamps, bounded time windows, aggregation caps, script/regexp/leading-wildcard blocks, and an index allowlist. Query results are digested (6,000 chars, at most 5 sample rows) before reaching the provider. Optional reversible pseudonymization, SSRF guard on provider traffic, and opt-in AES-256-GCM encryption of stored API keys.
  - Provider adapters share one canonical tool-calling contract. No npm runtime dependencies.
- Reference documentation under `docs/ref/modules/ai-assistant/`, indexed in `docs/SUMMARY.md` and `docs/ref/modules/README.md`. Developer docs in `plugins/wazuh-ai-assistant/docs/`: `ENCRYPTION.md` (API-key encryption format) and `CI.md` (test gates and eval harness).
- CI: plugin added to the per-plugin matrix in `5_builderprecompiled_base-dev-environment.yml`. Unit tests run under Jest via the platform runner and ESLint uses a flat config, both following `wazuh-core`; the plugin is formatted with the repo `.prettierrc`.

### Results and Evidence

Verified on a `wazuh-dashboard` v5.0.0-beta3 build tree (Node 22.22.0 / OSD 3.6.0):

| Gate | Result |
| --- | --- |
| `yarn typecheck` (tsc `--noEmit`) | 0 errors |
| `yarn test:jest` | 443/443, 33 suites (node + jsdom projects) |
| Guardrail DSL lint (adversarial corpus) | 28/28 |
[...]

PR #8790, opened 2026-07-27
Add Wazuh AI Assistant plugin

### LINKED ISSUE #8789: Create AI assistant
## Description

Wazuh dashboard users currently have no built-in way to ask natural-language questions about their security data; investigating an alert, agent, or vulnerability requires manually building queries and dashboards. This creates friction, especially for users who are not familiar with the underlying data model or query syntax.

We need an AI Assistant plugin that lets users ask questions in plain language and get grounded answers backed by read-only queries against the Wazuh Indexer and Server API, with results returned as both a short answer and a data table.

## Requirements

- New dashboard plugin providing an embedded AI chat experience.
- Read-only tool set for querying Indexer/Manager data (alerts, agents, vulnerabilities, etc.), enforced under the querying user's own RBAC (no service-account access).
- Guardrails on all queries: timeouts, size/time-window limits, aggregation caps, and an index allowlist, to prevent expensive or unsafe queries.
- Support for multiple LLM providers: OpenAI-compatible (OpenAI, Groq, Ollama, vLLM, LiteLLM), and Anthropic.
- Admin-configurable settings for provider selection, credentials, and privacy policy.
- Documentation covering architecture, tool catalog, providers, security, and configuration.

## Additional information

Tracked for Wazuh 5.0.0. See implementation in #8790.

asteriscos:
## Description

Adds the Wazuh AI Assistant plugin for Wazuh 5.0: an AI chat embedded in the dashboard that answers security questions by running read-only queries against the Wazuh Indexer and Server API and returning a short grounded answer together with the result table. It works with OpenAI-compatible providers (OpenAI, Groq, Ollama, vLLM, LiteLLM), Anthropic, and the Wazuh hosted provider.

## Proposed Changes

- New plugin `plugins/wazuh-ai-assistant/` (structure mirrors the sibling plugins):
  - 29 read-only declarative tools (`server/tools/catalog/`) plus a guarded free-search tool limited to `wazuh-events-v5-*`, `wazuh-findings-v5-*` and `wazuh-states-*`. No mutating tools.
  - Every Indexer/Manager query runs `asCurrentUser`, so the user's own RBAC is the enforcement boundary; there is no service-account path.
  - Server-side guardrails on every query: injected timeout, size clamps, bounded time windows, aggregation caps, script/regexp/leading-wildcard blocks, and an index allowlist. Query results are digested (6,000 chars, at most 5 sample rows) before reaching the provider. Optional reversible pseudonymization, SSRF guard on provider traffic, and opt-in AES-256-GCM encryption of stored API keys.
  - Provider adapters share one canonical tool-calling contract. No npm runtime dependencies.
- Reference documentation under `docs/ref/modules/ai-assistant/`, indexed in `docs/SUMMARY.md` and `docs/ref/modules/README.md`. Developer docs in `plugins/wazuh-ai-assistant/docs/`: `ENCRYPTION.md` (API-key encryption format) and `CI.md` (test gates and eval harness).
- CI: plugin added to the per-plugin matrix in `5_builderprecompiled_base-dev-environment.yml`. Unit tests run under Jest via the platform runner and ESLint uses a flat config, both following `wazuh-core`; the plugin is formatted with the repo `.prettierrc`.

### Results and Evidence

Verified on a `wazuh-dashboard` v5.0.0-beta3 build tree (Node 22.22.0 / OSD 3.6.0):

| Gate | Result |
| --- | --- |
| `yarn typecheck` (tsc `--noEmit`) | 0 errors |
| `yarn test:jest` | 443/443, 33 suites (node + jsdom projects) |
| Guardrail DSL lint (adversarial corpus) | 28/28 |
| `yarn lint` (ESLint 9 flat config) | 0 problems |
| `prettier --check` (repo config) | clean |

Integration baseline on a 5.0.0-beta3 AIO VM: plumbing 44/44, persistence 9/9, encryption round-trip 4/4, per-tool E2E matrix 29/29 (how to run each gate is documented in `plugins/wazuh-ai-assistant/docs/CI.md`).

### Artifacts Affected

- New installable plugin zip `wazuhAiAssistant-3.6.0.zip` (built via `plugin_helpers`, like the sibling plugins).
- No changes to existing plugins.

### Configuration Changes

- New optional keys under the `wazuh_ai_assistant` config path (`opensearch_dashboards.yml` / keystore):
  - `wazuh_ai_assistant.enabled` (default `true`)
  - `wazuh_ai_assistant.encryptionKey` (default unset; API keys are then stored in plaintext and a startup warning is logged — the OSD keystore is recommended)
- Providers and the privacy policy are configured at runtime by an administrator in AI Assistant → Settings. Backward compatibility: n/a (new plugin).

### Documentation Updates

- `docs/ref/modules/ai-assistant/`: README, architecture, tool catalog, providers, security, configuration
- `docs/SUMMARY.md` and `docs/ref/modules/README.md`: index entries
- `plugins/wazuh-ai-assistant/docs/`: `ENCRYPTION.md`, `CI.md`
- `CHANGELOG.md`: entry under v5.0.0 → Added

### Tests Introduced

- 443 Jest tests (33 suites), colocated with the modules they cover and run by the platform runner in two projects: `node` for the logic tiers and `jsdom` for the React layer.
  - Logic (362): guardrail lint/valves, the AES-GCM cipher, pseudonymizer and field policy, schema validation, wire-schema handling, retry/stall watchdog, conversation merge and owner scoping, and digest extraction.
  - React (81): message bubble, result table, status callout, chat input and conversation list, plus the `useSyncedState` hook and the chat/conversations HTTP services (SSE parsing, transport-failure and 401 paths).
- A zero-dependency integration harness (`eval/`) for live-stack checks: chat-route plumbing against a scripted mock provider, persistence CRUD and owner isolation, per-tool E2E matrix with a seeded dataset, and a load driver. Documented in `plugins/wazuh-ai-assistant/docs/CI.md`.

Known gap: `chat-page.tsx` and `settings-page.tsx` have no component tests of their own — they are large enough that testing them meaningfully needs a split first. Their behaviour is exercised through the components and services above, the `eval/` harness, and manual UAT.

## Notes for reviewers

- The plugin registers its own left-nav category; it can be moved into the 5.0 navigation groups if that is the preferred convention.
- Three CI checks fail for infrastructure reasons on this fork PR, not because of its content:
  - **Ensure the code format** — the workflow runs `git fetch origin $GITHUB_HEAD_REF` against the base repo, which cannot see a fork branch (`fatal: couldn't find remote ref`), so it exits before checking anything. `prettier --check` over the whole plugin passes locally with the repo config (prettier 2.8.8, the version the root lockfile resolves).
  - **verify-changelog** — `gh pr diff` returns HTTP 406 above 20,000 diff lines. `CHANGELOG.md` is modified; the check cannot see it.
  - **Run unit tests** — the plugin's own leg passes (`Test Suites: 33 passed`, `Tests: 443 passed`); the job is then marked failed by the `comment-test-coverage` step, which cannot post to a fork PR (`Resource not accessible by integration`). Because the matrix is `fail-fast`, whichever leg reaches that step first also cancels the others.

  All three are limitations of running these workflows from a fork rather than a branch in the base repo, and are worth fixing in the shared workflows independently of this PR.

## Review Checklist

- [x] Code changes reviewed
- [ ] Correct labels applied


plugins/wazuh-ai-assistant/common/discover-url.ts
```diff
@@ -0,0 +1,183 @@
...
+  from: string;
+  to: string;
+}
+
+export const DEFAULT_TIME_RANGE: TimeRange = { from: 'now-24h', to: 'now' };
+
+// Time fields a tool's DSL range clause might use, to reconstruct the Discover time window:
+// @timestamp (Wazuh 5.0 findings-v5/events-v5), state.modified_at (the wazuh-states-* families),
+// and legacy `timestamp` (4.14). extractTimeRange scans for whichever is present.
+const TIMESTAMP_FIELDS = ['@timestamp', 'state.modified_at', 'timestamp'];
```
Desvelao, line 20: **nitpick**: remove the `timestamp` field because this is related to 4.x and it wi ll not available in 5.x. I guess we could need to add other date fields here.

plugins/wazuh-ai-assistant/common/discover-url.ts
Desvelao, line 1: **suggestion**: research if we can use the `rison-node` library as done in `wazuh` plugin.

plugins/wazuh-ai-assistant/common/conversation-merge.ts
```diff
@@ -0,0 +1,66 @@
...
+import { ChatMessage } from './types';
+
+/** Structural equality for one message — deliberately NOT reference equality, since the two arrays
+ * being merged are independently-fetched/independently-held copies, never the same object. */
+function messagesEqual(a: ChatMessage, b: ChatMessage): boolean {
+  return (
+    a.role === b.role &&
+    a.content === b.content &&
+    a.toolCallId === b.toolCallId &&
+    JSON.stringify(a.toolCalls ?? null) === JSON.stringify(b.toolCalls ?? null)
```
Desvelao, line 32: **suggestion**: this comparation is flaky, it will work in the most of cases.

plugins/wazuh-ai-assistant/common/chat-history.test.ts
Desvelao, line 1: **suggestion**: use the expect of jest environment

Desvelao:
# Request changes

R1. Hide references to Wazuh brain provider. Not available at the moment.
R2. Clear references related to `alerts` to the expected entity (states, findings, events)
R3. Change fields in search to 5.0 schema (e.g rule.level -> wazuh.rule.level)

plugins/wazuh-ai-assistant/common/http-status.test.ts
Desvelao, line 1: **suggestion**: use `expect` isntead of `assert`.

plugins/wazuh-ai-assistant/eval/corpus.json
```diff
@@ -0,0 +1,342 @@
...
+      "id": "active_agents",
+      "category": "agents",
+      "prompt_en": "How many agents are currently online and reporting to the manager?",
+      "prompt_es": "¿Cuántos agentes están actualmente conectados y reportando al manager?",
+      "expect": {
+        "tool": "get_active_agents",
+        "params_subset": {},
+        "table": "required_empty_ok"
+      },
+      "notes": "UAT intent GET_ACTIVE_AGENTS -> server/tools/catalog/get_active_agents.ts. Manager GET /agents?status=active&q=id!=000 (excludes the manager's own pseudo-agent, get_active_agents.ts:20-25)."
```
Desvelao, line 23:
```suggestion
      "notes": "UAT intent GET_ACTIVE_AGENTS -> server/tools/catalog/get-active-agents.ts. Manager GET /agents?status=active."
```

plugins/wazuh-ai-assistant/eval/corpus.json
```diff
@@ -0,0 +1,342 @@
...
+      "id": "critical_alerts",
+      "category": "alerts",
+      "prompt_en": "Show me the critical-severity alerts from the past week, most recent first.",
+      "prompt_es": "Muéstrame las alertas de severidad crítica de la última semana, empezando por las más recientes.",
+      "expect": {
+        "tool": "get_critical_alerts",
+        "params_subset": {},
+        "table": "required_empty_ok"
+      },
+      "notes": "UAT intent GET_CRITICAL_ALERTS -> get_critical_alerts.ts, hardcodes rule.level>=15 (get_critical_alerts.ts:38); not a model-supplied param, so nothing to pin beyond tool identity."
```
Desvelao, line 47:
```suggestion
      "notes": "UAT intent GET_CRITICAL_ALERTS -> get_critical_alerts.ts, hardcodes wazuh.rule.level is critical (get-critical-alerts.ts:47); not a model-supplied param, so nothing to pin beyond tool identity."
```

yenienserrano:
## Review: Wazuh AI Assistant plugin

Context note: this PR merges the accumulated `enhancement/8789-create-dashboard-assistant-plugin` feature branch (where #8788 was an earlier intermediate merge) into `5.0.0` — verified via commit ancestry, not a duplicate of #8788.

New self-contained plugin (`plugins/wazuh-ai-assistant/`, 162 files, ~36k lines): an AI chat embedded in the dashboard that runs read-only, RBAC-scoped (`asCurrentUser`, no service account) queries against the Indexer/Server API through 29 declarative tools + one guarded free-search tool, with server-side guardrails, optional pseudonymization, SSRF-guarded provider calls, and opt-in AES-256-GCM API-key encryption. Overall this is unusually well-documented and defensively engineered — most of the obvious IDOR/SSRF/encryption/RBAC pitfalls are already closed with real, tested logic. Findings below survived independent re-verification against the source.

### Findings (ranked)

**1. (Medium-High, security) Aggregation-type gap in `guardrails.ts` also defeats `privacy.ts`'s field scrubbing**
`checkAggs()` (`server/tools/guardrails.ts:242-246,751`) only recognizes `terms`/`cardinality`/`significant_terms` for the low-cardinality field allowlist. It never checks `rare_terms` or `significant_text` — both bucket aggregations whose keys are raw field values. Since `search_wazuh_data` (`server/tools/catalog/search-wazuh-data.ts`) accepts an arbitrary `aggs` body against `wazuh-events-v5-*`/`wazuh-findings-v5-*`/`wazuh-states-*`, a call like `{"aggs":{"leak":{"rare_terms":{"field":"data.srcip","max_doc_count":100000}}}}` sails through `lintDsl` untouched, dumping every distinct source IP in the window.

Worse, `privacy.ts`'s `extractAggFields()` (`server/tools/privacy.ts:591-604`) has the *identical* omission — it also only reads `terms`/`significant_terms`/`cardinality`. So even with pseudonymization enabled, a `rare_terms` bucket's field resolves to `undefined` and its values pass through `applyFieldPolicy` completely unscrubbed. One aggregation-type gap simultaneously defeats both the guardrail and the privacy guarantee, sending raw IPs/usernames to the external LLM provider. A smaller instance of the same class: `composite` sources are only field-checked when the source uses `.terms` (guardrails.ts:763-793) — a `histogram`/`date_histogram`/`geotile_grid` composite source skips the allowlist too.

**Fix:** extend both `TERMS_LIKE_AGG_KEYS` (or an equivalent check) and `extractAggFields`'s type list to cover `rare_terms`, `significant_text`, and non-`terms` composite sources.

**2. (Medium, correctness/reliability) `chat-page.tsx` doesn't guard sidebar actions during an active stream**
`handleNewConversation`/`handleSelectConversation`/`handleDeleteConversation` (lines ~479-544) mutate `messages`/`activeConversationId` unconditionally, with no `isGenerating` check and no abort of the in-flight stream. Repro: send a message, then click "New conversation" while it's still streaming — the abandoned stream keeps running server-side, its deltas become no-ops against the swapped-out message array, and when it finishes its `finally` block persists the *new* (unrelated) conversation, silently dropping the completed turn. If a second message is sent before the first stream ends, `abortControllerRef` gets overwritten, and the first stream's `finally` later nulls it out from under the second — the Stop button stops working and a third overlapping stream becomes possible.

**3. (Low-Medium) `/chat` route has no size bound on `messages`/`content`, unlike its sibling `conversations.ts`**
`server/routes/chat.ts:326-352` — `content: schema.string()` (no `maxLength`), `messages: schema.arrayOf(...)` (no `maxSize`). `conversations.ts:252,267-269` bounds the identical shape before persisting, with an explicit comment that `/chat`'s copy is "intentionally NOT bounded the same way ... only what gets PERSISTED needs this bound." That reasoning covers storage growth but misses that `/chat` itself re-scrubs the whole array via regex on every tool round and forwards it to a billed upstream LLM — an unbounded payload here is a real cost/CPU lever, not just a storage one.

**4. (Low) Two provider routes skip the shared error wrapper**
`server/routes/settings.ts:569-601` and `:624-726` (`POST .../default`, `POST .../test`) aren't wrapped in `withInternalErrorHandling(...)` like every other mutating route in the file — inconsistent error-response shape on unexpected failures, not exploitable.

**5. (Low) No `maxItems` bound on `agent_names`**
`server/tools/catalog/search-alerts-by-multiple-agents.ts:35-39` — unlike `router.ts`'s `minItems`/`maxItems` pattern (which `schema-validator.ts` fully supports), this array param is unbounded; backstopped only by OpenSearch's own `index.max_terms_count` default, not by this plugin's own guardrails.

**6. (Low) `useSyncedState`'s "ref is synchronously updated" guarantee likely breaks on the 2nd+ back-to-back setter call in one tick**
`public/hooks/use-synced-state.ts:21-39` — plausible, not confirmed by a failing test. `chat-page.tsx`'s stream-end `finally` block calls the setter up to three times with no `await` between them, then immediately reads the ref in `persistConversationAfterTurn`. Untested edge case; recommend a test exercising two synchronous calls, or switching to a plain mutable ref instead of relying on React's eager-update optimization.

**7. (Low) No double-submit guard on the provider Save button**
`public/components/settings/settings-page.tsx:494-535` / button at ~1032-1046 — no `isLoading`/in-flight guard, unlike the Privacy/Retention save buttons in the same file. Double-click while creating a provider → duplicate rows.

**8. (Nit, convention) CHANGELOG entry links to the PR, not the issue**
`CHANGELOG.md:39` links to `#8788` (a PR). Per this repo's convention, changelog entries should link to the *issue* (`#8789`), not a PR.

**9. (Nit) Per-provider behavioral divergence**: `openai-compatible.ts` never sets a `max_tokens` cap while `anthropic.ts` hardcodes 4096 — real inconsistency in what's billed as "one canonical contract," though not a bug per se. Also `get-vulnerabilities-by-agent.ts` fetches but never displays `vulnerability.description`, unlike its three sibling tools (cosmetic drift).

### What's solid (verified, not just asserted)
- **RBAC**: every Indexer/Manager call goes through `asCurrentUser` — no service-account fallback found anywhere in the tool-execution path.
- **SSRF guard** (`url-guard.ts`): blocks metadata/link-local ranges (incl. `169.254.169.254`, embedded-v4 IPv6), resolves DNS before checking, documented residual TOCTOU tradeoff.
- **Encryption** (`api-key-cipher.ts`): AES-256-GCM, fresh IV per call, tag verified on decrypt, AAD binds ciphertext to the saved-object id, matches `ENCRYPTION.md`.
- **Owner isolation** on conversations: fails closed, never trusts a client-supplied owner field, 404s (not 403) to avoid confirming another user's row exists.
- **Admin gating** on provider/settings mutation routes: server-side, fails closed, strict boolean check.
- **XSS**: no `dangerouslySetInnerHTML`/`innerHTML` anywhere in `public/`; markdown rendering uses EUI's default (HTML-stripping) pipeline; settings page never round-trips a real/masked API key back into the form.
- **Layering**: no `public/` → `server/` imports found.
- Retry/SSE-stall handling, stream concurrency caps, and version-conflict handling on conversation writes are genuinely wired into the real routes, not just unit-tested in isolation.

### Recommendation
Fix #1 before merge — it's a real, if narrow, data-exfiltration path through the guardrail/privacy layer this PR advertises as the core safety story. #2 is a real UX/data-loss bug worth fixing given how central the chat stream is. The rest (#3-9) are reasonable as fast-follows; none block the architecture.

plugins/wazuh-ai-assistant/eval/corpus.json
```diff
@@ -0,0 +1,342 @@
...
+      "prompt_es": "¿Qué alertas ha generado web-prod-01 con severidad alta o superior?",
+      "expect": {
+        "tool": "search_alerts_by_agent",
+        "params_subset": {
+          "agent_name": "web-prod-01",
+          "min_severity": "high"
+        },
+        "table": "required_empty_ok"
+      },
+      "notes": "UAT intent SEARCH_ALERTS_BY_AGENT -> search_alerts_by_agent.ts. Both agent_name and min_severity are explicitly stated in the question, so both are pinned. 5.0 UPDATE: rule.level is categorical in Wazuh 5.0 (informational/low/medium/high/critical), so min_severity is a WORD, not a 0-15 number -- the prompt and the pinned value were both migrated off the 4.14 numeric scale (the pre-migration case asked for 'severity of 10 or higher' and pinned 10, which the model correctly answered with \"high\", failing the stale assertion)."
```
Desvelao, line 62:
```suggestion
      "notes": "UAT intent SEARCH_ALERTS_BY_AGENT -> search-alerts-by-agent.ts. Both agent_name and min_severity are explicitly stated in the question, so both are pinned."
```

plugins/wazuh-ai-assistant/eval/corpus.json
```diff
@@ -0,0 +1,342 @@
...
+      "id": "critical_vulnerabilities",
+      "category": "vulnerabilities",
+      "prompt_en": "List the critical vulnerabilities currently affecting our fleet and which agents they're on.",
+      "prompt_es": "Lista las vulnerabilidades críticas que afectan actualmente a nuestra flota y en qué agentes están.",
+      "expect": {
+        "tool": "get_critical_vulnerabilities",
+        "params_subset": {},
+        "table": "required_empty_ok"
+      },
+      "notes": "UAT intent GET_CRITICAL_VULNERABILITIES -> get_critical_vulnerabilities.ts, wazuh-states-vulnerabilities-* filtered to severity=Critical (hardcoded, get_critical_vulnerabilities.ts:26). Distinguish from get_vulnerabilities (no severity floor) -- see honesty_on_empty case below."
```
Desvelao, line 86:
```suggestion
      "notes": "UAT intent GET_CRITICAL_VULNERABILITIES -> get-critical-vulnerabilities.ts, wazuh-states-vulnerabilities-* filtered to severity=Critical (hardcoded, get-critical-vulnerabilities.ts:36). Distinguish from get-vulnerabilities (no severity floor) -- see honesty_on_empty case below."
```

plugins/wazuh-ai-assistant/eval/corpus.json
```diff
@@ -0,0 +1,342 @@
...
+      "id": "alerts_last_24h",
+      "category": "alerts",
+      "prompt_en": "What's happened across all our agents in the last 24 hours?",
+      "prompt_es": "¿Qué ha pasado en todos nuestros agentes en las últimas 24 horas?",
+      "expect": {
+        "tool": "get_alerts_by_time",
+        "params_subset": {},
+        "table": "required_empty_ok"
+      },
+      "notes": "UAT intent GET_ALERTS_LAST_24H -> get_alerts_by_time.ts, the generalized port (get_alerts_by_time.ts explains it replaces the 4.14 hardcoded now-24h so ALL severities show, not just critical). time_range_gte deliberately left unpinned -- see params_subset_note in _meta."
```
Desvelao, line 98:
```suggestion
      "notes": "UAT intent GET_ALERTS_LAST_24H -> get-alerts-by-time.ts, the generalized port (get-alerts-by-time.ts explains it replaces the 4.14 hardcoded now-24h so ALL severities show, not just critical). time_range_gte deliberately left unpinned -- see params_subset_note in _meta."
```

plugins/wazuh-ai-assistant/eval/corpus.json
```diff
@@ -0,0 +1,342 @@
...
+      "id": "brute_force",
+      "category": "alerts",
+      "prompt_en": "Are we seeing any brute-force login attempts lately?",
+      "prompt_es": "¿Estamos viendo intentos de fuerza bruta en los inicios de sesión últimamente?",
+      "expect": {
+        "tool": "get_brute_force",
+        "params_subset": {},
+        "table": "required_empty_ok"
+      },
+      "notes": "UAT intent GET_BRUTE_FORCE -> get_brute_force.ts. The 4.14 *brute* wildcard was rebuilt as a non-wildcard bool.should (get_brute_force.ts) to satisfy the lint."
```
Desvelao, line 110:
```suggestion
      "notes": "UAT intent GET_BRUTE_FORCE -> get-brute-force.ts. The 4.14 *brute* wildcard was rebuilt as a non-wildcard bool.should (get_brute_force.ts) to satisfy the lint."
```

plugins/wazuh-ai-assistant/eval/corpus.json
Desvelao, line 1: **thought**: the file seems that is not used. This should be reviewed.

plugins/wazuh-ai-assistant/eval/lint_cases.json
Desvelao, line 1: **nitpick**: this needs to be adapted to 5.0 schema.

plugins/wazuh-ai-assistant/public/components/chat/chat-input.tsx
```diff
@@ -0,0 +1,119 @@
...
+            // isGenerating is folded in here (not just at the parent's `disabled` prop) so typing
+            // or pressing Enter mid-stream can't be silently swallowed by handleSend's own
+            // isGenerating early-return below.
+            disabled={disabled || isGenerating}
+            value={value}
+            placeholder={i18n.translate(
+              'wazuhAiAssistant.chat.inputPlaceholder',
+              {
+                defaultMessage:
+                  'Ask the AI Assistant about your security alerts...',
```
Desvelao, line 79:
```suggestion
                  'Ask the AI Assistant about your security events...',
```

felipegonzalezmv:
# Code review — PR #8790 · Wazuh AI Assistant plugin

**Scope reviewed:** `git diff ffa0eff61..HEAD` on `enhancement/8789-create-dashboard-assistant-plugin`
— 179 files, ~36.5k insertions. New plugin at `plugins/wazuh-ai-assistant/`, plus dev-environment,
docs and CI wiring.

Only findings judged **vital** are listed (real bugs, security exposure, resource leaks/DoS,
serious convention violations). Nitpicks (naming, formatting preferences, minor a11y) are
deliberately omitted. Every finding below was verified by reading the code, not inferred.

---

## 1. Overview

The plugin adds a provider-agnostic AI chat over Wazuh data:

- **`server/`** — SSE chat route running a two-stage orchestration loop (route → act), a catalog of
  29 read-only tools (Indexer `_search` + Wazuh Manager API), server-side query guardrails, a
  pseudonymization layer, three provider adapters (OpenAI-compatible, Anthropic, Wazuh Brain), and
  AES-256-GCM encryption at rest for provider API keys.
- **`public/`** — React 18 chat + settings UI consuming the SSE stream.
- **`common/`** — shared stream/message contract and pure helpers.
- Three `hidden: true` saved-object types (provider, settings singleton, conversation).

### What is genuinely well done

This is unusually solid for largely AI-authored code, and several things that normally go wrong here
are right:

- **No privilege escalation.** Every Indexer call is `context.core.opensearch.client.asCurrentUser`
  and every Manager call `context.wazuh_core.api.client.asCurrentUser`. There is no
  `asInternalUser` anywhere in the tool path. The user's own RBAC is the real boundary, exactly as
  `docs/ref/modules/ai-assistant/security.md` claims.
- **No query injection.** Model-supplied strings only ever land in `term`/`terms`/`match`/`prefix`
  *values* — never in field paths, index names, or Lucene syntax. No `query_string` is ever built
  from model text. The escape hatch's `index_pattern` is enum-locked *and* re-checked against an
  anchored allowlist at execution time, so `.kibana`/security indices are unreachable.
- **Guardrails are enforced server-side at execution**, not merely suggested in the prompt.
- **Layering is clean.** Exhaustive grep confirms zero `public/ → server/` or `server/ → public/`
  imports; `common/` is genuinely isomorphic (no DOM, no Node builtins — `draft-stash.ts` even takes
  an injected `Storage` subset).
- **Secrets are handled correctly.** API keys are write-only from the UI (`hasApiKey: boolean` is
  all that ever leaves the server), `exposeToBrowser: {}`, the encryption key is never logged, and
  the `enc:v2` AAD binding against the saved-object id is a genuinely thoughtful touch.
- **SSRF guard** on provider URLs with DNS re-resolution and `redirect: 'error'`; residual DNS
  rebinding is documented rather than hidden.
- **Conversations are owner-scoped** and return 404 (never 403) cross-owner, so existence is not
  leaked.
- Retry logic honors `Retry-After`, clamps backoff, and does not retry the non-idempotent
  `wazuh_brain` webhook. Stream-slot accounting is correct and idempotent on all exit paths.

---

## 2. Blockers

### B1 · Privacy mode does not redact nested objects or arrays — High

`server/tools/privacy.ts:657-680`

`applyFieldPolicy` only pseudonymizes when `typeof value === 'string'`. Every other value type falls
through to `out[field] = value`, unredacted — in both the `anonymize` branch and the escape hatch's
fail-closed branch.

**Failure scenario.** With privacy mode ON, the model calls `search_wazuh_data` with
`{"_source": ["source"]}`. `deriveResultColumns` (`server/tools/digest.ts:275-284`) makes `source`
the sample column, `getByPath` returns the whole sub-object, and the digest ships
`source.ip` / `source.user.name` verbatim to the third-party provider — even though both have an
explicit `anonymize` entry at `privacy.ts:79-85`. The same hole hits typed tools whenever an ECS
field is multi-valued (an array `source.ip` skips the branch entirely). The chat-route pre-scan
catches embedded IPv4/IPv6/FQDNs but not usernames, bare hostnames, or command lines.

**Fix.** `deepMapStrings` already exists in the same file (`privacy.ts:168`). Replace the
`typeof value === 'string'` guard in both branches with a `deepMapStrings` walk over every string
leaf, and drop the whole subtree for `action: 'never'`. `privacy.test.ts` has no array/nested-object
case — add one.

### B2 · Markdown images give prompt injection a zero-click exfiltration channel — High

`public/components/chat/message-bubble.tsx:139`

Model output renders through a bare `<EuiMarkdownFormat>`. OUI's default pipeline is safe against
XSS (raw HTML nodes are dropped; `markdownLinkValidator` neutralizes non-`http(s)` links) — but that
validator visits **`link` nodes only**. `image` nodes pass straight through to an ``.

**Failure scenario.** Attacker-controlled text in an ingested alert instructs the model to emit
``. The analyst's browser silently issues the
GET. OSD sets no `img-src` CSP directive, so nothing blocks it. This is data exfiltration with no
user interaction, in a product whose whole threat model assumes alert text is attacker-influenced.

**Fix.** Pass a `processingPluginList` to `EuiMarkdownFormat` whose `rehype-react` `components` map
overrides `img` — render nothing, or only same-origin `src`.

### B3 · Switching conversation mid-stream silently destroys the answer — High

`public/components/chat/chat-page.tsx:479` (`handleNewConversation`), `:502`
(`handleSelectConversation`)

Neither handler aborts the in-flight stream, and `ConversationList` is never disabled while
generating (only `ChatInput` receives `isGenerating`, line 1585).

**Failure scenario.** User clicks another conversation while a response streams. `updateMessages`
replaces state wholesale, so every subsequent `current.map(m => m.id === assistantMessageId ? …)`
in `runChatStream` (lines 697, 728, 780, 791) matches nothing: remaining deltas, the table, and the
terminal `isStreaming: false` are all dropped. Conversation A's answer is lost and never persisted.
Worse, the stream keeps running (burning provider tokens), `isGenerating` stays true so B's input
stays disabled, privacy-map entries minted for A are merged into B's `pseudonymMap` (line 817) and
sent with B's next request, and `persistConversationAfterTurn()` (line 884) runs against B's id.

**Fix.** `abortControllerRef.current?.abort()` at the top of both handlers, and in
`handleDeleteConversation`'s active-id branch.

### B4 · Unbounded SSE line buffer can OOM the whole dashboard process — High

`server/providers/sse-utils.ts:162-165`

`buffer += decoder.decode(...)` grows without limit; the only bound is the idle watchdog, which
resets on *every* chunk. A peer streaming continuously with no `\n` never trips it.

**Failure scenario.** A compromised or simply broken provider gateway (or a self-hosted model server
with a bad framer) streams newline-free bytes. Node's heap grows until OOM — taking down the entire
OSD process for every user, not just assistant users.

**Fix.** After the `buffer +=`, check a `MAX_SSE_LINE_CHARS` ceiling (~1 MB) and
`throw new ProviderStallError('Provider sent an oversized SSE frame.')`. Both adapters' existing
`catch` already turn that into a clean terminal `error` event.

### B5 · Header timeout rejects but never aborts the fetch — socket leak — High

`server/providers/retry.ts:188-245`

On timeout, `fetchWithHeaderTimeout` does `clearTimeout` + `reject(new ProviderStallError())`. It
has no `AbortController` of its own (the doc comment concedes this), and the request-level `signal`
passed to `doFetch` is not aborted here. The `if (settled) response.body?.cancel()` branch only
helps if the fetch eventually settles.

**Failure scenario.** A provider that completes the TCP handshake then blackholes (firewall DROP, a
hung gateway). Every chat turn leaks one hanging undici request and socket. The 5-concurrent-stream
cap does not help — it releases as soon as the 30s `ProviderStallError` fires — so a user clicking
send repeatedly accumulates hung sockets until the process hits its fd/pool ceiling.

**Fix.** Create a per-attempt `AbortController` chained to `signal`, pass its signal to `fetch`, and
`abort()` it from both the timeout and `onAbort` paths.

### B6 · CI format gate fails — Blocker for merge

`prettier@2.7.1 --check --config .prettierrc` fails on four changed files:

- `docker/osd-dev/scripts/__tests__/helpers/setupTests.ts`
- `docker/osd-dev/scripts/src/services/repoResolver.test.ts`
- `docs/ref/modules/ai-assistant/providers.md`
- `docs/ref/modules/ai-assistant/tool-catalog.md`

Two are gratuitous reformat regressions in unrelated docker files (`} catch {}` → `} catch { }`).
Run `prettier --write` on those paths and drop the unrelated edits.

---

> **B7** is a seventh blocker. It is documented at the top of §3 rather than here because it began
> as a medium-severity guardrail finding and was promoted after cross-checking a parallel review —
> keeping it adjacent to the other guardrail items makes it easier to fix them together.

---

## 3. Should fix before merge

### S1 · Chat route body has no size bounds, and the pseudonymizer is quadratic

`server/routes/chat.ts:326-371` and `server/tools/privacy.ts:265-280`

The chat body schema puts **no** `maxSize` on `messages`, no `maxLength` on `content`, and no
`maxSize` on `privacy.map`. `conversations.ts` bounds all three; the comment at
`conversations.ts:242` explicitly says chat's copy is "intentionally NOT bounded… only what gets
PERSISTED needs this bound" — but the chat route does CPU work proportional to those values.

`applyToText` re-sorts `[...this.valueToPseudonym.keys()]` on **every call**, then runs an
`includes()` scan per key. `scrubMessagesForProvider` calls it once per message; `orchestrate` calls
that once per round plus stage 1.

**Failure scenario.** Within OSD's 1 MB default payload, a user posts ~20 000 `privacy.map` entries
and ~1 500 short messages with `privacy.enabled: true`. Each round performs ~1 500 × (20k-key sort +
20k substring scans), repeated across stage 1 and up to 4 rounds — hundreds of millions of
synchronous string ops on the single event loop, freezing the dashboard for every user. The
5-stream cap doesn't help: the cost is CPU-blocking, not concurrency.

**Fix.** Add `maxSize`/`maxLength` to all three, and cache the length-sorted key array inside
`Pseudonymizer`, invalidating only in `pseudonymize()`.

### S2 · Mandatory time bound is satisfied by a `range` anywhere in the body

`server/tools/guardrails.ts:524-526`

`hasTimeRange` starts `walkRequiredForTimeRange` at the **body root** with `required = true` and
recurses generically, so a `range` in `post_filter`, `aggs.*.filter` or `rescore` counts.

**Failure scenario.** `{"query":{"match_all":{}},"post_filter":{"range":{"@timestamp":{"gte":"now-1h"}}},"size":500}`
passes the lint while the executed query scans the entire events/findings corpus — precisely the
unbounded timeline scan the check exists to prevent.

**Fix.** Run the traversal over `body.query` only, returning `false` when `body.query` is absent.

### B7 · Aggregation-type gap defeats the guardrail *and* the privacy layer — High

`server/tools/guardrails.ts:744-853` (`TERMS_LIKE_AGG_KEYS` at `:242`) **and**
`server/tools/privacy.ts:590-597`

*(Raised to blocker after cross-checking a second reviewer's finding — the privacy half turns what
looked like a DoS lever into a data-exfiltration path.)*

The field allowlist and `MAX_AGG_SIZE` only apply to `terms`, `cardinality`, `significant_terms`,
`composite` and `multi_terms`. `rare_terms`, `significant_text`, `top_metrics`, `geo*_grid` and
`shard_size` are checked by neither, so
`{"aggs":{"a":{"rare_terms":{"field":"data.srcip","max_doc_count":100000}}}}` passes `lintDsl`
untouched.

**The exfiltration half.** `extractAggFields` (`privacy.ts:590-597`) has an *overlapping but
narrower* type list — `terms`, `significant_terms`, `cardinality` only. For any aggregation type it
doesn't recognize, `field` resolves to `undefined`, and `applyFieldPolicy`'s breakdown loop
(`privacy.ts:690-693`) pushes the bucket **unscrubbed**. So even with pseudonymization ON, a
`rare_terms` aggregation's bucket keys — raw source IPs, usernames — reach the external provider.

**And the two lists disagree, not merely share a gap.** `composite` and `multi_terms` are
field-checked by the guardrail but are invisible to `extractAggFields`. Since `agent.name` is in
`AGG_FIELD_ALLOWLIST` (`guardrails.ts:197`) *and* carries `action: 'anonymize'` (`privacy.ts:34`), a
perfectly legal `multi_terms` aggregation on `agent.name` passes the guardrail and then leaks real
hostnames — no unrecognized aggregation type required.

Smaller instance of the same class: `composite` sources are only field-checked when the source uses
`.terms` (`guardrails.ts:763-793`), so a `geotile_grid` or `histogram` source skips the allowlist.

**Fix.** Invert the guardrail to deny-by-default, clamp `shard_size` alongside `size`, and derive
`extractAggFields`' type list from the *same* constant so the two cannot drift again. Add a test
asserting every guardrail-permitted aggregation type is resolvable by `extractAggFields`.

### S4 · Tool execution has no signal and no timeout

`server/tools/executor.ts:140-143, 208-213`, called from `server/routes/chat.ts:841`

`signal` is in scope in `orchestrate` but never threaded into `executeToolCall`. Neither
`asCurrentUser.search` nor the Manager client call carries a `signal` or `requestTimeout`.

**Failure scenario.** A 90-day `search_wazuh_data` takes 3 minutes on a loaded cluster. The user
closes the tab; `aborted$` fires and releases the stream slot, but the generator is still parked on
the `await`, the cluster keeps working, and the generator's `.return()` (and therefore
`iterateSseLines`' `finally`) cannot run until the query settles.

**Fix.** Thread `signal` down and pass `{ signal, requestTimeout: 30000 }`; add a `Promise.race`
timeout on the Manager path.

### S5 · No wall-clock cap on stream lifetime — slow reader pins slots

`server/routes/chat.ts:463-544`

`readWithStallTimeout` only arms a timer when a `reader.read()` is actually requested. Under
consumer backpressure `Readable.from` stops pulling the generator, no read is requested, and no
timeout runs.

**Failure scenario.** An authenticated user opens 5 chat requests with a zero TCP receive window and
never reads. All 5 per-user slots are pinned indefinitely; six accounts pin
`MAX_CONCURRENT_STREAMS_GLOBAL = 30` and the assistant returns 429 for everyone until restart.

**Fix.** Alongside the `aborted$` subscription, arm a ~10 min `setTimeout` that calls
`controller.abort()` + `releaseStreamSlot()`, cleared in the existing `finally`.

### S6 · `MAX_TOOL_ROUNDS` bounds rounds, not tool executions (Anthropic path)

`server/providers/anthropic.ts:158-182`, `server/routes/chat.ts:805-894`

The Anthropic adapter yields a `tool_call` per `content_block_stop` and then does
`toolUses.delete(blockIndex); continue;` — it keeps reading the same stream. (The OpenAI adapter
correctly `return`s after `finalizeToolCalls`.) `orchestrate` awaits `executeToolCall` and
`continue`s, so every block runs a real query. `disable_parallel_tool_use: true` is *requested*
(`anthropic.ts:98`) but not enforced server-side.

**Failure scenario.** A loop-prone model emits ~100 `tool_use` blocks within its 4096-token budget:
~100 sequential OpenSearch searches per round × 4 rounds, and `messages` grows by 200 entries of up
to 6000-char digests that are re-sent and re-scrubbed every subsequent round.

**Fix.** Add a per-turn `toolCallsThisTurn` counter (cap ~8); on exceeding it, append
`toolErrorContent('Tool-call budget exhausted for this turn.')` and break out of the stream read.

### S7 · `useSyncedState`'s core guarantee doesn't hold under React 18

`public/hooks/use-synced-state.ts:27-36`

The ref is written **inside** the `setValue` updater. The doc comment justifies this with React 16
semantics, but `public/application.tsx:134` mounts with React 18 `createRoot`, which batches updates
from async contexts. React 18 only evaluates an updater eagerly when the fiber has no pending lanes,
so the second and later updates in a batch defer — and the ref write never happens for them.

**Failure scenario.** In `runChatStream`, a fast provider whose chunks are already buffered resolves
several `reader.read()` promises as microtasks with no render in between, so two
`flushPendingDelta()` commits land in one batch. `messagesRef.current` misses the second;
`persistConversationAfterTurn` (`chat-page.tsx:623`) then saves a transcript with the answer's tail
truncated.

**Fix.** Compute `next` outside the updater and write the ref there:
`const next = typeof update === 'function' ? update(ref.current) : update; ref.current = next; setValue(next);`

### S8 · Unrelated dev-environment behaviour change bundled into this PR

`docker/osd-dev/config/osd/opensearch_dashboards.yml:29,41` — `run_as: false` → `run_as: true` on
both the `manager` and `manager-local` hosts.

This is not a plugin file and it changes Manager API auth for **every** plugin in the local dev
stack: with `run_as: true`, wazuh-core issues API calls under the dashboard user's RBAC identity, so
any developer whose dashboard user isn't mapped in the manager's RBAC now gets 401/403 on agents,
rules, Dev Tools where it previously worked.

**Fix.** Revert, or split into its own PR with a stated rationale. If the assistant's tool calls
genuinely require it, document that in `docs/ref/modules/ai-assistant/configuration.md` rather than
flipping a shared default silently.

---

## 4. Worth addressing (lower priority)

| # | Location | Issue |
|---|---|---|
| L1 | `server/providers/sse-utils.ts:178-181` | `finally` calls `releaseLock()` but never `reader.cancel()`. Both adapters `return` out of the `for await` on their terminal frame, so on every normal turn the upstream body is left unconsumed and uncancelled — the socket isn't returned to the pool and is only reclaimed by GC. Add `reader.cancel().catch(() => undefined)`. |
| L2 | `server/providers/openai-compatible.ts:269-274`, `anthropic.ts:165-171` | Raw, untruncated provider output is interpolated into a client-facing `error` event without passing through `sanitizeProviderErrorBody` — the redaction+truncation helper every other provider error in the plugin uses. |
| L3 | `server/routes/chat.ts:729-738` | The stage-1 router generator is abandoned without `.return()` on abort, so the nested adapter stream and `iterateSseLines`' `finally` never unwind (the abort listener is never removed). Wrap the drive loop in `try/finally { await stage1.return(undefined) }`. |
| L4 | `common/constants.ts:98-100` | `1000 messages × 100 000 chars` allows a ~100 MB conversation on paper, while OSD's default `server.maxPayloadBytes` is 1 MB. `toPersistedMessages` trims to the *declared* limits, so a conversation whose serialized body crosses 1 MB still 413s and auto-save fails silently — the exact bug the constants' doc comment says was fixed. Add a total-byte budget to the client-side trim. |
| L5 | `public/components/chat/chat-page.tsx:651`, `public/services/conversations-service.ts:36` | `persistConversationAfterTurn` calls `refreshConversations()` after **every** turn, and `list()` pages at 100 up to 20 serial round-trips; each GET also runs server-side retention deletes. A user with 800 conversations pays 8 sequential requests after every assistant message. Patch the single changed summary into state from the POST/PUT response instead. |
| L6 | `opensearch_dashboards.json:3`, `package.json:3-4` | `5.0.0-00` / revision `00` while all sibling plugins are at `5.0.0-05`. The build doesn't break (artifact naming derives from `plugins/main`), but the shipped manifest reports the wrong revision until the next bump. |
| L7 | `CHANGELOG.md:39` | Entry links the PR (`/pull/8788`); `CLAUDE.md` and `docs/dev/pull-requests.md` require linking the **issue**, as every neighbouring entry does. |
| L8 | `server/routes/settings.ts:569, 624, 731, 752, 776` | Five routes skip `withInternalErrorHandling(...)` while the other four in the file use it — inconsistent error shape on unexpected failures, and a raw saved-objects error can surface where the others return a normalized 500. |
| L9 | `server/tools/catalog/search-alerts-by-multiple-agents.ts:38` | `agent_names` has no `maxItems`, unlike every other bounded array param in the catalog. The model controls the array; the only backstop is OpenSearch's `index.max_terms_count`, not this plugin. |
| L10 | `public/components/settings/settings-page.tsx:1035` | No in-flight guard on the provider Save button (only `canSave`), unlike the Privacy/Retention buttons in the same file. Double-click while creating → duplicate provider rows. |
| L11 | `server/providers/openai-compatible.ts` | Never sets `max_tokens`, while `anthropic.ts:84` sends 4096. For a plugin presenting one canonical provider contract with every other resource dimension clamped, worth aligning or documenting as deliberate. |
| L12 | `server/wazuh-core.d.ts:32-122`, `server/identity.ts:41-53` | Cross-plugin contracts are hand-copied structural types plus a **global** augmentation of the platform's `RequestHandlerContext` from inside a plugin. If wazuh-core changes `isAdministratorUser`'s shape nothing fails to compile — the admin gate would just read `undefined` and (fail-closed) 403 every provider mutation. If `plugins/main` ever declares its own differently-typed `wazuh_core`, TS fails with "Subsequent property declarations must have the same type". At minimum add a compile-time assignability assertion so drift is caught by CI. |

---

## 5. Design decision needing an explicit product sign-off

**Privacy mode is OFF by default and redaction is allow-by-omission for typed tools.**

`DEFAULT_ASSISTANT_SETTINGS.privacyDefaultOn: false` (`server/routes/settings.ts:38`) means
`privacyCtx` is `undefined` on a fresh install and `finalizeDigest` is a documented byte-identical
passthrough. Compounding it, `resolveFieldEntry` (`privacy.ts:552`) returns `undefined` → **allow**
for any unlisted field on all typed tools; only the escape hatch is fail-closed.

So out of the box, an operator who points the assistant at a hosted provider ships production
hostnames, internal IPs, usernames and process command lines off-site with no redaction and no
warning. `field-policy-coverage.test.ts` is a good guard against new fields leaking, but its
`KNOWN_SAFE_STRUCTURAL_FIELDS` escape valve lets a reviewer wave a field through, and it doesn't
cover `tableSpec.rowFields` or the Manager `message` field.

The docs are honest about this (`security.md` § "What leaves the cluster"), and self-hosted gateways
are the stated primary target — so this is a deliberate posture, not an oversight. But it deserves a
conscious product decision rather than defaulting by omission. Suggested minimum: default
`privacyDefaultOn` to `true` for any provider that is not demonstrably self-hosted, and flip typed
tools to fail-closed now that `field-policy-coverage.test.ts` proves every declared column is
classified.

---

## 6. Test coverage

33 test files for 93 source files. The pure-logic security surface is well covered (cipher,
pseudonymizer, guardrail lint/valves, schema validator, URL guard, digest, owner-scoping helpers,
stream limiter, admin gate).

The gap is the **orchestration and UI layers** — the largest and most stateful files have no
colocated tests at all:

| Lines | File |
|---|---|
| 1597 | `public/components/chat/chat-page.tsx` |
| 1444 | `public/components/settings/settings-page.tsx` |
| 968 | `server/routes/chat.ts` |
| 844 | `server/routes/settings.ts` |
| 572 | `server/routes/conversations.ts` |
| 332 | `server/tools/executor.ts` |

Three of the blockers above (B3, S6, S7) live in exactly that untested region. At minimum, add tests
for the abort-on-conversation-switch path and the orchestration loop's tool-call budget.

Also note that `eval/` (~7k lines of harness plus a Python seeder) is excluded from ESLint, knip and
typecheck. It is not shipped in the built package (`plugin-helpers` only copies
`{lib,server,common,translations}` + manifests), so this is repo hygiene rather than a runtime
concern — but the `NODE_TLS_REJECT_UNAUTHORIZED=0` guidance in `eval/README.md` should stay clearly
scoped to dev.

---

## 7. Style note

The code carries an unusually high volume of explanatory prose — many files open with 40–60 line
essays, and several comments narrate the change history ("a pre-release format", "the exact pre-fix
call shape", "before this fix"). Per `STYLEGUIDE.md`, comments should state constraints the code
can't express, not the PR's reasoning. Once merged, "before this fix" has no referent for the next
reader. This is not a merge blocker, but a pruning pass over the doc comments would meaningfully
reduce the maintenance surface.

---

## 8. Suggested merge checklist

- [ ] B1 — recursive redaction in `applyFieldPolicy` (+ nested/array test)
- [ ] B2 — disable image rendering in `EuiMarkdownFormat`
- [ ] B3 — abort in-flight stream on conversation switch/new/delete
- [ ] B4 — cap the SSE line buffer
- [ ] B5 — per-attempt `AbortController` in `fetchWithHeaderTimeout`
- [ ] B6 — `prettier --write` the four failing files; revert unrelated `catch { }` edits
- [ ] B7 — deny-by-default agg types, clamp `shard_size`, and share one type list between
      `guardrails.ts` and `privacy.ts`'s `extractAggFields`
- [ ] S1–S8 triaged (S1, S2, S8 recommended before merge)
- [ ] §5 privacy default explicitly signed off by product/security
- [ ] CHANGELOG entry relinked to the issue; version bumped to `5.0.0-05`

yenienserrano:
## Test

<details><summary>Differences Between the Chat and What's Saved in the History</summary>

When you're in a chat and, for example, it returns a list of agents, and then you switch to a different chat, the list is lost

image

image

```
{
    "id": "9b923810-89d0-11f1-a5f2-79da1add011b",
    "title": "How many agents do i have?",
    "createdAt": "2026-07-27T15:34:08.129Z",
    "updatedAt": "2026-07-27T15:36:56.503Z",
    "messages": [
        {
            "role": "user",
            "content": "How many agents do i have?"
        },
        {
            "role": "assistant",
            "content": "Please reload the page and sign in again. Your session has expired or is missing, which is preventing me from retrieving the current list of agents for you."
        },
        {
            "role": "user",
            "content": "How many agents do i have?"
        },
        {
            "role": "assistant",
            "content": "You have a total of 2 active agents currently connected to the Wazuh manager.\n\nThe count reflects all currently active connections at this time.\n\nWould you like me to list the names or IP addresses of these specific agents?"
        }
    ],
    "version": "WzUxNiwxXQ=="
}
```

image



</details> 

<details><summary>Multiple conversations</summary>
<p>

If you send something to a chat and then switch to a different chat, the message gets lost

1. Sent a second message in X ("...incident response plan...") — it started streaming (Stop button visible).
2. Clicked "New conversation" while it was still streaming.
3. The UI switched to a brand-new, empty conversation (welcome screen) — but the "Stop" button was still showing on that empty screen and its input stayed disabled. This is the desync symptom exactly as predicted: the abandoned stream from X was still holding the global isGenerating/abortControllerRef state, now bleeding into the new conversation's UI.
4. Waited (~45s) for X's abandoned stream to actually finish server-side. The stray "Stop" then reverted to "Send" on the new empty conversation.
5. Re-fetched conversation X via the API: messages array and updatedAt are byte-identical to before step 1 — still just question1+answer1. The second question and its completed answer are gone.
6. Checked the full conversations list: no conversation anywhere contains the incident-response-plan content. It wasn't misdirected into the new conversation either — it was simply never persisted anywhere, despite the assistant actually finishing that turn server-side.


https://github.com/user-attachments/assets/7b36f9d4-b5fb-4df3-a70b-cedd6623d827

If you send a new chat and then change it or try to create another chat, the first new chat will be lost, and you won't be able to send another message. 

https://github.com/user-attachments/assets/bcb7f718-4c35-4b1d-8f31-37328e369e00


</p>
</details>

Ripdiegozz (commented):
While testing the chat manually I found two issues.

**Finding 1: a text-based tool call leaks into the chat as raw markup**

Some providers and models (observed with an OpenRouter-hosted Nemotron model, and also with Ling-3.0-flash) do not always populate the structured `tool_calls` field on the streamed response. Instead they emit the tool call as literal `<tool_call><function=...><parameter=...>` (or `<tool_call>name<arg_key>...</arg_key><arg_value>...</arg_value>`) markup inside the plain response text. The tool call itself does run and the real results come back correctly, but the raw markup is not stripped from what gets shown to the user, so it appears as literal text above the results table instead of a clean assistant reply.

Steps to reproduce: ask a question that requires a tool call while using a provider/model prone to this behavior.

Image

**Finding 2: switching conversations mid response can silently lose the turn**

Sending a message starts a stream, and switching to a different saved conversation, or starting a new one, while that stream is still in flight does not cancel it.

Reproduced this directly with a provider that returned a rate limit error: send a question, then switch to another conversation before the response comes back. When the rate limit error arrives afterward, none of it is visible. The error banner is gone, so there is no record anywhere in the UI of why the turn failed, and the assistant bubble for that turn is left empty with no reply and no table.

Net effect: send a message, switch conversations before the response arrives, and the conversation ends up showing only the user message and an empty assistant bubble, with no error explaining what happened and no way to recover that context afterward.

Steps to reproduce: send a question, then switch to another saved conversation or start a new one before the provider responds, ideally with a provider likely to fail (rate limit works well for this).

Image

Image

jcorredor-spec (commented):
| # | Test | Result |
| --- | --- | --- |
| 1 | Result table survives resuming a conversation | ❌ table is lost |
| 2 | Switching conversation while an answer streams | ❌ turn lost, composer stuck on "Stop" |
| 3 | Malformed tool call from the model | ❌ turn dies, empty answer persisted |
| 4 | `eval/seed_uat_dataset.py` against the current 5.0 templates | ❌ 222/222 findings rejected |
| 5 | Plugin loads (`plugin:wazuhAiAssistant` green), app renders | ✅ |
| 6 | Provider CRUD + Test connection (mock and Groq) | ✅ |
| 7 | SSRF guard on provider URL (7 hostile URLs) | ✅ all blocked |
| 8 | Chat → tool → real indexer query → result table | ✅ |
| 9 | Real LLM: tool selection, grounded answers, multi-turn | ✅ 4/6 questions, answers verified |
| 10 | Guardrails: script / wildcard / index / size | ✅ |
| 11 | Privacy mode pseudonymizes the provider payload | ✅ |
| 12 | Stop button aborts and keeps the partial answer | ✅ |
| 13 | Provider rate-limit retry against real 429s | ✅ |
| 14 | Startup warning when `encryptionKey` is not set | ✅ (server log only) |
| 15 | `yarn test:jest` | ✅ 443/443, 33 suites |

<details><summary><b>1 · ❌ The result table disappears when a conversation is resumed</b></summary>
<p>

Ask something that returns a table, click **New conversation**, then click the saved conversation in the sidebar: the same turn now shows only the text answer — the table, its rows and the "Open in Discover" link are gone, while the answer text still refers to them ("here are the 10 critical alerts…").

Turn with the 10-row result table
Image

Same conversation after resuming it — the table is gone
Image

Confirmed at the API level, not inferred: `GET /api/wazuh_ai_assistant/conversations/{id}` only ever stores `role` + `content`:

```json
"messages": [
  { "role": "user",      "content": "Show me the critical alerts of the last 24 hours" },
  { "role": "assistant", "content": "MOCK ANSWER: {\"tool\":\"get_critical_alerts\",\"counts\":{\"total\":24,\"returned\":10,…" }
]
```

`toPersistedMessages` (`common/chat-history.ts`) drops `table` on purpose and
`handleSelectConversation` rebuilds the messages without it.

**Suggestion:** persist the `TableSpec` (or at least tool name + params + row count) and rebuild it on resume. If re-running the query on resume is preferred instead (fresher data, smaller saved objects), the UI should say so rather than silently showing nothing.

</p>
</details>

<details><summary><b>2 · ❌ A turn sent right before switching conversation is lost, and the composer stays stuck on "Stop"</b></summary>
<p>

1. Conversation X had 1 completed turn (2 messages).
2. Sent a second question there whose answer streams for ~25 s. Streaming started, **Stop** visible.
3. Clicked **New conversation** while it was still streaming.
4. The UI switched to a brand-new empty conversation (welcome screen) — but it still showed **Stop**
   and the input stayed **disabled**. The abandoned stream from X still owns the global
   `isGenerating` / `abortControllerRef` state, now bleeding into the new conversation.
5. Waited ~28 s for X's stream to finish server-side. The stray Stop then reverted to Send.
6. Re-read every conversation through the API: X is byte-identical to before step 2 — same `updatedAt`, same 2 messages — and no other conversation holds the answer. The model produced a  complete answer, the server streamed it in full, and it exists nowhere.

```json
{ "totalConversations": 1,
  "details": [{ "messageCount": 2, "updatedAt": "2026-07-27T01:45:30.632Z" }] }
```

Answer streaming in conversation X
11-streaming-in-progress


New empty conversation still showing Stop, input disabled
12-BUG-new-conversation-stuck-stop


Neither `handleNewConversation` nor `handleSelectConversation` aborts the in-flight stream, so its deltas target a message id that no longer exists in `messages`, and the auto-save in `runChatStream`'s `finally` reads an empty `messagesRef.current` and returns early. The same applies to switching to another saved conversation, which is the more common action. The Stop button itself works correctly (14), so the abort machinery is fine — it is just not wired into these two handlers.

**Suggestion:** abort on any conversation switch, or better, finish and persist the in-flight turn against the conversation it belongs to (capture `activeConversationId` + the message list at send time). At minimum, block/confirm the switch while `isGenerating` is true.

</p>
</details>

<details><summary><b>3 · ❌ A malformed tool call kills the turn, shows a developer-facing error, and saves an empty answer</b></summary>
<p>

Only reproducible with a real LLM. `llama-3.3-70b-versatile` on Groq sometimes emits the tool name and its arguments concatenated into the *name* field, and the whole turn dies:

```
Something went wrong tool call validation failed: attempted to call tool 'get_active_agents={"limit": "500"}' which was not in request.tools
```

Something went wrong — tool call validation failed
Image

Reproduced twice, on two different tools (`get_active_agents`, `get_critical_vulnerabilities`) — 2 of 6 real questions failed this way. A sibling variant comes straight from Groq: `Failed to call a function. Please adjust your prompt. See 'failed_generation' for more details.`

Two separate problems:

1. **No recovery.** When the tool *arguments* are wrong, the executor returns the error to the model
   for bounded self-correction, and that path works well (12). When the tool *name* is malformed, the
   turn hard-fails with zero retries and no answer.
2. **The failure is persisted as an empty assistant message.** After a reload, the saved conversation
   shows a blank bubble with no hint that anything failed (`contentLength: 0` via the API).

Resumed conversation — empty assistant bubble, no trace of the error
Image

**Suggestion:** tolerate the `name={json}` shape (split on the first `=` and parse the remainder, the same way the arguments path already tolerates model sloppiness), or at minimum retry once with a corrective tool message. And never persist an assistant turn whose content is empty.

</p>
</details>

<details><summary><b>4 · ❌ <code>eval/seed_uat_dataset.py</code> cannot seed findings on the current stack</b></summary>
<p>

Running the harness's own seeding script as documented rejects every finding:

```
findings: sent 222, errors=True, rejected=222
REJECT: mapper_parsing_exception — failed to parse field [rule.mitre.technique] of type [keyword]
=== what the AI Assistant can now see ===
findings total: 0
```

The script writes `rule.mitre = {"technique": {"id": …, "name": …}, "tactic": {"name": …}}`
(`eval/seed_uat_dataset.py:122`), but the current `wazuh-findings-v5-*` template maps `technique`,
`tactic` and `subtechnique` as **keyword**:

```json
"mitre": { "properties": { "subtechnique": {"type":"keyword"}, "tactic": {"type":"keyword"}, "technique": {"type":"keyword"} } }
```

Vulnerabilities, events and FIM documents are accepted — only the findings (the alerts the assistant is mostly asked about) are dropped, and the script's own summary then prints `findings total: 0`. Flattening the shape to strings locally made all 222 documents land, which is what the rest of this report runs on.

**Suggestion:** fix the shape in the script (or read it from the live mapping), otherwise anyone following `docs/CI.md` ends up with an assistant that has zero alerts to talk about, with no error unless they read the reject lines.

</p>
</details>

<details><summary><b>5 · ✅ The plugin loads and the app renders</b></summary>
<p>

`GET /api/status` after `./dev.sh up` + `yarn start`:

```json
{ "id": "plugin:wazuhAiAssistant@3.6.0", "state": "green", "message": "All dependencies are available" }
```

The app renders at `/app/wazuhAiAssistant` with the Chat/Settings tabs, the conversation sidebar and the "No AI provider configured" empty state, which links to the provider form. No console errors, and `dev.yml` + `scripts/src/constants/app.ts` pick the plugin up automatically.

</p>
</details>

<details><summary><b>6 · ✅ Provider CRUD and Test connection</b></summary>
<p>

Created both providers through the UI. The row is created, the first provider is automatically marked default, and the API key column correctly reads *Not set* / *Configured*. `Test` returned **OK (34 ms)** for the mock and **OK (326 ms)** for Groq, updating the status chip in place.

`GET /api/wazuh_ai_assistant/providers` never returns the key itself, only `hasApiKey`.

</p>
</details>

<details><summary><b>7 · ✅ SSRF guard on provider URLs</b></summary>
<p>

Eight `POST /providers` calls from an authenticated admin session:

| baseUrl | Result |
| --- | --- |
| `http://169.254.169.254/latest/meta-data` | 400 — blocked link-local/metadata range |
| `http://metadata.google.internal/computeMetadata/v1` | 400 — blocked cloud-metadata endpoint |
| `http://100.100.100.200/` (Alibaba) | 400 — blocked range |
| `http://192.0.0.192/` (Oracle) | 400 — blocked range |
| `file:///etc/passwd` | 400 — only http(s) allowed by policy |
| `http://[fe80::1]/v1` | 400 — blocked range |
| `http://[::ffff:169.254.169.254]/v1` | 400 — blocked range |
| `http://10.0.0.5:11434/v1` (RFC1918) | 200 — allowed by design (self-hosted gateways) |

The IPv6-mapped-IPv4 case, the one that is easy to get wrong, is correctly caught.

Side note: running the same battery without the `wz-token` Manager cookie returns
`403 … (No token provider)` *before* the URL policy runs — the admin gate fails closed, which is right, but it means any API-only client (including the `eval/` harness) must establish the Manager session first or it can never reach these routes.

</p>
</details>

<details><summary><b>8 · ✅ Chat → tool → real indexer query → result table</b></summary>
<p>

`get_critical_alerts` ran for real against `wazuh-findings-v5*` with the caller's own credentials. The digest came back as `{"total":24,"returned":10,"truncated":true}`, matching the seeded dataset (24 critical findings over 7 days), and the table rendered 10 rows with Time / Agent / Description / Level / Rule ID, severity colouring, per-row expansion and the **Open in Discover** link. Streaming was smooth and the pane auto-scrolled to the new turn.

</p>
</details>

<details><summary><b>9 · ✅ Real LLM: tool selection, grounded answers, multi-turn</b></summary>
<p>

Questions asked in plain language, no hints. Every number was re-checked by querying the indexer / Server API directly:

| Question | Answer | Ground truth |
| --- | --- | --- |
| "Show me the critical alerts of the last 24 hours" | 5 critical alerts, 2× malware drop + 3× Shellshock, agents web-prod-01/02 | 5 ✔ |
| follow-up: "which of those two agents is the most affected?" | web-prod-01 (3) vs web-prod-02 (2) | 3 / 2 ✔ |
| "Any brute force attempts today?" | 42 attempts, all on web-prod-01 from 198.51.100.77 | 42, same agent/IP ✔ |
| "What are the top 5 rules firing this week?" | 2501=31, 550=30, 5501=29, 5502=24 | matches (listed 4 of the 5) ✔ |
| "Which agents are disconnected?" | 1 agent, id 002 | exactly what the Server API returned ✔ |

Real LLM — grounded answer with its result table
Image

Real LLM — multi-turn follow-up answered from history
Image

The follow-up was answered **without a new query**, from the digest kept in history — that mechanism works. Multi-tool turns also work (one question chained `get_active_agents` → `get_disconnected_agents`). The other 2 of the 6 questions failed with the tool-name problem in 3.

Note: the `imposter` Server API mock returns the same single agent (`002 wazuh-manager-master-0`, whose own status is `active`) for both `status=active` and `status=disconnected`, so agent answers can only be checked for faithfulness to the API, not for real-world correctness.

</p>
</details>

<details><summary><b>10 · ✅ Guardrails: script, wildcard, index allowlist, size clamp</b></summary>
<p>

Driven through the real chat route with hostile DSL on the free-search tool:

| Case | Result |
| --- | --- |
| `script` in the query | rejected — *Scripted queries, sorts, or aggregations … are not allowed* |
| leading wildcard (`*admin`) | rejected — *"wildcard" clauses with a leading wildcard ("\*"/"?") are not allowed* |
| no `@timestamp` range | rejected — *Queries against "wazuh-findings-v5-\*" must include a "range" clause on "@timestamp"* |
| `index_pattern: ".kibana"` | rejected — *Property "index_pattern" must be one of [wazuh-findings-v5-\*, wazuh-e…]* |
| `size: 10000` | accepted, clamped, query executed (73 rows returned, no error) |

Each rejection comes back to the model as a correctable tool error rather than killing the turn.

</p>
</details>

<details><summary><b>11 · ✅ Privacy mode pseudonymizes the provider payload</b></summary>
<p>

With privacy on, the payload the provider actually received (captured on the provider side, not
inferred) carried pseudonyms while the UI kept showing the real values:

```json
{"tool":"get_brute_force","counts":{"total":42,"returned":5,"truncated":true},
 "samples":[{"@timestamp":"2026-07-28T01:27:43.000Z","wazuh.agent.name":"HOST_1",
 "rule.description":"Attempt to login using a non-existent user","rule.level":"high",
 "source.user.name":"USER_1","source.ip":"IP_1"}]}
```

The same turn rendered `web-prod-01` and `198.51.100.77` in the result table for the analyst.

</p>
</details>

<details><summary><b>12 · ✅ Stop button</b></summary>
<p>

Stopping mid-stream aborts immediately, keeps the partial answer on screen, restores **Send**, and the partial turn is auto-saved as a conversation. The abort machinery works — it is simply not reused by the conversation-switch handlers (2).

</p>
</details>

<details><summary><b>13 · ✅ Provider rate-limit retry against real 429s</b></summary>
<p>

Groq's free tier returned real 429s repeatedly during the session. The server honoured the backoff, retried, and surfaced the wait inline in the assistant bubble as *"Provider rate limit reached — retrying in 30s…"*, then completed the answer normally. When the daily budget was finally exhausted it gave up cleanly instead of hanging.

</p>
</details>

<details><summary><b>14· ✅ Encryption-at-rest startup warning</b></summary>
<p>

`docker/osd-dev` does not set `wazuh_ai_assistant.encryptionKey`, and the plugin logs exactly what `docs/ENCRYPTION.md` promises:

```
["warning","plugins","wazuhAiAssistant"] wazuhAiAssistant: provider API key encryption at rest is DISABLED — provider API keys are stored UNENCRYPTED in the saved-objects index. To enable, set a base64 32-byte key as wazuh_ai_assistant.encryptionKey via the OSD keystore …
```

**Suggestion:** this only reaches the server log, so an admin using the UI never sees it. The same sentence belongs in Settings → Providers, next to the field where the key is typed.

</p>
</details>

<details><summary><b>15 · ✅ Unit tests</b></summary>
<p>

`yarn test:jest` inside the dev container:

```
Test Suites: 33 passed, 33 total
Tests:       443 passed, 443 total
Coverage:    51.3% statements
```

Matches the PR description exactly.

</p>
</details>

<details><summary><b>Other observations (not tied to a single test)</b></summary>
<p>

**The router narrows what the model sees, not what it can run.** In one turn the model was offered only `search_wazuh_data`, answered with a call to `get_brute_force`, and the server executed it — the routed subset is not re-checked at execution time. Harmless today (every tool is read-only, RBAC-bound and guardrailed), but worth stating explicitly, or enforcing.

**Tab state is not in the URL.** Chat/Settings do not change the URL, so a reload always lands on Chat and Settings cannot be linked to.

**Provider state resets on reload.** The `Status` column goes back to *Not tested*, and the selected provider resets to the default — so resuming a conversation started on one provider silently continues on another.

**Token cost per question is high on a free tier.** Each question costs at least three provider calls (stage-1 router + tool round + answer) with the tool catalog in the prompt. In this session ~11 real questions consumed **94,976 of Groq's 100,000 tokens/day** — exhausted in roughly 15 minutes of testing. The provider-side metrics show where the ceiling actually is: requests peak at ~10/min against a 30/min limit, while *total tokens* spike to ~13.4K and cross the rate-limit line three times in the same window. The same chart also shows a steady trickle of HTTP 400s alongside the 429s, consistent
with the tool-format rejections in 3.

Provider metrics — requests well under the limit, tokens over it
Image

**Provider errors are dumped raw to the user.** The 429 is rendered as the provider's entire JSON body, including the organization id and a billing URL:

```
Provider responded with HTTP 429: {"error":{"message":"Rate limit reached for model
`llama-3.3-70b-versatile` in organization `org_01k…` service tier `on_demand` on tokens per day (TPD): Limit 100000, Used 94976, Requested 5632. Please try again in 8m45.312s. Need more tokens? Upgrade to Dev Tier today at https://console.groq.com/settings/billing","type":"tokens","code":"rate_limit_exceeded"}}
```

Raw provider 429 JSON rendered in the chat
Image

A short *"the AI provider is rate-limited, try again in ~9 minutes"* with the detail behind a
"show details" toggle would be enough. The same applies to the `request.tools` message in 3.

</p>
</details>

yenienserrano (approved):
LGTM :green_circle:

Issue #8797, opened 2026-07-28
AI assistant improvements


asteriscos:
## Description

The objective of this issue is to improve the following bugs and user experience items in the AI assistant plugin.

#### UI improvements

- [x] Hide references to the "Wazuh AI assistant brain" provider as it's not available yet.
- [x] In the AI settings view, automatically test the AI connection.
- [x] In the AI settings > Privacy card, allow filtering and hiding the list of all fields.
- [x] Notify when a button is disabled because of RBAC permissions.
- [x] Move the form to create an AI assistant provider to flyout
- [x] Enhancements in AI assistant Chat view

#### Adapt 4.x fields to 5.0.0 fields and indices

- [x] Clarify references to `alerts`, should map to the correct 5.0 entities (`wazuh-states`, `wazuh-findings`, `wazuh-events`).
- [x] Update search fields to the 5.0 schema (e.g. `rule.level` to `wazuh.rule.level`).
- [x] Verify the predefined available fields list exists in Wazuh 5.0.0 schema. Adapt the field name if needed.

#### AI session improvements

- [x] Fix persisting the current session state.
- [x] Fix retrieving a previous session.
  - [x] Verify the integrity of the previous responses
- [x] Support navigation while a query is being processed.
- [x] Improve retrieving the current session state; sometimes the chat is blank or in an infinite loading state.

#### Tools

- [x] Agent tools ignore pending/never_connected agents
- [x] No generic "find document(s) by field" capability
- [x] Compliance tools are hardcoded to PCI DSS
- [x] No tools cover Security Analytics

#### Other improvements

- [x] Improve key storing.

Issue #39, opened 2017-10-19
Any chance for wazuh-app 2.1.1 for ES 5.5.0 ?


rossengeorgiev:
Or if it's not possible, point me to instruction on how to build it myself.

malbarral:
Hi @rossengeorgiev, 

The Wazuh app 2.1.0 supports Wazuh 2.1.1. I have generated for you a package for Kibana 5.5.0, but it has not been tested. It is based in the package https://packages.wazuh.com/wazuhapp/wazuhapp-2.1.0_5.5.1.zip. You can download it here:

https://packages.wazuh.com/wazuh-dev/wazuhapp-2.1.0_5.5.0.zip

We will not develop for wazuh-app 2.1.1 for Elastic 5.5.0 anymore, sorry for the inconvenience.

Hope it works,
Manuel Albarral

rossengeorgiev:
Getting access denied on that URL. Is it possible to  build the packages myself?

malbarral:
Hello @rossengeorgiev,

Here you can download the package: http://packages.wazuh.com.s3.amazonaws.com/wazuh-dev/wazuhapp-2.1.0_5.5.0.zip

It will be available til the server refreshs.

Best regards,
Manuel Albarral

rossengeorgiev:
@malbarral I end up putting https://packages.wazuh.com/wazuhapp/wazuhapp-2.1.0_5.5.1.zip and ES 5.5.0. Seems to work. Overview wouldn't load, but after adding the `wazuh-monitoring-*` index pattern it started working.

PR #8811, opened 2026-07-29
AI session improvements

### LINKED ISSUE #8798: AI session improvements
- [x] Fix persisting the current session state.
- [x] Fix retrieving a previous session.
  - [x] Verify the integrity of the previous responses
- [x] Support navigation while a query is being processed.
- [x] Improve retrieving the current session state; sometimes the chat is blank or in an infinite loading state.

felipegonzalezmv:
## Description

The AI Assistant lost the conversation almost any time the user did something other than sit still
and wait. The root cause is that a session existed **only** as React state inside `ChatPage`: the
server is stateless per request and persistence was decided entirely by the browser, after a turn
finished. Everything below follows from that.

- Switching conversation, starting a new one, or leaving the app mid-answer **lost the answer**.
  Nothing aborted the stream, so the remaining deltas targeted an assistant message id that no
  longer existed; `isGenerating` stayed `true` and left the newly opened conversation's input
  blocked; the outgoing turn's pseudonym entries were merged into it; and the auto-save ran against
  the wrong conversation id.
- Visiting the **Settings tab** destroyed the chat entirely — the app shell swapped the two
  components instead of hiding one.
- A **reload or a deep link** came back to an empty chat, with no record of which conversation had
  been open, and the next turn then created a *second* saved conversation instead of continuing the
  user's.
- A **reload mid-answer lost the question too**: the conversation was only created once the turn
  ended.
- **Resuming** a conversation dropped its result tables (leaving prose describing tables that were
  no longer there), stamped every message with the moment of the resume, and lost the tool-call
  history — so the model saw prose alone and re-ran the same queries.
- An answer **cut short** was indistinguishable from a finished one, and a **failed save** was
  swallowed silently, so a user could keep chatting believing their history was being kept.

Related to #8797 — this PR implements the **AI session improvements** section of that issue. The
other sections (UI improvements, 4.x→5.0.0 field adaptation, key storage) are untouched, so the
issue stays open.

The four checklist items of that section:

| Item                                                                                | Status |
| ----------------------------------------------------------------------------------- | ------ |
| Fix persisting the current session state                                            | ✅     |
| Fix retrieving a previous session · verify the integrity of the previous responses   | ✅     |
| Support navigation while a query is being processed                                  | ✅ (client-side; see _Out of scope_) |
| Improve retrieving the current session state; sometimes the chat is blank or in an infinite loading state | ✅     |

## Proposed Changes

Seven commits, each self-contained.

**1. Keep answers when a turn is left mid-stream** (`272e1558a`)

A turn now carries its own identity and the conversation it belongs to. When it is abandoned it
stops writing shared state and persists what it produced **to the conversation it started in**,
rebuilt from its own local record rather than from React state (which by then describes a different
conversation). `Stop` is deliberately not abandonment: it keeps the turn attached to the conversation
on screen. Also fixes `useSyncedState`, whose ref mirror was written *inside* the `setValue` updater
— under React 18 `createRoot` an updater is only evaluated eagerly while the fiber has no pending
lanes, so batched delta flushes skipped the ref write and a conversation could be saved with the
answer's tail missing. Saves are serialized instead of dropped on overlap, since an abandoned turn
and the active one no longer target the same row.

**2. Restore the open conversation** (`457f19ca9`)

The app shell hides the inactive tab instead of unmounting it (Settings still mounts lazily on first
open), so a detour into Settings keeps the conversation — and keeps an in-flight answer streaming
into it. New `common/conversation-location.ts` records which conversation is open outside React: the
URL hash (`#/conversation/<id>`, shareable and bookmarkable) plus a per-tab `sessionStorage` pointer
for a bare app URL. On mount the hash wins. A conversation that no longer exists (deleted elsewhere,
or pruned by retention) is not an error: the pointer is forgotten and the user gets a clean chat.

**3. Persist faithfully, and before generating** (`80e6f5a0a`)

A turn saves twice — once when the question is sent, once when the answer completes — through one
shared target, so the second updates the row the first created instead of creating a second one.
Messages now carry their real `createdAt`, their row-capped `table`, and their tool exchanges
(interleaved as `[assistant{toolCalls}, tool{digest}]`, the shape the wire format already used).
The client-side trim also enforces a total serialized budget, shedding tables and tool history
before prose: the per-message limits bounded the payload's *shape* but not its size, so a
conversation could satisfy every limit and still be rejected as too large — a failure auto-save
swallows silently and permanently.

**4. Show turns that were cut short, and saves that failed** (`a18c3a239`)

A turn that ends without a terminal frame is marked interrupted, labelled in the transcript,
persisted that way, and can be retried (the retry *replaces* the partial answer instead of appending
a second one to the same question). A failed save now says so, and clears itself as soon as one
succeeds; a `401` keeps its own session-expired callout instead of raising both. The provider list —
which gates the whole Chat tab's rendering — gets a 20 s deadline, so a request that never settles
no longer leaves the chat spinning forever.

**5. Remove a dead constant** (`63b6296f6`)

`WAZUH_APP_CATEGORY` stopped being referenced when the app moved to the platform's `explore`
category and was already failing `yarn lint` on the plugin. Pre-existing, unrelated to the rest, kept
as its own commit.

**6. An unanswered question is a retryable turn, and show what ran** (`015c9d801`)

Marking a turn interrupted happens in the browser, so the two cases that matter most never got
marked — a reload and a navigation away kill the page first. A conversation that **ends on a
question** is now itself the evidence of an unfinished turn (the question is saved before generating
starts, so that trailing message is all a dead page leaves behind): same label, same Retry. Plus two
feedback gaps this exposed — an assistant bubble waiting for its first token was an empty box and now
shows placeholder lines, and the queries behind an answer were invisible and now appear in a
collapsed panel with their real-form arguments, plus the executed DSL and index for an indexer
search. Restored on a resumed conversation too.

**7. Documentation** (`7830d933c`)

`architecture.md` and `security.md` updated for the new persistence contract and limits.

### Technical details worth a reviewer's attention

- **No migration, by design.** `PersistedChatMessage` adds only optional fields, and
  `messages` is an `enabled: false` opaque object in the saved-object mappings, so conversations
  saved before this resume unchanged (missing fields degrade to the previous behaviour).
- **No new server route and no server-side session.** The turn still lives in the browser; what
  changed is *when* and *what* it persists.
- The abandoned-turn path deliberately reads nothing from React state — it rebuilds its transcript
  from `baseMessages` + the locally accumulated content/table/tool calls captured before the first
  `await`.

### Artifacts Affected

None. No packaging, manifest or platform-version change; `plugins/wazuh-ai-assistant` only, plus its
documentation.

### Configuration Changes

None. No new settings, no changed defaults. New internal limits live in `common/constants.ts`
(`CONVERSATION_MAX_TABLE_ROWS = 100`, `CONVERSATION_MAX_SERIALIZED_BYTES = 700_000`) and are shared
by the client trim and the server route schema so the two cannot drift.

Saved conversations do grow: they now also store result tables (row-capped) and tool-call digests.
Bounded by the serialized budget above, and by the existing per-owner conversation cap and retention
policy.

### Documentation Updates

- `docs/ref/modules/ai-assistant/architecture.md` — what the conversation saved object holds, the new
  payload limits, and how the open conversation is tracked and saved.
- `docs/ref/modules/ai-assistant/security.md` — what a saved conversation stores at rest, and that
  pseudonymization protects what goes to the provider, never what is stored locally.

### Tests Introduced

453 → **507 tests, 36 suites, all passing**. `yarn lint` and Prettier clean on the plugin.

| File                                            | Scope                                                                                                                                                                              |
| ----------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `public/components/chat/chat-page.test.tsx` (new, 32 cases) | First colocated coverage of `chat-page.tsx` — the largest untested file and where three of the code-review blockers lived. Abandoned turns (abort, save target, no state leakage into the next conversation, pseudonym isolation), conversation restore (hash vs. session pointer, gone/failed loads), save-before-generating, resume fidelity (tables, timestamps, tool history resent), interrupted turns + retry, failed-save notice, skeleton, query panel. |
| `public/application.test.tsx` (new)             | The shell's structural guarantee: the Chat tab is hidden, never unmounted, and neither page remounts on tab switching.                                                              |
| `common/conversation-location.test.ts` (new)    | Hash parse/build round-trip, id validation (rejects traversal/query strings), storage fallback, never throws when storage is unavailable.                                           |
| `common/chat-history.test.ts` (extended)        | Table row cap, tool-exchange interleaving, byte-budget degradation order, `reconstructConversation` round-trip and legacy-shape fallback.                                           |
| `public/hooks/use-synced-state.test.ts` (extended) | The React 18 batching regression: several updater calls in one render must all reach the ref.                                                                                    |

Each new test was checked to **fail against the pre-fix code** (6/8 of the abandoned-turn cases, 2/4
of the shell cases, 2 of the hook cases), so they are regression cover rather than description.

One measured fix worth noting: the first version of the byte-budget trim was O(n²) and took 178 s on
the pathological case in `chat-history.test.ts`; rewritten to a single measuring pass it is 2.7 s.

## Out of scope

- **Server-side resumable generation.** Navigation while a query is processing is handled
  client-side: the turn is saved as interrupted and offered with Retry — the generation does not
  continue in the background. A resumable stream would need a server-side job registry and a
  reattachable stream id; deliberately deferred, and a prerequisite for it is fixing the streaming
  resource issues below.
- **Streaming resource findings from the #8790 review** (unbounded SSE line buffer, socket leak on
  header timeout, no wall-clock cap on stream lifetime). Server-side, separate concern, still open.
- If the browser is killed before the abandoned turn's save request completes, the partial answer
  text is lost and the turn is presented as an unanswered question with Retry. Nothing guarantees a
  `PUT` survives an `unload` without a server-side session.

## Review Checklist

Manual verification performed: none — see _Results and Evidence_. Unit tests, lint and Prettier were
run locally; `yarn typecheck` needs the platform checkout and was not run.

Manual tests still required for approval:

- [ ] Switch conversation mid-answer → answer kept in its own conversation, new conversation's input
      usable, no cross-contamination
- [ ] Chat → Settings → Chat mid-answer → transcript intact and the answer still arriving
- [ ] Reload and navigate to another app mid-answer → same conversation restored, unanswered turn
      labelled and retryable
- [ ] Deep link to `#/conversation/<id>`, and to a non-existent id (clean start, no error banner)
- [ ] Resumed conversation shows its result tables and original timestamps; a follow-up question
      does not re-run the same queries
- [ ] Stop → Retry replaces the partial answer without duplicating the question
- [ ] Blocked `**/conversations/**` shows the not-saved callout, which clears after a successful save
- [ ] Blocked `**/providers**` ends in the load-error state, not an endless spinner

---

- [ ] Code changes reviewed
- [x] Correct labels applied (e.g., `no-changelog`) — CHANGELOG entries added, so no `no changelog`
      label


yenienserrano:
### Test

I think it would be a good idea to let the user know that if they leave the chat, the chat will be interrupted.

https://github.com/user-attachments/assets/5bbd15f4-df67-480a-a390-a90cff841b8c

felipegonzalezmv:
> ### Test
> I think it would be a good idea to let the user know that if they leave the chat, the chat will be interrupted.
> 
>  Screen.Recording.2026-07-29.at.1.50.24.PM.mov

Resolved https://github.com/wazuh/wazuh-dashboard-plugins/pull/8811/changes/014cfdf87b78c4bf45c9962a8f2cb01fd3cbc65e

Ripdiegozz (commented):
🟡 Tests

The leaving conversation alert should not trigger if I'm selecting the same chat:

https://github.com/user-attachments/assets/43711de2-1964-4d54-b5f5-51a8ac3d4ba9

felipegonzalezmv:
> 🟡 Tests
> 
> The leaving conversation alert should not trigger if I'm selecting the same chat:
> 
>  2026-07-29.14-17-40.mp4

Fix it https://github.com/wazuh/wazuh-dashboard-plugins/pull/8811/changes/07b5fe635c952b66974163c9d694e975a4b2e091

CHANGELOG.md
```diff
@@ -127,6 +129,13 @@
...
+- Fixed the AI Assistant losing a question asked while a reload or navigation interrupted the answer: the question is now saved before the response starts generating [#8797](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797)
+- Fixed a resumed AI Assistant conversation losing its result tables, its real message times, and the tool-call history a follow-up question depends on [#8797](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797)
+- Fixed the AI Assistant presenting an answer cut short by Stop, a navigation, or a dropped connection as though it were complete: it is now labelled as interrupted and can be retried [#8797](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797)
+- Fixed the AI Assistant silently failing to save a conversation, and the chat hanging on a loading spinner when the provider list never responded [#8797](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797)
+- Fixed the AI Assistant leaving a question unanswered with no way to ask it again after a reload or a navigation interrupted the response: the unanswered turn is now labelled and retryable [#8797](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797)
```
Ripdiegozz: All these new changelog entries should be associated with the entry on line `#39`

Ripdiegozz (dismissed):
🟢 Tests

https://github.com/user-attachments/assets/479fece6-11ff-4be2-9e42-d9cd1039ad48

jcorredor-spec (dismissed):
## Test

### E1 · Resuming a conversation keeps its table, times and queries

Image
Image

Opening a saved conversation brings the transcript back with its result table (paginated `Results (1 rows)`), and the timestamps are the **message's own** (`20:53`), not the moment of the resume. The collapsed `1 query executed` panel expands to show the tool that ran (`get_active_agents`) with its real arguments (`{ "limit": 500 }`).

Also checked against the API: the saved record contains the `[assistant{toolCalls}, tool{digest}]` pair, i.e. tool-call history is persisted and rehydrated, not just the prose.

### E2 · Placeholder while the first token is on its way
Image

The assistant bubble shows `Routing…` plus skeleton lines instead of an empty box.

### E3 · Settings no longer destroys the chat, not even mid-answer

Image
Image

With an answer generating, the Settings tab is opened (it mounts and loads providers and privacy normally) and coming back to Chat the transcript is intact and the turn is still running. The conversation is not lost.

### E4 · Answer cut short: label + Retry

Image

A turn that dies with the page is visibly marked `Response interrupted` with a `Retry` beside it,instead of looking like a finished answer.

### E5 · A single confirmation before interrupting

Image

Image

Both paths that can cut an answer short show **the same dialog**, with the same title (`A response is still generating`), the same body and the same button colour. Reopening the conversation already on screen asks nothing.

### E6 · The abandoned turn is saved to its own conversation

Image
Image

Image

After confirming, the target conversation shows its own content with no contamination from the outgoing turn; the abandoned turn stays in the conversation where it started, with its interruption label and its `Retry`. When text had already arrived, that partial is kept (`12-...`: the half-written answer is still in its conversation after leaving the app).

### E7 · Deep link and shareable URL

The `#/conversation/<id>` hash appears in the address bar when any conversation is opened, and navigating straight to that URL (or reloading) brings the same conversation back open. An id that no longer exists is not an error: it falls back to a clean chat.

### E8 · Reloading a completed conversation: integrity

Image

Reloading on a finished conversation brings the whole transcript back with its table and its`3 queries executed` panel.

CHANGELOG.md
felipegonzalezmv: Resolved https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797

jcorredor-spec (approved):
LGTM

PR #8812, opened 2026-07-29
Adapt 4.x fields to 5.0.0 fields and indices in AI assistant

### LINKED ISSUE #8802: Adapt 4.x fields to 5.0.0 fields and indices in AI assistant
## Description
Sub-issue of #8797, scoped to the "Adapt 4.x fields to 5.0.0 fields and indices" checklist item. The AI assistant plugin still references 4.x index/field names that need to map to the 5.0 entities.

### Requirements
- [x] Clarify references to `alerts`; map to the correct 5.0 entities (`wazuh-states`, `wazuh-findings`, `wazuh-events`).
- [x] Update search fields to the 5.0 schema (e.g. `rule.level` → `wazuh.rule.level`).
- [x] Verify the predefined available-fields list exists in the Wazuh 5.0.0 schema; adapt field names where needed.

Ripdiegozz:
## Description

Closes #8802

The AI assistant plugin's query tools, guardrails, and privacy field-policy defaults were built against 4.x/ECS-generic bare field names (`rule.level`, `rule.id`, `agent.name`, etc.). Verified live against the dev indexer: real Wazuh 5.0 documents in `wazuh-findings-v5*`/`wazuh-events-v5*` only populate the `wazuh.*` namespace (`wazuh.rule.level`, `wazuh.agent.host.*`, ...) — the bare paths never exist in real data, so every filter using them was silently matching nothing. This also meant the privacy/anonymization field policy was keyed on fields that don't exist, so anonymization was never actually firing on the real, populated fields before being sent to an external AI provider.

## Proposed Changes

- New `common/wazuh-fields.ts`: single source of truth for the `wazuh.*` field vocabulary, the (non-trivial) old→new field mapping, and the 5-value severity scale.
- Renamed every retired bare `rule.*`/`agent.*` reference to its correct `wazuh.*` equivalent across all 26 catalog tools, `guardrails.ts`'s aggregation allowlist, `privacy.ts`'s field policy, `digest.ts`, and `prompts.ts`. Per design, catalog tools keep field names as plain string literals in their query bodies (not behind an indirection layer) so they stay grep-able; a new CI test (`field-policy-coverage.test.ts`) permanently guards against any retired literal being reintroduced.
- `agent.*` is not a flat prefix rename: e.g. `agent.os.name` → `wazuh.agent.host.os.name`, `agent.ip` → `wazuh.agent.host.ip` — the irregular mapping is explicit in `wazuh-fields.ts`, not guessed.
- Retired `rule.groups` removed; `search-findings-by-rule-group.ts` retargeted to `wazuh.rule.tags`.
- `guardrails.ts` now actively rejects a numeric `range` query against `wazuh.rule.level` (now a string/keyword field), instead of silently doing a wrong lexicographic comparison.
- `result-table.tsx`'s severity classifier now has 5 genuinely distinct visual buckets (critical/high/medium/low/**informational**, each its own color) instead of folding `informational` into `low`.
- Fixed the eval harness itself (`eval/seed_uat_dataset.py`, `eval/es_discovery.js`, `eval/lint_cases.json`), which was seeding/asserting on the same stale bare field names — without this, the eval suite would have false-greened against fixtures the real pipeline never produces.
- This is a first-ever build of this plugin (no prior installations), so no data-migration path was needed — confirmed with the issue reporter before implementing.
- Separately, the alert table's Description column was bound to `wazuh.rule.description`. Switched it to `wazuh.rule.title` instead: `title` is the field that's consistently populated with a descriptive, environment-specific message (indexed with placeholders filled in from the real event data, e.g. "Wazuh Rootcheck - Rootkit or malware detected"), so it's the correct source for a human-readable description. Renamed the column label to "Title" to match. Checked every field currently referenced by this plugin against the live indexer; several are schema-valid (`wazuh.rule.category`, `wazuh.agent.host.ip`, `source.user.name`/`destination.user.name`, `host.os.name`/`host.os.platform`, `process.name`/`process.command_line`) but not populated in this dev dataset — population is decoder-dependent (this dataset only exercises a couple of decoders), so these were kept as-is rather than "fixed" toward a field with no real equivalent.
- **Slightly out of the original field-naming scope, found during manual testing**: the `min_severity` parameter on `get_alerts_by_time`/`search_alerts_by_agent`/`search_alerts_by_multiple_agents`/`search_alerts_by_os` carried over 4.x "numeric floor" semantics (`min_severity=medium` meant medium-or-above) that don't make sense now that severity is a 5-word category — asking for "informational alerts" silently returned every severity, since informational is the bottom of the list. Replaced `min_severity` with `severity` (exact match by default) plus an opt-in `severity_comparison` (`at_or_above`/`at_or_below`), used only when the user explicitly says "or above"/"or below". Happy to split this into its own PR if preferred, since it's a behavior fix rather than a field-name fix.
- Follow-up adversarial review of that change caught an inconsistency: an unrecognized/hallucinated `severity_comparison` value (e.g. a model sending `"gte"` instead of `"at_or_above"`) silently fell through to an exact match instead of failing open like an unrecognized `severity` value does. `severityFilterValues` now validates `severity_comparison` and fails open to the full severity list on anything unrecognized, consistent with the rest of this module's fail-open philosophy. Also unified a `!== undefined` vs. truthy check that differed between `get_alerts_by_time` and the other three tools, removed the now-unneeded unsafe parameter cast, and anchored the exact-match default in each tool's top-level description (not just the nested parameter schema).
- The CI regression guard against retired field literals originally only scanned the catalog tools, `digest.ts`, and `guardrails.ts` — not `privacy.ts`, the file where a reintroduced retired literal has the worst consequence (unmasked PII sent to an external provider). Expanded its scan to include `privacy.ts` and `prompts.ts`; this immediately caught a stale `GeoLocation.*` example (a retired field name) left in two doc comments, now replaced with a real field example.
- `privacy.ts`'s field-policy `.*`-suffix (prefix-match) capability existed but had no real entry using it and no test coverage. Replaced the single `wazuh.rule.compliance.pci_dss` allow-entry with a `wazuh.rule.compliance.*` wildcard, since all 10 compliance-framework subfields (hipaa, gdpr, iso_27001, nis2, ...) are equally curated, non-analyst-supplied requirement tags — confirmed via the live indexer that all 10 are `keyword`-typed. Added test coverage for the prefix match.
- During manual testing of the running app, found that the assistant has no reliable way to answer questions about compliance frameworks other than PCI DSS (no dedicated tool, and the aggregation guardrail only allow-lists `pci_dss`), producing different hallucinated answers across runs. That's a separate product gap, not a field-naming bug, so it's filed as its own issue (#8814) rather than fixed here.

### Addressed maintainer review feedback

- Removed the legacy numeric→severity-word fallback classifier from `result-table.tsx` (5.0 severity is never numeric) and rewrote the severity badge lookup to key directly off `SEVERITY_BUCKETS`, closing an unguarded-property-access gap the previous two-collection approach left open.
- `get-brute-force.ts`'s hardcoded 4.x-style numeric rule IDs (e.g. `"5710"`) could never match — `wazuh.rule.id` is always a UUID in 5.0. Verified against the live rule catalog (`wazuh-threatintel-rules-a`) that real T1110-tagged rules exist with tags like `attack.t1110`/`attack.credential-access`; removed the dead rule-id clause and corrected the tag values to match.
- Renamed `search_alerts_by_rule_id` → `search_findings_by_rule_title` (now filters on `wazuh.rule.title`, not `wazuh.rule.id`): a UUID isn't a human-memorable identifier the way the old 4.x numeric ID was, so a "search by rule ID" premise didn't survive the 5.0 port.
- Removed the `field-policy-coverage.test.ts` retired-literal regression guard (`FORBIDDEN_LEGACY_LITERALS` + its source scan) after review discussion concluded it wasn't worth its maintenance cost once the `wazuh.*` vocabulary is established team knowledge.
- Generalized the numeric-range-on-keyword-field guardrail's rejection message (`guardrails.ts`) from a `Set` to a `Map<field, description>`, so a future second field added to that guard gets an accurate message instead of a hardcoded "severity levels" one.
- Full "alerts" → "findings" terminology sweep per the general review comment: renamed the remaining 9 tool identifiers and files (`get_critical_alerts` → `get_critical_findings`, etc.), the router's `alerts` category → `findings`, and reworded the remaining "alert(s)" references across tool descriptions, `prompts.ts`, `guardrails.ts`, `privacy.ts`, `digest.ts`, the eval harness (`run_plumbing.js`/`run_tool_matrix.js`/`run_load.js`, which still called the retired tool names — a real bug, not just cosmetic), `eval/corpus.json`/`measurement-set.json` case IDs and prompts, docs, the public chat UI, and both translation files. Historical references to the retired 4.x `wazuh-alerts-*`/`wazuh-archives-*` index names, and the original 4.14 UAT intent taxonomy names cited in eval notes, are intentionally left as-is — they're correct historical facts, not current terminology.
- Also fixed a bug this rename exposed: the eval harness's live-fixture discovery (`es_discovery.js`) still aggregated on `wazuh.rule.id` and coerced it to `Number` for the rule-title tool's test fixture — always `NaN` against a UUID. Switched discovery to aggregate on `wazuh.rule.title` instead.
- Two review threads (an `AGG_FIELD_ALLOWLIST`-from-index-mappings suggestion, and a general terminology question) are still open pending reviewer follow-up; see PR comments.

### Results and Evidence

- `yarn test:jest --runInBand`: 543 passed, 0 failed, 39 suites.
- `yarn lint`: clean (one pre-existing, unrelated error in `public/plugin.ts`, untouched by this change).
- `npx tsc --noEmit -p tsconfig.json` (no dedicated `typecheck` script in this plugin): no errors in any file this change touched.
- Field-vocabulary claims verified directly against a live indexer (`wazuh-findings-v5*`/`wazuh-events-v5*`), the Wazuh Common Schema repo's own field docs, and the original plugin PR's maintainer review comment — not just static schema inspection.
- Manually tested the running app end to end: created chat conversations exercising the severity fix ("medium alerts" now returns exactly the medium-only count instead of medium-and-above; "informational alerts" now correctly returns 0 instead of every severity) and confirmed the Settings → Privacy field-policy list reflects the new `wazuh.*` vocabulary and the compliance wildcard on a fresh (no prior saved settings) load.

### Artifacts Affected

- `plugins/wazuh-ai-assistant` only. No changes to other plugins.

### Configuration Changes

None.

### Documentation Updates

- `docs/ref/modules/ai-assistant/tool-catalog.md`: documented the new numeric-range-on-keyword-field guardrail rejection, updated the tool table's "Alerts" category → "Findings" and every renamed tool name.
- `CHANGELOG.md`: folded into the existing "Added Wazuh AI Assistant plugin" entry, linking issue #8802 alongside #8788.
- `eval/README.md`, `eval/measurement-set.README.md`, plugin root `README.md`: updated for the renamed tools/category and reworded remaining "alert(s)" prose.

### Tests Introduced

- `common/wazuh-fields.test.ts` (new field-vocabulary module).
- `guardrails.test.ts`: new tests for the numeric-range-on-keyword rejection.
- `result-table.test.tsx`: new assertion that `informational` and `low` render with genuinely distinct visual attributes, not just different labels.
- `server/tools/catalog/common.test.ts` (new): coverage for `severitiesAtOrAbove`/`severitiesAtOrBelow`/`severityFilterValues`, including the exact-match default, fail-open behavior on an unrecognized severity value, an unrecognized `severity_comparison` value, case-insensitivity, and empty-string input.
- `assistant-settings.test.ts`: coverage for the saved-object type's schema/registration.
- `field-policy-coverage.test.ts`'s retired-literal regression guard (`FORBIDDEN_LEGACY_LITERALS`) was added then removed per review feedback (see above) — its other two tests (privacy-classification coverage) remain.

## Review Checklist

- [x] Code changes reviewed


plugins/wazuh-ai-assistant/common/discover-url.test.ts
```diff
@@ -42,13 +42,13 @@ test('risonEncode: full nested query DSL clause', () => {
   const dsl = {
     bool: {
       filter: [{ range: { timestamp: { gte: 'now-24h', lte: 'now' } } }],
-      must: [{ match: { 'rule.level': 12 } }],
+      must: [{ match: { 'wazuh.rule.level': 12 } }],
```
Desvelao:
```suggestion
      must: [{ match: { 'wazuh.rule.level': 'medium' } }],
```

plugins/wazuh-ai-assistant/common/discover-url.test.ts
```diff
@@ -42,13 +42,13 @@ test('risonEncode: full nested query DSL clause', () => {
...
     },
   };
   const out = risonEncode(dsl);
   assert.equal(
     out,
-    "(bool:(filter:!((range:(timestamp:(gte:'now-24h',lte:'now')))),must:!((match:('rule.level':12)))))",
+    "(bool:(filter:!((range:(timestamp:(gte:'now-24h',lte:'now')))),must:!((match:('wazuh.rule.level':12)))))",
```
Desvelao:
```suggestion
    "(bool:(filter:!((range:(timestamp:(gte:'now-24h',lte:'now')))),must:!((match:('wazuh.rule.level':'medium')))))",
```

plugins/wazuh-ai-assistant/public/components/chat/result-table.tsx
```diff
@@ -30,38 +30,52 @@ const TABLE_SCROLL_MAX_HEIGHT = 400;
...
- * 15+ critical), the Wazuh 4.14 model. In Wazuh 5.0 (findings-v5) rule.level is ALREADY a
- * categorical word (informational/low/medium/high/critical) and never numeric, so this function is
- * now only a fallback for any severity column whose value still arrives numeric; the word path in
- * `renderSeverityBadge` is the normal 5.0 route. Kept so a mixed/older data source still renders.
+ * Numeric-to-word severity classification. `wazuh.rule.level` is a categorical word, never
+ * numeric, so this is only a defensive fallback for a severity value that arrives numeric from
+ * some other source; the string path in `renderSeverityBadge` is the normal route.
  */
-function severityWordFromLevel(level: number): SeverityLevel {
+function legacySeverityWordFromNumericLevel(level: number): SeverityLevel {
```
Desvelao: **question**: what is the purpose of this? there is no level as numbers at 5.0.

plugins/wazuh-ai-assistant/public/components/chat/result-table.tsx
```diff
@@ -74,9 +88,10 @@ function severityWordFromLevel(level: number): SeverityLevel {
...
 }
 
-/** `value` is numeric (or a numeric-looking string) when it came from a rule.level column; a
- * severity column sourced from a field that is already a word (e.g. vulnerability.severity)
- * fails this check and falls through to the pre-existing word-matching path below unchanged. */
+/** `value` is numeric (or a numeric-looking string) when it came from a legacy rule.level column;
+ * a severity column sourced from a field that is already a word (the normal 5.0 case, e.g.
+ * `wazuh.rule.level`/`vulnerability.severity`) fails this check and falls through to the string
+ * path below unchanged. */
 function toSeverityLevelNumber(value: unknown): number | undefined {
```
Desvelao: **thought**: I do not know why is this is required. There is no severity as number in 5.0.

plugins/wazuh-ai-assistant/public/components/chat/result-table.tsx
```diff
@@ -95,26 +110,14 @@ function renderSeverityBadge(value: unknown): React.ReactNode {
...
   const isKnown = (SEVERITY_LEVELS as readonly string[]).includes(word);
-  const color = isKnown ? SEVERITY_COLORS[word as SeverityLevel] : 'default';
-  const label = isKnown
-    ? SEVERITY_LABELS[word as SeverityLevel]
-    : String(value ?? '');
-  return <EuiBadge color={color}>{label}</EuiBadge>;
+  if (!isKnown) {
+    return <EuiBadge color='default'>{String(value ?? '')}</EuiBadge>;
+  }
+  const bucket = SEVERITY_BUCKETS[word as SeverityLevel];
```
Desvelao: **nitpick**: if the `word` variable is not one of SEVERITY_BUCKETS, the access to properties of `bucket` will fail.

plugins/wazuh-ai-assistant/server/tools/catalog/get-alerts-by-time.ts
```diff
@@ -4,38 +4,40 @@ import {
...
+ * is scoped to the critical severity word only). `severity` matches that exact severity word by
+ * default; `severity_comparison` opts into "or above"/"or below" (see common.ts's
+ * severityFilterValues).
  */
 export const getAlertsByTimeTool: ToolDefinition = {
   spec: {
     name: 'get_alerts_by_time',
     description:
       'Searches security findings for ALL findings of any severity within a time range, most ' +
       'recent first. Use for general "show me the alerts"/"what happened in the last N hours" ' +
```
Desvelao:
```suggestion
      'recent first. Use for general "show me the findings"/"what happened in the last N hours" ' +
```

plugins/wazuh-ai-assistant/server/tools/catalog/get-brute-force.ts
```diff
@@ -73,7 +73,13 @@ export const getBruteForceTool: ToolDefinition = {
                     // why an exact `terms` filter is used rather than a description match).
                     {
                       terms: {
-                        'rule.id': ['5710', '5712', '5716', '5720', '5760'],
+                        'wazuh.rule.id': [
+                          '5710',
+                          '5712',
+                          '5716',
+                          '5720',
+                          '5760',
```
Desvelao: **nitpick**: these IDs are from 4.x, in 5.x they are a UUID. We could consider using other fields to filter.

plugins/wazuh-ai-assistant/server/tools/catalog/search-alerts-by-rule-id.ts
```diff
@@ -58,7 +57,7 @@ export const searchAlertsByRuleIdTool: ToolDefinition = {
         query: {
           bool: {
             filter: [
-              { term: { 'rule.id': String(ruleId) } },
+              { term: { 'wazuh.rule.id': String(ruleId) } },
```
Desvelao: **nitpick**: the `wazuh.rule.id` field is a UUID, I am not sure if this make sense instead of another field as `wazuh.rule.title` or `wazuh.rule.tags`.

plugins/wazuh-ai-assistant/server/tools/field-policy-coverage.test.ts
```diff
@@ -190,3 +201,91 @@ test('isFieldCovered mechanism: an unclassified field is correctly flagged as NO
     true,
   );
 });
+
+/**
+ * Bare field literals this test denies — not part of the plugin's vocabulary (see
+ * `common/wazuh-fields.ts`'s `WAZUH_FIELD`), kept here only as this regression guard's own
+ * scan target.
+ */
+const FORBIDDEN_LEGACY_LITERALS = [
```
Desvelao: **question**: what is the purpose of this? There are some fields that seem to be related to 4.x.

plugins/wazuh-ai-assistant/server/tools/guardrails.ts
```diff
@@ -279,6 +273,61 @@ const VULN_FIELD_CLAUSE_KEYS = new Set([
...
+  let reason: string | undefined;
+  walk(body, (key, value) => {
+    if (reason || key !== 'range' || !value || typeof value !== 'object') {
+      return;
+    }
+    for (const [field, rangeValue] of Object.entries(
+      value as Record<string, unknown>,
+    )) {
+      if (
+        !KEYWORD_RANGE_REJECT_FIELDS.has(field) ||
```
Desvelao: **suggestion**: this comes from a hardcoded field, this forces the maintenance if new fields are added or chagned. If this use the indices mappings to be generated, this approach avoid the maintenance.

plugins/wazuh-ai-assistant/server/tools/guardrails.ts
```diff
@@ -279,6 +273,61 @@ const VULN_FIELD_CLAUSE_KEYS = new Set([
...
+        bounds.lte,
+        bounds.lt,
+      ].some(bound => typeof bound === 'number');
+      if (hasNumericBound) {
+        reason =
+          `Range on "${field}" is not allowed: it is a keyword field holding one of the string ` +
+          `severity levels (${SEVERITY_LEVELS.join(
+            '/',
+          )}), not a number. Use "term" or "terms" ` +
+          `against one or more of those string values instead.`;
```
Desvelao: **nitpick**: this seems to be "hardcoded" to the expected case of `wazuh.rule.level`. It is not reusable.

Desvelao:
**suggestion**: consider replacing/remove the references to "alerts". In 5.x there are findings, events and states.
image

plugins/wazuh-ai-assistant/public/components/chat/result-table.tsx
Ripdiegozz: Solved at: https://github.com/wazuh/wazuh-dashboard-plugins/pull/8812/commits/66ac67bcffc5d2217180355b2e4258a1e5a82e3d

plugins/wazuh-ai-assistant/public/components/chat/result-table.tsx
Ripdiegozz: Solved at: https://github.com/wazuh/wazuh-dashboard-plugins/pull/8812/commits/416606a8799c1fdfa5402f64aa0fbfe556910449

plugins/wazuh-ai-assistant/server/tools/catalog/get-brute-force.ts
Ripdiegozz: Solved at: https://github.com/wazuh/wazuh-dashboard-plugins/commit/774b044b9e7ea8357558b9baea4c6c43793f2187

plugins/wazuh-ai-assistant/server/tools/catalog/search-alerts-by-rule-id.ts
Ripdiegozz: Solved at: https://github.com/wazuh/wazuh-dashboard-plugins/commit/f7d6f6b8995dff306831b431bf0897d8d618632c

plugins/wazuh-ai-assistant/server/tools/field-policy-coverage.test.ts
Ripdiegozz: Removed in b2848c9

plugins/wazuh-ai-assistant/server/tools/guardrails.ts
Ripdiegozz: Solved at: https://github.com/wazuh/wazuh-dashboard-plugins/commit/e3a8f3747cfcce32da7421f04ebb1f714e59f681

plugins/wazuh-ai-assistant/public/components/chat/result-table.tsx
Ripdiegozz: Removed at: https://github.com/wazuh/wazuh-dashboard-plugins/commit/66ac67bcf

plugins/wazuh-ai-assistant/server/tools/guardrails.ts
Ripdiegozz: This hand-curated allowlist already existed on 5.0.0 (just with several broken 4.x-style names like rule.id/agent.name that don't match the real 5.0 schema; this PR fixed the names, not the approach.

On generating it from the index mapping: I ran a POC against the live mapping and it surfaces 383 wazuh.* keyword fields, 375 of them not in today's list, including UUIDs and file hashes that are exactly the high-cardinality fields this allowlist exists to keep out of terms aggs. Mapping type alone (keyword) doesn't tell us cardinality, so a straight generate-from-mapping would widen this into an unsafe state.

I think this deserves a proper design on its own, how to fetch/cache mappings safely, and how to vet cardinality without a manual per-field call, rather than a quick fix bundled into this PR.

yenienserrano (approved):
LGTM :green_circle:

PR #8822, opened 2026-07-30
Change AI assistant's field privacy policy

### LINKED ISSUE #8821: Improvements to the AI Assistant's Privacy Policy
| Wazuh    | Rev  | Browser                      |
|----------|------|------------------------------|
| 5.0.0-00 | xyzw | Chrome                       |

**Description**

The AI assistant privacy settings let the user assign one of three actions per field
(`plugins/wazuh-ai-assistant/public/components/settings/settings-page.tsx`):

- **Allow** — the field is sent to the model as-is.
- **Anonymize** — the field is pseudonymized before being sent.
- **Never send** — the field must not be sent at all.

The chat answer itself honors these actions correctly: the field policy is applied to the
*digest* (the payload the model sees) via `applyFieldPolicy` in
[privacy.ts](plugins/wazuh-ai-assistant/server/tools/privacy.ts), so anonymized fields arrive
pseudonymized and `never` fields are dropped.

However, the policy is **not** applied to two other outputs of the same tool call:

1. **The results table rendered in the chat.** In
   [executor.ts](plugins/wazuh-ai-assistant/server/tools/executor.ts#L166-L170) the `table` event
   is built from the raw indexer response and deliberately bypasses the policy pass
   ("it renders locally in the browser and never reaches the model"). As a result, a field set to
   **Anonymize** or **Never send** is still displayed as a plain, clear-text column in the table.
2. **The query that is executed.** The DSL body is built and sent to the indexer *before* the
   policy pass runs (`buildDigest` → `finalizeDigest` happen on the response). The `_source` list
   still requests fields marked `anonymize`/`never`, so those values are always retrieved from the
   indexer, and they are also attached to the table's `discover` payload
   (`tableSpec.discover.dsl`) for the "Open in Discover" action.

This makes the privacy setting misleading: the user configures a field as **Never send** and still
sees its raw value in the UI, and the field is still read out of the index on every tool call.

**Preconditions**

1. A Wazuh dashboard with the AI assistant plugin enabled and a configured model provider.
2. Data available in the alerts/findings indices so a tool call returns rows (e.g. `wazuh.agent.name`
   populated).
3. Privacy mode enabled for the chat session.

**Steps to reproduce**

1. Navigate to the AI assistant **Settings** page → **Privacy** section.
2. Set a visible field (for example `wazuh.agent.name`) to **Never send**. Save the settings.
3. Open the AI assistant chat with privacy mode enabled.
4. Ask a question that triggers an indexer tool call returning that field, e.g.
   "show me the last findings by agent".
5. Inspect the assistant's textual answer, then inspect the results table rendered below it.
6. Repeat with the field set to **Anonymize**.

**Expected Result**

1. The textual answer omits the field (**Never send**) or shows only its pseudonym (**Anonymize**).
2. The results table respects the same policy: the column is absent for **Never send** and shows the
   pseudonymized value for **Anonymize**.
3. The executed query does not request fields marked **Never send** (they are excluded from
   `_source`), and the DSL attached to "Open in Discover" does not reference them either.

**Actual Result**

1. The textual answer is correct — the policy is applied to the digest.
2. The results table shows the field as a clear-text column regardless of whether it is set to
   **Anonymize** or **Never send**.
3. The query is built and executed before the policy pass, so the field is always retrieved from the
   indexer and is included in the `discover` DSL.

**Screenshots**

_To be added: chat answer with the field anonymized/omitted next to the results table still showing
the clear-text column._

**Additional context**

Relevant code paths:

- [privacy.ts](plugins/wazuh-ai-assistant/server/tools/privacy.ts) — `FieldPolicyAction` is
  `'allow' | 'anonymize' | 'never'`; `applyFieldPolicy` is the only place the policy is enforced,
  and it operates on the digest.
- [executor.ts:150-170](plugins/wazuh-ai-assistant/server/tools/executor.ts#L150-L170) —
  `finalizeDigest` applies the policy to the digest, while `buildTableSpec(result, def, body)` builds
  the table from the raw response and explicitly does not.

Two decisions are needed as part of the fix:

- Whether the field policy is meant to be a **model-boundary** control only (in which case the UI
  wording must make that explicit, e.g. "not sent to the model") or a **display + retrieval** control
  (in which case the table and the `_source` projection must both honor it).
- If it is the latter, the policy must be resolved **before** the request body is built so `never`
  fields can be excluded from `_source` and from the `discover` DSL, rather than filtered out of the
  response afterwards.

felipegonzalezmv:
## Description

Information is added about the different values that can be entered in the fields

Never send 
anonymous 
allow

Closes #8821

## Review Checklist

- [ ] Code changes reviewed


rodrigofez (approved):
🟢 Test

Image

yenienserrano (approved):
LGTM :green_circle:

PR #8823, opened 2026-07-30
Fix AI Assistant view routing on page refresh#

### LINKED ISSUE #8820: Add routing the AI assistant plugin
| X.Y.Z | xyzw | Chrome, Firefox, Safari, etc |

**Description**
The AI assistant does not manage the view routing, refresing the page goes to chat tab.

Image

jcorredor-spec:
## Description

The AI Assistant app did not manage view routing: the active tab (Chat/Settings) was kept only in local React state, so refreshing the page always returned to the Chat tab, and browser back/forward navigation had no effect on the view. Deep links to the Settings view were also not possible.

## Proposed Changes

- Introduced a `react-router-dom` `Router`/`Route`/`Redirect` setup (backed by a `history` created with `createHashHistory`) as the single source of truth for the active tab: `/` ↔ Chat and `/settings` ↔ Settings. A path-less top-level `<Route>` redirects (via `replace`, not `push`) any path other than `/`, `/settings`, or `/conversation/*` back to `/`, so a stale/mistyped/removed deep link doesn't linger in the address bar or in browser history.
- The Settings route is matched with `<Route exact path="/settings">` using a `children` render function, which runs on every render regardless of match — this is what keeps both tabs mounted (never unmounted) across a tab switch, preserving `ChatPage`'s in-memory state (transcript, active conversation id, pseudonym map, tool-call history). `settingsEverOpened` is now tracked with a ref instead of state, flipped directly from the route's render.
- The open-conversation address moved from a hash fragment (`#/conversation/<id>`) to a real path segment (`/conversation/<id>`, `common/conversation-location.ts`). `ChatPage` now receives the app shell's router `history` as a prop and reads/writes the conversation route through `history.location.pathname` / `history.replace` instead of raw `window.location.hash` / `window.history.replaceState`, so both the tab route and the conversation route go through one writer of location changes.
- The conversation route is only written while Chat is the visible tab (`isActive` prop on `ChatPage`), so a background restore no longer rewrites `/settings` into `/settings/conversation/<id>`; it re-syncs when Chat becomes active again.

### Results and Evidence

https://github.com/user-attachments/assets/42f5d35e-769f-4015-acd1-205ad1f41463

### Artifacts Affected

- `wazuh-ai-assistant` plugin (public bundle only). No server-side changes, executables, or packages affected.

### Tests Introduced

- `plugins/wazuh-ai-assistant/public/application.test.tsx`: `view routing (#8820)` suite mounts the app by seeding `window.location.hash` before mount (the router now owns its own hash history internally rather than taking one as a prop) and asserts against the resulting hash/path, including a new case for an unknown deep link redirecting back to Chat.
- `plugins/wazuh-ai-assistant/public/components/chat/chat-page.test.tsx`: conversation-route tests now seed/assert via `window.history.replaceState` / `window.location.pathname` (a real path) instead of `window.location.hash`, and pass a `history` prop (`createBrowserHistory()`) into `ChatPage`.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


Desvelao:
# Changes

- Replace routing using ReactJS Router, Route and Redirect components

Desvelao (approved):

PR #8835, opened 2026-08-03
Improve provider form UX and error messages

### LINKED ISSUE #8829: Enhancements in AI assistant
**Description**

- [x] Change the placeholder of endpoint URL when changing the provider type. This should display the placehodler with the valid URL for OpenAI or Anthropic
- [x] Add example under the URL endpoint input with the URLs OpenAI and Anthropic, this should display according to the provider type. Add the link to the documentation of these services.
- [x] Enhance error messages in settings.
> Depending on the case we have seen HTML content or JSON. Try to capture some cases, else display all.
- [x] Expand the panel width of the settings page.

Desvelao:
## Description

The AI Assistant provider settings form (the "Add/Edit provider" flyout) had a
few rough edges reported in the linked issue: the endpoint URL field always
showed an OpenAI-shaped placeholder regardless of the selected provider, there
was no example/documentation to help the admin fill in the URL or model name
correctly, raw provider/gateway error bodies (which can be JSON or HTML
depending on what rejected the request) were shown to the user unparsed, and
the flyout was narrow enough to make the form cramped.

Closes #8829

## Proposed Changes

- The **Endpoint URL** field's placeholder and inline examples now switch
  based on the selected **Provider type** (OpenAI-compatible vs Anthropic),
  instead of always showing the OpenAI format.
- The **Model** field (free text, no enum) gets the same treatment: per-type
  example model names (e.g. `gpt-4o-mini`, `llama-3.3-70b-versatile` for
  OpenAI-compatible; `claude-sonnet-4-5` for Anthropic).
- Both fields expose their documentation links behind a single **popover**
  trigger ("API documentation" / "See available models") instead of inlining
  every link in the help text — for `openai_compatible` this lists one link
  per actual service the type covers (OpenAI, Groq, Ollama), rather than a
  single misleading "OpenAI" link. A note in that popover points admins using
  another OpenAI-compatible provider or gateway (Gemini, LM Studio, vLLM,
  etc.) at that provider's own documentation. The popover panel width is
  capped so the note wraps instead of stretching the whole panel/trigger.
- Provider/gateway HTTP error bodies are now parsed for two well-known shapes
  before falling back to the raw (truncated, redacted) body: the
  `{"error": {"message": "..."}}` JSON contract used by OpenAI, Anthropic, Groq
  and OpenAI-compatible gateways (vLLM, Ollama, LM Studio...), and an HTML error
  page's `<title>` (or tag-stripped text), which is common when a reverse proxy
  or load balancer rejects the request before it reaches the provider. Any
  other shape still falls through to the previous raw-body behavior — "capture
  some cases, else display all". This applies to both the streaming chat path
  (`server/providers/retry.ts`) and the hosted-brain webhook adapter
  (`server/providers/wazuh-brain.ts`).
- Removed the settings page's `restrictWidth={960}` cap on `EuiPage`, so the
  page (provider list, privacy policy, etc.) uses the full available width
  instead of being capped at 960px.
- Clarified the **API key** field's help text: it's optional for endpoints
  that don't require authentication (e.g. a local Ollama server without
  auth) and is stored encrypted at rest when an encryption key is
  configured; edit mode keeps the existing "leave empty to keep the
  current key" note.

### Results and Evidence

image
image
image
image

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (public UI + server provider adapters). No
  executables, packaging, or default configuration files affected.

### Configuration Changes

None.

### Documentation Updates

None required — no public API or configuration surface changed.

### Tests Introduced

- `provider-form-flyout.test.tsx`:
  - Endpoint URL placeholder/example switch, and the documentation popover
    (OpenAI/Groq/Ollama links, the other-providers note) for both
    `openai_compatible` and `anthropic`.
  - Model field example switch, and its own documentation popover
    (OpenAI/Groq/Ollama model-list links, the other-providers note) for both
    provider types.
  - API key field's clarified help text in create mode (optional /
    encrypted-at-rest wording).
- `retry.test.ts`: new `extractProviderErrorMessage` coverage — unwraps the
  `{error:{message}}` JSON shape, extracts an HTML `<title>` (and decodes
  entities), falls back to tag-stripped text when no `<title>` is present,
  passes through plain text/unrecognized JSON/malformed JSON unchanged, and
  confirms an unwrapped JSON message still goes through
  `sanitizeProviderErrorBody`'s secret redaction. Also adds a regression test
  proving a bare `{error: "..."}` / `{message: "..."}` body is deliberately
  **not** unwrapped, since a sibling field in that shape could carry something
  that still needs redaction (this guards the existing
  `wazuh-brain.test.ts` API-key-echo redaction test, which exercises exactly
  that ambiguous shape).

## Review Checklist

- [x] Code changes reviewed
- [x] Configuration changes documented (none needed)
- [x] Developer documentation reflects the changes (none needed)
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,7 @@
 - Added `wazuh.disabledSettings` configuration to hide specific settings in the Indexer Settings UI [#8643](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8643) [8693](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8693)
 - Expanded the case management form with title, description, severity, priority and TLP fields, comments that can be added or edited individually, and a confirmation dialog before discarding unsaved changes [#8718](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8718)
 - Added queue usage in bytes and agent cache visualizations to `Server Management` > `Statistics`, and relabeled the Comms "Queue usage" Y-axis to Bytes [#8768](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8768)
-- Added Wazuh AI Assistant plugin: provider-agnostic AI chat over Wazuh data with a read-only tool catalog, server-side guardrails, optional pseudonymization, and API-key encryption at rest [#8788](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8788) [#8797](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797)
+- Added Wazuh AI Assistant plugin: provider-agnostic AI chat over Wazuh data with a read-only tool catalog, server-side guardrails, optional pseudonymization, and API-key encryption at rest [#8788](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8788) [#8797](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8797) [#8829](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8829)
 
 ### Changed
 
--- a/plugins/wazuh-ai-assistant/public/components/settings/provider-form-flyout.test.tsx
+++ b/plugins/wazuh-ai-assistant/public/components/settings/provider-form-flyout.test.tsx
@@ -246,3 +246,33 @@ describe('ProviderFormFlyout — API key encryption gate', () => {
     expect(screen.getByRole('button', { name: /^save$/i })).toBeEnabled();
   });
 });
+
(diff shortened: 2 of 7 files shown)


yenienserrano (dismissed):
LGTM :green_circle:

Image
Image

Image
Image

Image

asteriscos (approved):
CR: 🟢 
Test: 🟢 

### Error handling

Image

### Form placehoder and documentation

Image

Image

Issue #8841, opened 2026-08-04
Implement AI assistant session history and provider indexes


asteriscos:
## Description

The AI assistant session history and provider configurations will be migrated to independent indexes. We need to make the following changes to implement the new source of truth.

Related indexer issue:
- https://github.com/wazuh/wazuh-indexer-plugins/issues/1422
 
### Tasks

- [x] Retrieve the session data from the index alias
- [x] Change the session history retention period using the ISM of the `wazuh-ai-assistant-sessions` index
  - API reference: https://docs.opensearch.org/latest/im-plugin/ism/api/
  - The value is obtained from the ISM policy
  - Updating the setting should update the ISM policy and the existent indices
- [x] Adapt the AI provider and privacy settings CRUD to implement the API endpoints provided by Wazuh indexer.
  - [x] Use the current user to retrieve the settings (the retention ISM policy is obtained through the template)
  - [x] Create or update providers/settings should be executed using the current user using the AI assistant plugin endpoints from Wazuh indexer
- [x] Define the minimal permissions to use the AI Assistant
- [x] Remove the requirement of `administrator` role of manager API to manage the settings
- [x] Add docs

Desvelao:
# Research

The internal user could need permissions to retrieve the data of the ISM template related to `wazuh-ai-assistant-sessions` indices
- `cluster:admin/opendistro/ism/policy/get`

Desvelao:
# Changes

- Adapt to Wazuh indexer API spec: settings and providers
- Add callout related to conversation historial settings
  - Unify message for generic and conversation historial settings

Desvelao:
Waiting for package of Wazuh indexer with the API endpoints definitions to validate the changes in dashboard.

Desvelao:
# Permissions - Use cases

Related to: https://github.com/wazuh/wazuh-indexer/pull/1795/changes

## Minimal permissions to access to dashboard

| Permission | Operation | Description |
| --- | --- | --- |
| `cluster_composite_ops_ro` | READ | Read minimal information, this allows to access and navigate at least to the AI assistant |

## AI chat

| Permission | Operation | Description |
| --- | --- | --- |
[...]

PR #8844, opened 2026-08-05
Move conversation retention to a scheduled job

### LINKED ISSUE #8830: The AI assistant conversation retention policy is managed in a GET endpoint
**Description**

The AI assistant conversation retention policy is managed in a GET endpoint using write operations, if the endpoint is never accessed, the conversatiosn are not deleted.

Consider using a background cron-job to ensure the retention policy.

This is affected by the index where the conversations data is stored.

jcorredor-spec:
## Description

Conversation retention was enforced only on-access: `GET /conversations` filtered and best-effort deleted expired rows, so conversations past the retention window persisted indefinitely if never listed, and a read endpoint performed writes.

Closes #8830

## Proposed Changes

- Added `server/conversation-retention.ts`: a background job that runs at plugin start and every hour, deleting conversations (across all owners) older than `conversationRetentionDays`. Deletes run in batches of 100; overlapping passes are skipped; a pass that makes no progress stops and retries on the next tick; delete 404s count as already deleted.
- The job uses the internal saved-objects repository (no request scope) and is stopped on plugin `stop()`.
- `GET /conversations` is now read-only: it still hides expired rows between passes but no longer deletes them.
- Extracted `retentionCutoffMs()` so the job and the list route share the same cutoff arithmetic.
- Updated the retention help text in the settings page to reflect the background cleanup.

### Results and Evidence

- Verified E2E in the `osd-dev` environment: with a retention window set, expired conversations are removed by the job and the server logs `wazuhAiAssistant: retention pruned N expired conversation(s)`.

https://github.com/user-attachments/assets/211c785a-2530-4088-8df2-cef84482bf61

Note : Just for local test  setup ``` export const CONVERSATION_RETENTION_INTERVAL_MS = 60 * 1000;   ```

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (server and public bundles). No packages or executables.

### Configuration Changes

- None. The existing `conversationRetentionDays` setting keeps its default (`0` = keep forever) and meaning; only the enforcement mechanism changed. Backward compatible.

### Documentation Updates

- None (inline doc comments and the settings help text were updated in this PR).

### Tests Introduced

- `server/conversation-retention.test.ts` (13 unit tests): cutoff arithmetic and clamping; prune pass behavior (keep-forever default, missing/legacy settings, batch deletion until empty, 404 and failure handling); scheduler behavior (startup run, hourly re-run, `stop()`, error recovery, overlap skip).

## Review Checklist

- [x] Code changes reviewed


asteriscos:
Closed due to requirement changes. See:
- https://github.com/wazuh/wazuh-dashboard-plugins/issues/8841

PR #8880, opened 2026-08-06
Fix blank answers, tool_choice, and temperature in AI Assistant

### LINKED ISSUE #8862: Assistant returns a blank answer when model output arrives on the reasoning channel
## Description

The assistant sometimes returns a **completely blank answer**. The model generates a full response,
the tokens are billed, and the user sees nothing.

The cause is that `openai/gpt-oss-120b` streams its output on a separate **reasoning channel**, and the
OpenAI-compatible adapter reads only `choice.delta.content`. Everything on `delta.reasoning` is
discarded. The turn terminates normally with a `done` event, so nothing surfaces as an error — the
answer is simply empty.

### Evidence

Captured from the raw upstream response, verbatim:

```
data: {"choices":[{"index":0,"delta":{"reasoning":"The","channel":"analysis"},...
data: {"choices":[{"index":0,"delta":{"reasoning":" user","channel":"analysis"},...
data: {"choices":[{"index":0,"delta":{"reasoning":" asks","channel":"analysis"},...
```

Counted across one captured response: **42 `"reasoning":` deltas against 6 `"content":` deltas.**

Three runs of the same question (*"What should I look at first this morning?"*):

| Run | Output tokens billed | Characters shown to the user |
|---|---|---|
| 1 | 136 | 206 |
| 2 | — | 0 (failed — see the `tool_choice` issue) |
| 3 | **636** | **0** |

**636 output tokens generated and paid for; the user saw an empty reply.** Two of three runs on this
question produced nothing usable.

Confirmed **not** an artifact of the test harness: it reproduces with no proxy in the request path
(direct capture, run 2 of 2, `answerLen: 0`, `outputTokens: 439`).

**Not confined to one model family.** `qwen/qwen3.6-27b` also produced a zero-length answer with a
clean `done` and no error (question L15). Qwen 3.x is likewise a reasoning model, so the most likely
cause is the same. This matters for the fix decision: it cannot be avoided by choosing a different
model, so the adapter is the right place to fix it.

**Why the main test matrix missed it.** The blank answer occurs when the stage-1 router picks
`general` (no Wazuh data needed) and stage 2 therefore runs without tools. Every question in the main
matrix routed to a data category, so the model emitted normal `content`. The `general` path is what a
user hits when they open the assistant and type something conversational or vague — so the real-world
rate is likely higher than a data-heavy test suite suggests.

**There is already a no-text fallback, and it does not cover this path.** `chat.ts` (~L957-963) handles
the case where the model produced no text at all:

```ts
if (!sawAnyDelta && toolUsedThisTurn) {
  yield { type: 'delta',
          content: sawNonEmptyTable ? NO_ANALYSIS_TEXT_MESSAGE : NO_MATCHING_RESULTS_MESSAGE };
}
```

The guard requires `toolUsedThisTurn`. On a `general`-routed turn **no tool runs**, so the fallback never
fires and the user gets a genuinely empty bubble. The product already knows this failure mode is
possible and already has copy for it — the guard is simply too narrow.

### Code references

- `plugins/wazuh-ai-assistant/server/providers/openai-compatible.ts` — the delta parse; it reads
  `choice?.delta?.content` only, and the `parsed` type declaration does not include `reasoning`
- `plugins/wazuh-ai-assistant/server/routes/chat.ts` (~L957-963) — the existing `!sawAnyDelta`
  fallback, gated on `toolUsedThisTurn`

### Suggested treatment

Two changes, and the second is a cheap safety net worth having regardless.

**1. Read `delta.reasoning` in addition to `delta.content`.**

**Recommended:** treat `content` as the answer and use `reasoning` as a fallback **only when no
`content` arrived at all for the turn**, so the user is never shown a blank reply. Groq's chunk carries
`channel` (`"analysis"` versus `"final"`), which distinguishes intermediate reasoning from the answer
if the two should be presented differently.

Alternatives, briefly: render reasoning in a collapsible "thinking" disclosure (more product surface,
more tokens shown to the user); or request that the provider suppress the reasoning channel (not
available on all providers, and it would discard useful content).

The fallback approach is the minimum that guarantees the user never sees an empty answer, which is the
defect being fixed.

**2. Widen the existing `!sawAnyDelta` guard so it also covers the no-tool path.** Dropping
`&& toolUsedThisTurn` (with copy appropriate to a no-tool turn) means that even if a future provider
invents a third output channel nobody has read yet, the user gets a sentence rather than an empty
bubble. This is defence in depth, it reuses copy that already exists, and it is a one-line change.

Do both. Change 1 fixes the known cause; change 2 makes the *symptom* — a blank answer — structurally
impossible.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: **N=10** runs of a `general`-routed question, blank-answer
      count stated (previously observed 2 of 3 on one question)
- [ ] After the fix, a turn whose model output arrives only on `delta.reasoning` renders non-empty text
- [ ] **N>=10** post-fix runs of the same question, blank-answer count stated — target zero
- [ ] Verified on **both** `openai/gpt-oss-120b` and `qwen/qwen3.6-27b`, since both exhibited it
- [ ] A turn that emits normal `content` is unchanged — reasoning must not be appended to a working
      answer (regression)
- [ ] Colocated unit test in `server/providers/openai-compatible.test.ts` with a reasoning-only stream
      fixture and a content-only fixture
- [ ] Regression gate: 16-question set passes; **zero blank answers**

### Notes

- **Intermittent — roughly 2 of 3 runs on the affected question.** A single passing run after the fix
  proves nothing; the N>=10 requirement is not optional here.
- Touches `openai-compatible.ts`, shared with the `tool_choice` issue, the usage-accumulation issue and
  the prompt-caching issue. Land in that order: this one and `tool_choice` first (both urgent), the
  other two after.
- `anthropic.ts` has the same delta-parsing shape and should be checked for the equivalent gap.
- Builds on #8831 (tools revamp), which introduced the existing `!sawAnyDelta && toolUsedThisTurn`
  fallback quoted above — this issue targets the gap that fallback leaves open, not a regression of it.
- Fixing the `general` routing path (never send zero tools) reduces how often this path is entered but
  does **not** fix this defect — the two are independent and both are needed.

### Update — 2026-08-06 validation gate (Amazon Bedrock, `openai.gpt-oss-120b`)

A candidate fix implementing both suggested changes was measured live:

- **Blank bubbles: 0 of 27 records** across the full gate run (vs 3 of 10 on the pre-fix
  baseline for the affected question).
- **The underlying model silence is unchanged: 3 of 10 runs both before and after** on the same
  question. On those runs the model streamed zero deltas for the entire turn
  (`ms_to_first_delta == latency_ms` exactly) and the widened `!sawAnyDelta` fallback supplied
  the text. The fix removes the *symptom* (empty bubble), as designed; it does not reduce how
  often the model emits nothing.
- One observed hazard of the fallback copy: a turn made 4 tool calls that all **correctly**
  returned 0 rows (the queried time windows genuinely held no findings), the model then streamed
  nothing, and the fallback printed "No matching results were found for that query." — accurate
  here, but only coincidentally; on a turn whose tools *did* return rows, the same fallback text
  would be wrong. Fallback copy should not assert facts about the query result.
- Caveat: baseline was captured on Groq, post-fix on Bedrock (same gpt-oss-120b weights,
  different serving stack), ~17 commits apart. Whether the `delta.reasoning` readback path
  specifically fired was not isolated in this run; the zero-blank result is the fallback and
  readback acting together.

### LINKED ISSUE #8863: `tool_choice: 'none'` on the final tool round hard-fails with gpt-oss models
## Description

On the last permitted tool round the orchestration loop sends `tool_choice: 'none'` to stop the model
looping on tools forever. Some models ignore that setting and attempt a tool call anyway, and the
provider then rejects the whole request. The user gets a failed turn.

The intent is right — the loop does need a terminator. The mechanism depends on the model complying
with a hint, and the models this product is about to recommend do not reliably comply.

### Evidence

Observed as the terminal error of a chat turn on `openai/gpt-oss-120b`, verbatim:

> `Tool choice is none, but model called a tool`

Rate on the affected question: **1 of 3 runs.**

The only place the plugin sends that value is the final-round branch of the orchestration loop in
`chat.ts`:

```ts
const streamOptions = isFinalRound ? { tools, toolChoice: 'none' as const } : ...
```

Groq documents this exact hazard for `tool_choice: "none"` (verified 2026-08-05,
`console.groq.com/docs/tool-use/local-tool-calling`):

> "With some models, the model may still attempt to use tools despite this setting - if this happens,
> the API will return an error (400 Bad Request) since tool execution was blocked. This behavior varies
> by model."

So the failure is documented upstream behaviour, not a surprise — the setting is simply not a reliable
terminator across models.

### Code references

- `plugins/wazuh-ai-assistant/server/routes/chat.ts` — the `isFinalRound` branch that sets
  `toolChoice: 'none'`
- `plugins/wazuh-ai-assistant/server/providers/openai-compatible.ts` — `toOpenAiToolChoice`, and the
  `if (options?.tools?.length)` guard that decides whether `tool_choice` is sent at all

### Suggested treatment

**Recommended: do not send `tool_choice: 'none'`. Omit `tools` entirely on the final round.**

With no tools offered there is nothing for the model to call, no compliance to depend on, and no 400 to
trigger. The model is forced to answer in prose — the same outcome the current code is reaching for,
achieved structurally instead of by request. The adapter already only sets `tool_choice` when
`options.tools` is non-empty, so omitting `tools` removes the offending parameter for free.

Alternative considered: keep `tool_choice: 'none'` and catch the resulting 400 as a retry-with-no-tools.
That works but costs an extra round trip on every occurrence and leaves the fragile path in place.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: **N=10** runs of a question that reaches the final tool
      round, failure count for `Tool choice is none, but model called a tool` stated
- [ ] After the fix, the final round is sent with no `tools` array and no `tool_choice`
- [ ] **N>=10** post-fix runs, failure count stated — target zero
- [ ] The loop still terminates: a question that would otherwise keep calling tools produces a prose
      answer at the round limit rather than looping (regression)
- [ ] Colocated unit test asserting the final-round request body contains neither `tools` nor
      `tool_choice`
- [ ] Regression gate: 16-question set passes; zero raw upstream error text

### Notes

- **Intermittent — 1 of 3 runs observed.** N>=10 before and after is required to claim a fix.
- Touches `chat.ts` and `openai-compatible.ts`. `openai-compatible.ts` is shared with the
  `delta.reasoning` issue — land these two together or sequence them; they are the two urgent ones in
  that file.
- Worth checking whether `MAX_TOOL_ROUNDS` is reached often enough in practice to matter beyond this
  bug; if the round limit is rarely hit, the blast radius is smaller than the 1-in-3 rate suggests.

### Update — 2026-08-06 validation gate (Amazon Bedrock, `openai.gpt-oss-120b`)

With a candidate fix (omit `tools` entirely on the final round) in the build under test:
**0 occurrences of `Tool choice is none, but model called a tool` across 22 live records**
(10 + 5 runs of the originally-affected question, 6 regression questions, 1 usage probe), vs
**2 of 10** on the pre-fix baseline of the affected question. Zero raw upstream error text of any
kind reached a user-visible answer in the same 22 records. Caveat: baseline Groq, post-fix
Bedrock (same weights, different serving stack), multi-commit delta.

### LINKED ISSUE #8865: No `temperature` is sent, so tool calls run at the provider default of 1.0
## Description

The plugin never sends a `temperature`, so every request — including the structured tool calls the whole
feature depends on — runs at the provider's default. On Groq that default is **1.0**, which is the value
Groq's own tool-use documentation advises against for tool calling, and lowering it is the mitigation
they prescribe for malformed tool calls.

This is the most likely contributing cause of the intermittent "the model produced an invalid tool call"
failures reported against this plugin.

### Evidence

`temperature` appears **nowhere** in the plugin — zero matches across the whole plugin tree. The
outbound request body is built explicitly and contains only `model`, `stream`, `messages`, and (when
tools are offered) `tools`, `tool_choice` and `parallel_tool_calls`:

```ts
const body: Record<string, unknown> = {
  model: config.model,
  stream: true,
  messages: messages.map(toOpenAiMessage),
};
if (options?.tools?.length) {
  body.tools = toOpenAiTools(options.tools);
  body.tool_choice = toOpenAiToolChoice(options.toolChoice ?? 'auto');
  body.parallel_tool_calls = false;
}
```

With the field absent, Groq applies its default of 1.0.

Groq's guidance (verified 2026-08-05, `console.groq.com/docs/tool-use/local-tool-calling`):

- recommends **0.0–0.5** for tool-calling workloads;
- prescribes, as the remedy for an invalid tool call, retrying with **progressively lower temperature** —
  starting near 1.0 and decreasing by 0.2 to a floor of 0.2 — with the stated rationale that lowering it
  "reduce[s] hallucinations, which is critical for structured outputs like tool calls".

So the plugin currently runs structured tool calling at the exact setting the provider warns about, and
the documented remedy is not implemented.

Supporting observation: the malformed-tool-call failure was reproduced in this QA pass on
`qwen/qwen3.6-27b` (`Failed to call a function. Please adjust your prompt.`) on a question that
`openai/gpt-oss-120b` answered correctly — and community reports describe it as **non-deterministic on
the same prompt and model**, which is the signature of a sampling problem rather than a prompt problem.

Note this is **contributing, not proven causal**: no controlled before/after temperature comparison has
been run yet. That comparison is the first acceptance criterion below.

### Code references

- `plugins/wazuh-ai-assistant/server/providers/openai-compatible.ts` — the request-body construction
- `plugins/wazuh-ai-assistant/server/providers/retry.ts` — already owns retry policy, including a
  bounded `tool_use_failed` retry with `TOOL_USE_FAILED_MAX_RETRIES`; the natural home for a
  lower-temperature retry ladder

### Suggested treatment

**Recommended: send a low `temperature` on tool-bearing requests.** 0 for the stage-1 router call, whose
only job is to emit one structured pick from a fixed enum, and 0–0.2 for stage 2.

Optionally, and separately: extend the existing `tool_use_failed` retry in `retry.ts` to retry at a lower
temperature, matching Groq's documented ladder. The retry mechanism and its message already exist, so
this is an addition to a working path rather than new machinery.

Consider whether `temperature` should also be exposed as a provider setting. Recommendation: not yet —
a sensible fixed default fixes the observed problem, and a user-tunable sampling parameter on a
structured-output path is a support burden.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: **N=10** runs of a tool-heavy question at the current
      implicit temperature, stating the malformed-tool-call rate
- [ ] After the fix, requests carry an explicit low `temperature` (assert on the outbound body)
- [ ] **N>=10** post-fix runs at the new temperature, stating the rate — this comparison is the point of
      the issue, so both numbers must be reported even if the difference is small
- [ ] Answer quality on the 16-question set does not regress — lower temperature must not make answers
      terser or less accurate against the QA pass's ground-truth answer set
- [ ] Colocated unit test asserting `temperature` is present on tool-bearing requests
- [ ] Regression gate: 16-question set passes; zero raw upstream error text

### Notes

- **Report the honest result.** If the before/after rates are statistically indistinguishable at N=10,
  say so — the change is still correct per the provider's guidance, but the issue should not claim a fix
  it did not demonstrate.
- Related to the existing issue on the "Failed to call a function" error copy: that issue fixes what the
  user is *told*; this one attacks the underlying rate. Cross-reference both ways.
- Touches `openai-compatible.ts`, shared with the `delta.reasoning`, `tool_choice`, usage-accumulation
  and prompt-caching issues. This one is small and low-risk — it can land alongside the two urgent ones.
- The Anthropic adapter should get the same treatment for consistency.

yenienserrano:
## Description

Closes #8862
Closes #8863
Closes #8865

## Proposed Changes

Three related fixes to the AI Assistant's OpenAI-compatible provider adapter and orchestration loop, all touching the same tool-round code path:

- **Blank answers on reasoning-only streams (#8862).** Some reasoning models (`openai/gpt-oss-120b`, `qwen/qwen3.6-27b`) stream their entire answer on `delta.reasoning` instead of `delta.content` on a `general`-routed (no-tool) turn. `openai-compatible.ts` now buffers `delta.reasoning` and emits it as a last-resort fallback, but only once the whole call is confirmed to have produced no `content` and finalized no tool calls — it is never appended alongside a working answer. As defence in depth, `chat.ts`'s existing `!sawAnyDelta` no-text guard is widened to fire on every no-text ending (not just when a tool ran this turn), with a new `NO_ANSWER_MESSAGE` for the no-tool case so the copy never misdescribes a turn that had no table and no query.
- **`tool_choice: 'none'` 400 on the final round (#8863).** The forced-final round used to send `{tools, toolChoice: 'none'}` to stop the tool loop, which some models (gpt-oss) ignore and call a tool anyway, causing the provider to reject the whole request (`Tool choice is none, but model called a tool`). The final round now omits `tools` entirely instead, relying on the adapters' existing `if (options?.tools?.length)` guard to drop `tools`/`tool_choice` from the wire body — a structural terminator instead of a request-level hint the model can ignore.
- **No `temperature` sent for structured/tool calls (#8865).** Both adapters now forward `options.temperature` verbatim when defined (checked against `undefined`, not truthiness, so an explicit `0` survives). `chat.ts` sets `0` for the stage-1 router (a single structured pick out of a fixed enum) and `0.2` on tool-bearing orchestrate rounds, per Groq's tool-use guidance (0.0–0.5) — their default of `1.0` is a documented contributing cause of malformed tool calls. Plain-prose rounds and the connectivity-test call site are left unset, so a provider's own default is unaffected.

### Results and Evidence

Live validation gate (Amazon Bedrock, `openai.gpt-oss-120b`), from issue #8862:

- Blank bubbles: **0 of 27** records across the full gate run (vs 3 of 10 on the pre-fix baseline for the affected question).
- The underlying model silence itself is unchanged (3 of 10 runs both before and after, on questions where the model streams zero deltas for the entire turn) — the fix removes the *symptom* (empty bubble) by design; it does not reduce how often a model emits nothing.

### Artifacts Affected

- `plugins/wazuh-ai-assistant/server/providers/anthropic.ts`
- `plugins/wazuh-ai-assistant/server/providers/types.ts`
- `plugins/wazuh-ai-assistant/server/routes/chat.ts`

### Configuration Changes

None. `temperature` is an internal server-side option on `ChatStreamOptions`, not exposed as user-facing configuration.

### Documentation Updates

None.

### Tests Introduced

Colocated tests in `plugins/wazuh-ai-assistant/server/providers/openai-compatible.test.ts`:

- Reasoning-only stream fixture renders the buffered `reasoning` text; a content-only fixture is unaffected (regression); reasoning is never appended alongside a working answer.
- A stream with neither `content` nor `reasoning` still ends cleanly with just `done` (no phantom fallback text).
- Final-round request body carries neither `tools` nor `tool_choice` when no tools are offered; a tool-bearing request still carries both (regression).
- `options.temperature` is forwarded verbatim, including a literal `0`; omitting it leaves the field out of the body entirely (no default injected).

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


Ripdiegozz (approved):
## 🟡  Tests / 🟢 CR

### 🟢 Test 1 - Blank answers on reasoning-only streams (#8862)

https://github.com/user-attachments/assets/1e249e7c-6fc1-4577-ad37-3315efa88361

### 🟡 Test 2 — `tool_choice: none` 400 on final round (#8863)

**Setup:** bonsai-27b reasoning model, forced a long tool-calling chain to reach the orchestrator's final forced round.

**Prompt used:**
> "Get all agents, then for each agent group get the vulnerability count by severity, then cross-reference with the top 5 most triggered rules this week, and finally tell me which agent group has the highest critical-vulnerability-to-alert ratio."

**Result:** No 400 error ("Tool choice is none, but model called a tool") on the final round. Network tab confirms the mechanism directly: on the final round (no `tools` in the request body), the model attempts one more tool call (`get_vulnerabilities_by_agent`) and — with nothing to call — streams it as literal text via `delta.content` (`<tool_call><function=get_vulnerabilities_by_agent></function></tool_call>`) instead of a structured `tool_call` event. The request itself is accepted; the provider never rejects it. Confirms the fix achieves its goal (no 400), at the cost of the model's tool-call attempt leaking as raw text.

https://github.com/user-attachments/assets/aaf210c9-6761-41d7-88db-f05515eeaaf7

**Side finding (out of scope for this PR):** same leak observed on a prior run with `get_top_rules`. This appears to be a systematic behavior of bonsai-27b on the forced final round (it always tries one more tool call when it has more steps queued), not an intermittent glitch. Worth a follow-up issue to strip/suppress literal `<tool_call>` tags from `delta.content` before rendering.

### 🟢  Test 3 — `temperature` forwarded on structured/tool calls (#8865)

**Setup:** bonsai-27b, temporary `console.log` on the outgoing `body` in `openai-compatible.ts` (`temperature`, `hasTools`, `tool_choice`), removed after verification. Ran the full Batch B conversation end-to-end.

**Server log output, in order:**
```console
[openai-compatible] outgoing body { temperature: undefined, hasTools: false, toolChoice: undefined }
[openai-compatible] outgoing body {
  temperature: 0,
  hasTools: true,
  toolChoice: { type: 'function', function: { name: 'route_question' } }
}
[openai-compatible] outgoing body { temperature: 0.2, hasTools: true, toolChoice: 'auto' }
[openai-compatible] outgoing body { temperature: 0.2, hasTools: true, toolChoice: 'auto' }
[openai-compatible] outgoing body { temperature: 0.2, hasTools: true, toolChoice: 'auto' }
[openai-compatible] outgoing body { temperature: undefined, hasTools: false, toolChoice: undefined }
```

PR #8881, opened 2026-08-06
Keep minimal tool set on general-routed turns

### LINKED ISSUE #8864: A `general` routing decision removes the data path for the whole turn
## Description

When the stage-1 router classifies a question as `general` ("no Wazuh data needed"), stage 2 runs with
**no tools at all**. If that classification was wrong, the model cannot recover — it has no way to reach
the data even if it works out mid-answer that it needs it. One misclassification silently removes the
data path for the entire turn.

This is reachable with ordinary operational questions, because a vague question looks like chit-chat to
a classifier.

### Evidence

*"What should I look at first this morning?"* — a Tier-1 triage question — routed to `general`, was
offered **0 tools**, and fetched **no data**. Captured from the router probe:

```
stage1-router: 1 tools offered -> emitted [{"name": "route_question",
                                           "arguments": "{\"categories\":[\"general\"]}"}]
stage2-tools:  0 tools offered -> emitted []
```

The same question routed correctly to `get_critical_findings` on a different run, so this is
**non-deterministic**, not a fixed mapping.

Rate is not marginal. On `qwen/qwen3.6-27b`, **4 of 13 successful long-tail runs fetched no data at
all**. In the static coverage classification this question is `TYPED` — a purpose-built tool exists and
its parameters can express the question. The catalog was not the limitation; the routing was.

Two compounding causes:

**(a) The category description invites it.** `general` is described as:

> "No Wazuh data needed at all: greeting, meta-question, clarification, chit-chat."

*"What should I look at first?"* and *"anything I should worry about?"* are legitimately ambiguous
against that description.

**(b) The failure is terminal, not recoverable.** `resolveStage2Tools` returns `undefined` when the only
routed category is `general`, and the orchestration loop then calls the provider with `tools` unset. The
model has no escape hatch, no `search_wazuh_data`, nothing.

### Code references

- `plugins/wazuh-ai-assistant/server/tools/router.ts` — `CATEGORY_DESCRIPTIONS.general`, and
  `resolveStage2Tools` returning `undefined` for `general` alone
- `plugins/wazuh-ai-assistant/server/routes/chat.ts` — the orchestrate loop, which passes
  `tools: undefined` through to the adapter

### Suggested treatment

Two changes, both small. The second is the important one.

**1. Narrow the `general` description and add an explicit exclusion:**

> `general`: greetings, thanks, or questions about the assistant itself. **If the user is asking
> anything about their own environment — however vaguely — do NOT pick general.**

**2. Never send zero tools.** Even on `general`, offer a minimal set — `get_security_summary` plus
`search_wazuh_data`. Cost is roughly two schemas (~800 tokens) and only on turns that were classified
as needing no data, which are the cheapest turns anyway. In exchange, a misclassification becomes
**self-correcting** rather than terminal.

Change 1 alone would reduce the rate; only change 2 removes the failure mode, because no classifier is
perfect and this one is demonstrably non-deterministic.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: **N=10** runs of *"What should I look at first this
      morning?"*, stating how many routed to `general` and fetched no data
- [ ] After the fix, a `general`-routed turn is still offered a minimal tool set, and the model can
      call it
- [ ] **N>=10** post-fix runs, stating how many turns answered with grounded data
- [ ] A genuine greeting ("hello", "what can you do?") still answers without a pointless tool call —
      the fix must not force a query on every conversational turn (regression)
- [ ] Colocated unit test: `resolveStage2Tools(['general'])` returns the minimal set rather than
      `undefined`
- [ ] Regression gate: 16-question set passes, including the deliberately ambiguous question

### Notes

- **Highest-leverage routing fix in this batch.** It converts a terminal misclassification into a
  recoverable one, and it is two small edits.
- Interacts with the blank-answer issue: the `general` path is also where blank answers were observed,
  because it is the path that runs without tools. They are **independent defects** and both need
  fixing — this one reduces how often that path is entered, it does not stop the blank answer.
- Touches `router.ts` (not shared with any other issue in this batch) and one line in `chat.ts`
  (shared with the `tool_choice` issue).
- Watch the token cost claim during QA: confirm the added schemas really are only sent on
  `general`-routed turns and do not leak into every request.

### Update — 2026-08-06 validation gate (Amazon Bedrock, `openai.gpt-oss-120b`)

A candidate fix implementing both suggested changes was measured live on the reproduction
question (*"What should I look at first this morning?"*):

- **No-tool rate: 2 of 10** (N=10), vs the pre-fix baseline of **8 of 10** on the same question
  (Fisher exact p ≈ 0.007), replicated same-day at **2 of 5** in a second run. The result is
  robust to every counting convention tried by an independent re-derivation of the raw records.
- Mechanically confirmed: general-routed turns now carry tool schemas (input tokens 1,281 →
  6,814 on such turns; `resolveStage2Tools` can no longer return `undefined` or an empty set).
- **Attribution caveats** (state these in any PR that cites the numbers): (a) the baseline was
  captured on Groq, post-fix on Bedrock — same gpt-oss-120b weights, different serving stack
  (~3.3× slower per output token, so latency is not comparable); (b) ~17 commits separate the
  two builds, including a temperature change that directly affects routing determinism — the
  delta is not attributable to this fix alone; (c) at the outcome level, 2 of 10 post-fix runs
  still answered without any data, and several data-fetching runs chose 4–24h windows against a
  dataset whose newest finding was 11 days old (honest zero-row answers). Offering tools is
  fixed; window selection and residual no-tool turns remain model behavior.

yenienserrano:
## Description

Closes #8864

When the AI Assistant's stage-1 router classified a question as `general` ("no Wazuh data needed at all"), stage 2 ran with **zero tools**. If that classification was wrong — which happens with ordinary operational questions that look like chit-chat to the classifier — the model had no way to recover the data path for the rest of the turn. Measured on `qwen/qwen3.6-27b` long-tail runs: 4 of 13 successful runs fetched no data at all because the same question routed to `general` on one run and to `get_critical_findings` on another (non-deterministic).

## Proposed Changes

- `server/tools/router.ts`: `resolveStage2Tools` never returns `undefined`/empty anymore. A lone `general` route now resolves a minimal recovery set (`get_security_summary` + `search_wazuh_data`) instead of no tools, so a stage-1 misclassification is recoverable mid-turn instead of losing the data path for the whole turn. `tool_choice` stays `'auto'`, so a genuine greeting still gets a plain-text answer — the minimal set is only an escape hatch the model may reach for.
- `server/tools/router.ts`: narrowed the `general` category's routing description and the routing prompt instructions with an explicit exclusion — do NOT pick `general` when the user asks anything, however vaguely, about their own environment — to reduce the misclassification rate at the source.
- `server/routes/chat.ts`: updated the `Stage1Result.tools` type from `ToolSpec[] | undefined` to `ToolSpec[]` and its doc comment to match the new invariant.
- `eval/run_plumbing.js`: updated the `[[route:general]]`-only eval comment to reflect that the turn now resolves the minimal recovery set rather than no tools (the assertion itself is unchanged — it's a genuine greeting with no tool marker, so the plain-text branch fires regardless of which tools were offered).
- `server/tools/router.test.ts` (new): unit tests proving `resolveStage2Tools(['general'])` returns the minimal set (never undefined/empty), that a normal data-category resolution is unaffected, that `general` combined with a data category still resolves that category, and that the routing prompt's `general` description carries the exclusion.

### Results and Evidence

Backend-only fix (routing/tooling logic), no UI changes.

```
plugins/wazuh-ai-assistant$ yarn test:jest --runInBand
Test Suites: 56 passed, 56 total
Tests:       718 passed, 718 total
```

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (server)

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `plugins/wazuh-ai-assistant/server/tools/router.test.ts` (new): 4 unit tests covering the "never route zero tools" fix and the unaffected data-category paths.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


Ripdiegozz (dismissed):
## 🟢 Tests / 🟢 CR

### 🟢 Test 1 — `general` route no longer strands the turn without tools (#8864)

**Prompts used:**
> "Hi there, thanks!"
> "How's everything looking?"
> "Anything I should worry about?"
> "What can you help me with?"

**Result:** No regression and no misroute across 5 live runs. The pure greeting answered in plain text with zero tool calls, confirming `tool_choice: 'auto'` still lets a genuine chit-chat turn skip tools entirely. Both ambiguous operational questions ("How's everything looking?", "Anything I should worry about?") correctly routed to real data categories, never `general`, pulling in `get_agents`, `get_critical_vulnerabilities`, and up to three more tools in a single turn, all resolving with actual data. The narrowed routing-prompt exclusion is doing its job in practice, not just on paper.

https://github.com/user-attachments/assets/55e018d9-f5b5-4e70-b0ec-691d4f912a06

**Side finding (out of scope for this PR):** `get_security_summary` returns an empty table on every run, regardless of routing. It aggregates on `wazuh.rule.category`, but some documents matching its own severity filter have that field unset.

Ripdiegozz (approved):

Desvelao (approved):

PR #8882, opened 2026-08-06
Fix in-stream tool-call failure copy blaming the user's prompt

### LINKED ISSUE #8855: A tool-call failure tells the user to change their prompt when the model is the problem
## Description

When the configured model fails to produce a valid tool call, the chat shows the provider's raw explanation, which tells the user to change their question. In most cases the question was fine and the model is the problem.

Reproduced with a Groq provider on `llama-3.3-70b-versatile` — a model the provider markets as tool-capable, and close to the examples the settings form itself suggests. The question was ordinary: *"Which critical findings should I look at first, and which agents are affected?"* The result:

> **Something went wrong**
> Failed to call a function. Please adjust your prompt. See 'failed_generation' for more details.

Two problems with that as user-facing copy:

- **It misattributes the cause.** The user is told to adjust their prompt. The actual failure is that the model emitted a malformed tool call, and the usual remedy is a model with stronger function-calling support, not a reworded question. A simpler question (*"How many critical findings are there?"*) succeeded immediately on the same model and configuration, which is exactly the pattern that makes "adjust your prompt" misleading — it sometimes works, for the wrong reason.
- **It points at something the user cannot see.** `failed_generation` is a field in the provider's API response. There is no way to view it from the chat, so the instruction is a dead end.

The plugin already knows this failure mode. The provider form's own help text says: *"Tool calling needs a model with solid function-calling support. Small or base models often fail with tool errors."* That knowledge just is not applied when the predicted thing actually happens.

### Suggested treatment

Detect the tool-call failure case in the provider adapter's error handling and replace the copy with something that names the likely cause and the action, for example:

> The model could not produce a valid tool call, so the assistant could not query your data. This usually means the configured model's function-calling support is unreliable for this question. Try a model with stronger tool support, or ask something simpler.

The provider's raw text can stay available for debugging (server log, or a details disclosure), just not as the primary instruction.

### Task

- [ ] Detect the malformed-tool-call failure in the provider error path (OpenAI-compatible adapters surface it as an in-stream error mentioning a failed function call)
- [ ] Replace the user-facing copy so it names the likely cause (model tool-calling support) instead of instructing the user to change their prompt
- [ ] Do not surface provider-internal field names (`failed_generation`) in the UI
- [ ] Keep the raw provider message available for diagnosis, out of the primary message

### Additional information

Verified on Wazuh 5.0.0-beta3 against a real Groq provider. The same provider and model answer correctly for simpler questions, so this is about the failure copy rather than the provider being misconfigured — Status showed `OK (171 ms)` at the time.

Ripdiegozz:
## Description

When a provider (e.g. Groq) reports that a model failed to produce a valid tool call, it sometimes surfaces this as an **in-stream** error frame (HTTP 200) rather than a pre-stream HTTP 400. That in-stream frame was passed straight through to the user, showing raw provider text such as:

> Failed to call a function. Please adjust your prompt. See 'failed_generation' for more details.

This misattributes the cause (telling the user to change their prompt when the actual problem is the configured model's unreliable function-calling support) and leaks the provider-internal `failed_generation` field name, which the user cannot inspect from the chat.

The pre-stream HTTP-400 case for this same failure mode was already handled correctly in `retry.ts` (friendly message, redaction, one-shot retry) — this PR extends that same classification to the two in-stream error branches that were still leaking raw provider text.

Closes #8855

## Proposed Changes

- Added `describeToolUseFailedStreamMessage()` in `plugins/wazuh-ai-assistant/server/providers/retry.ts`, reusing the existing `tool_use_failed` / `failed_generation` marker detection to classify an in-stream provider error message.
- Wired it into the in-stream error branches of `openai-compatible.ts` and `anthropic.ts`, replacing the raw provider message with friendly copy when a marker is present, and falling back to the original message unchanged otherwise (no regression for other error types).
- The new in-stream message is intentionally distinct from the existing HTTP-400 terminal message, since the in-stream frame carries no `failed_generation` snippet to append.
- Scope is deliberately limited to these two in-stream branches — the HTTP-400 path and the separate malformed-tool-call-arguments JSON path were already correct and are untouched.

### Results and Evidence

Manually reproduced against a real Groq provider (`llama-3.3-70b-versatile`) on `5.0.0` pre-fix, and confirmed the fix resolves it. Also manually confirmed no regression: unrelated provider failures (HTTP 429 rate limit, HTTP 413 request-too-large, and an HTTP 400 tool-schema `invalid_request_error`) still surface their own descriptive messages unchanged, since none of them contain the `tool_use_failed`/`failed_generation` markers.

Before (raw provider text, current `main`/`5.0.0` behavior):
> Failed to call a function. Please adjust your prompt. See 'failed_generation' for more details.

After (this PR):
> The assistant couldn't get a valid tool call from the configured model for this request. This usually means the model's function-calling support isn't reliable enough for this kind of question. Try a model with stronger tool-calling support, or ask a simpler question.

### How to Test

1. Configure a Groq provider (Settings → Providers) with model `llama-3.3-70b-versatile`.
2. Start a new chat and ask a compound question that requires querying Wazuh data, e.g.:
   > "Which critical findings should I look at first, and which agents are affected?"
3. Repeat the same question a few times if needed — Groq's constrained tool-call decoding occasionally produces a malformed call, which is what this PR fixes.
4. When it fails, confirm the chat shows the friendly copy naming the model's function-calling support as the likely cause — **not** the raw `"Failed to call a function. Please adjust your prompt. See 'failed_generation' for more details."` text, and **not** the literal string `failed_generation` anywhere in the banner.
5. Sanity-check no regression: trigger an unrelated provider failure (e.g. a rate limit, by sending several requests quickly) and confirm its own descriptive message still shows normally — this fix must only intercept the tool-call-failure shape, nothing else.

### Artifacts Affected

- `plugins/wazuh-ai-assistant/server/providers/retry.ts`
- `plugins/wazuh-ai-assistant/server/providers/anthropic.ts`

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `retry.test.ts`: unit tests for `describeToolUseFailedStreamMessage()` — marker present (`tool_use_failed`, `failed_generation`) returns the friendly message; no marker returns `undefined`; the friendly message never contains the literal `failed_generation` substring.
- `openai-compatible.test.ts` (new): wiring test confirming an in-stream error frame with a marker yields the friendly message, and one with no marker passes through unchanged.
- `anthropic.test.ts` (new): same wiring coverage for the Anthropic adapter.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


Desvelao (dismissed):

rodrigofez (dismissed):
> [!IMPORTANT]
> This fix depends on a specific, intermittent way a model can fail, so I could not
> reproduce it with any provider I have access to. I verified it against a mocked
> provider response that emits the exact trigger instead.

Added a mock provider which emits the HTTP-200 in-stream error frame on demand. Confirmed the friendly copy is shown, `failed_generation`

Image

gonzaarancibia (dismissed):

gonzaarancibia (approved):

PR #8885, opened 2026-08-06
Validate DSL fields and aggregate breakdowns over matched rows

### LINKED ISSUE #8868: `search_wazuh_data` silently returns nothing when the model guesses a wrong field name
## Description

`search_wazuh_data` — the escape-hatch tool used when no typed tool covers a question — accepts a raw
OpenSearch DSL body from the model but never checks that the field names in it actually exist. When
the model guesses wrong, OpenSearch does not error: it returns zero buckets for a query that matched
real documents, and the model reports that as a real answer. Neither the model nor the user can tell
the difference between "nothing is happening" and "the query was wrong." Observed in 1 of 3 escape-hatch
queries captured (small sample, stated as such), on a fallback that ~22% of realistic questions
depend on — a silent-wrong-answer defect on the exact path that covers the catalog's gaps.

A companion defect makes this easier to miss: the prompt already tells the model to retry when a
narrowly-filtered query returns 0 rows for activity that plausibly exists, and on the reproduction
below it did not comply. The fix in this issue is therefore mechanical, not a second instruction.

### Evidence

Asked "Which host is noisiest right now?" — a question with no per-agent aggregation tool —
`openai/gpt-oss-120b` recognised the gap and hand-wrote a real aggregation through the escape hatch:

```json
{"size":0,
 "query":{"bool":{"filter":[{"range":{"@timestamp":{"gte":"now-1h","lte":"now"}}}]}},
 "aggs":{"top_agents":{"terms":{"field":"agent.name.keyword","size":5,"order":{"_count":"desc"}}}}}
```

Structurally correct. `agent.name.keyword` does not exist: `_field_caps` on `wazuh-findings-v5*`
returns only `agent.name` (keyword) and `wazuh.agent.name` (keyword) — there is no `.keyword`
subfield, because these fields are already `keyword`.

Replaying the model's query with a 90-day window, to isolate the field error from its (also wrong)
1-hour window:

```
"hits":{"total":{"value":229}}                       <- 229 documents matched
"aggregations":{"top_agents":{"buckets":[]}}         <- zero buckets, NO ERROR
```

**229 documents matched, the aggregation returned nothing, and no error was raised.** The correct
field name yields web-prod-01 (91), web-prod-02 (45), win-dc-01 (41), db-prod-01 (31), wazuh-aio-5
(21).

This is a pattern, not a one-off: the same guessing appeared in another query's `_source` list, which
requested `src.ip` and `dst.ip` — neither exists (harmless there, since a missing `_source` field is
simply absent, not an error). **Sample size, stated plainly: 3 escape-hatch queries were captured.
2 of 3 contained at least one invented field name; in 1 of 3 it broke the answer.** Small n, but the
failure shape is deterministic once it occurs — OpenSearch will never error on it — so the risk does
not shrink with scale.

The prompt already covers the general case and was not followed. It instructs the model to "retry
once with a broader filter" when "a narrowly-filtered query returns 0 rows for activity that
plausibly exists" — and on this reproduction the model accepted the 0 buckets and reported nothing.
Instruction has been tried and failed, which is why this issue proposes a mechanical check instead of
another prompt change.

### Code references

- `plugins/wazuh-ai-assistant/server/tools/catalog/search-wazuh-data.ts` — the `query_dsl` parameter;
  no field-name check exists before the request is sent
- `plugins/wazuh-ai-assistant/server/tools/executor.ts` — `executeIndexerRequest`, where
  `checkIndexAllowlist` + `applySafetyValves` + `lintDsl` run on every indexer request; a
  field-existence check belongs in this same chokepoint
- `plugins/wazuh-ai-assistant/server/tools/catalog/common.ts` — `validateTimeBound`'s doc comment:
  "throws a descriptive Error otherwise, which the orchestration loop turns into a bounded
  tool_result error for the model to self-correct" — the pattern this issue proposes reusing
- `plugins/wazuh-ai-assistant/server/tools/digest.ts` — `Digest.counts` / `buildDigest`, where the
  companion zero-row hint below would attach

### Suggested treatment

**Recommended: validate field names in the submitted DSL against the live mapping before executing,
and return a corrective tool error rather than an empty result** — e.g. `field
"agent.name.keyword" does not exist; available: "agent.name", "wazuh.agent.name"`. The orchestration
loop already turns a thrown tool error into a bounded self-correction opportunity — `common.ts`'s
`validateTimeBound` documents exactly this pattern for time-bound validation — so the model would very
likely fix itself on the retry, at the cost of one extra round trip only when the guess is wrong.

Two weaker alternatives, in order of value: detect the specific "query matched rows but every
aggregation is empty" shape and flag it in the digest so the model cannot read it as a true zero; or
put the real field names in the `search_wazuh_data` tool description so the model is less likely to
guess in the first place. Fix 1 is the one that converts a silent wrong answer into a self-healing
retry; the other two reduce the rate but do not close it.

**Companion change — mechanical zero-row hint.** The prompt already instructs a broader retry on a
suspicious zero-row result, and the model did not comply on the reproduction above, so make it
mechanical instead of relying on a second instruction: when a tool returns 0 rows and the query
carried 2 or more filters, add that fact to the digest:

```json
"counts": {"total": 0, "returned": 0, "truncated": false},
"hint": "0 rows. Filters applied: agent_name, time_range, rule_tags.
         A wrong field name or an over-narrow filter produces this same result."
```

This pairs directly with the field-name validation above: one catches the wrong field, the other
catches an over-narrow query on a field that does exist.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: a `search_wazuh_data` aggregation on a nonexistent
      field (e.g. `agent.name.keyword`) matches documents but returns zero buckets with no error (N
      runs stated; observed rate ~1 in 3 on escape-hatch queries)
- [ ] After the fix, the same query is rejected with a tool error naming the invalid field and the
      valid alternatives, before it reaches OpenSearch
- [ ] A tool result with 0 rows and ≥2 filters carries a `hint` field in the digest naming the
      filters applied
- [ ] A tool result with 0 rows and <2 filters carries no hint (no noise on a legitimately narrow,
      correct query)
- [ ] N>=10 post-fix runs of the noisiest-host reproduction, failure rate stated — target zero
      silent-empty results
- [ ] Colocated unit test asserting field-name validation rejects `agent.name.keyword` and accepts
      `agent.name`; a second test asserting the hint appears only at ≥2 filters
- [ ] Regression gate: 16-question set passes; zero blank answers, zero raw upstream error text

### Notes

- Folds in the zero-row hint safeguard from the routing-quality QA pass — same digest change, same file
  (`digest.ts`), filed together rather than separately.
- `search-wazuh-data.ts`'s `buildRequest` already `JSON.parse`s `query_dsl` into an object before it
  reaches `executor.ts`, so field-name validation does not strictly require the `query_dsl`
  string-or-object fix (#8861). Both issues still touch the DSL-handling area closely
  enough — and #8861's notes already say to land it first — that landing #8861 first
  removes one variable when testing this one.
- Field-name validation needs a source of truth for "which fields exist" (a live mapping lookup, or a
  maintained allowlist) — left to the implementer, not specified further here.
- Wave 2, alongside #8869 — both touch `guardrails.ts` / `digest.ts`.

### LINKED ISSUE #8870: Aggregative answers are extrapolated from truncated sample rows
## Description

For aggregative questions ("which agents are affected", "which rules fired most"), the model reasons
over `MAX_SAMPLES = 5` sample rows and its prose reads as if it covered the full result set. The
objection that sending more rows would inflate tokens and latency is correct, and `MAX_SAMPLES` should
stay as it is — the fix is not "send more data" but "send the right shape of data." An aggregation is
O(buckets), not O(rows): a small breakdown answers the same question more accurately than 5 sample
rows, at a fraction of the cost of sending every row.

A second, independent defect compounds this: the 5 rows sent today are not a random sample. They are
the newest 5 rows from a timestamp-descending sort, so an agent whose only matching rows are older is
invisible to the model regardless of how many times the question is asked.

### Evidence

`server/tools/digest.ts:27` — `const MAX_SAMPLES = 5`. The file's own comment: "the model only gets
this bounded digest — counts, a breakdown when the query was an aggregation, and up to 5
whitelisted-column sample rows."

Captured live: for a query returning all 26 critical findings
(`"counts":{"total":26,"returned":26,"truncated":false}`), the digest handed to the model contained
5 sample rows, drawn from the newest end of a timestamp-descending sort. The UI table beneath the
answer showed all 26.

The consequence, observed: `llama-3.3-70b-versatile` on "how many critical findings do we have?"
correctly answered 26 (from `counts.total`) and then asserted *"The rule 'Executable file dropped in
folder commonly used by malware' triggered 2 times for agent 'web-prod-02' and once for
'web-prod-01'. The rule 'Shellshock attack detected' triggered 3 times..."* — the true counts are
**13 and 13** for those two rules. The numbers are not supported by the 5 rows the model saw, nor by
the data.

The 26 critical findings belong to three agents, not two: web-prod-01 (16), web-prod-02 (8),
wazuh-aio-5 (2). `wazuh-aio-5`'s 2 critical findings are the oldest in the dataset, so in a 5-row
sample drawn from the newest end of a timestamp-descending sort they never appear — which is
precisely why every model observed named two affected agents instead of three. Because the sort order
is deterministic, the same 5 rows are returned on every run against the same data; repeat testing
cannot detect this, since the same agent is excluded every time.

The token cost of the alternatives, measured:

| Approach | Payload | Accuracy on "which agents are affected?" |
|---|---|---|
| Today: 5 sample rows | ~1,860 chars ≈ 520 tokens | wrong — names 2 of 3 |
| All 26 rows | ~9,700 chars ≈ 2,700 tokens | right, but expensive |
| 5 samples + a 3-bucket breakdown | ~520 + ~40 tokens ≈ 560 tokens | right |

Sending every row is roughly 5x the token cost of today's sample, and even then "right" still depends
on the model correctly counting 26 rows by hand rather than misreading them — the same failure mode as
the rule-count example above. The breakdown avoids that dependency entirely because it is a real
aggregation, not something the model counts.

The `Digest` type already has a `breakdown` field (`digest.ts`). Today it is populated only when the
executed query itself contained an aggregation — `buildBreakdown` reads `result.aggregations`, which
only exists when the request sent to OpenSearch had an `aggs` clause — not when the natural-language
*question* is aggregative but the typed tool it maps to only returns rows.

### Code references

- `plugins/wazuh-ai-assistant/server/tools/digest.ts#L27` — `MAX_SAMPLES = 5`
- `plugins/wazuh-ai-assistant/server/tools/digest.ts` — `Digest.breakdown` and `buildBreakdown`
  (~L409-437), which derives the breakdown from `result.aggregations` regardless of caller
- `plugins/wazuh-ai-assistant/server/tools/digest.ts` — `buildDigest` (~L448-489), where
  `samples = rows.slice(0, MAX_SAMPLES)` takes the first N rows in whatever order the underlying
  query returned them (the sort-bias point above)

### Suggested treatment

**Recommended: populate `breakdown` for aggregative questions, not only for aggregative queries.**
Where a typed tool's question is really an aggregation ("which agents", "which rules"), have it also
run (or attach) a real grouped count instead of relying on the model counting sample rows — the
mechanism already exists in `digest.ts`; it just needs a producer on the typed-tool path. This turns a
model-judgement answer into a measured one, at roughly 40 tokens over today's cost, and it answers the
token-cost objection directly because it does not require raising `MAX_SAMPLES`.

Keep `MAX_SAMPLES` at 5. Raising it — e.g. to fit small result sets like the 26-row case — also helps
in that one case, but costs more tokens for a narrower fix and still depends on the model reading the
rows correctly rather than being handed a count.

The sampling-bias defect (newest-5, not random) is separate from the breakdown fix and needs its own
remedy — either sample across the result set rather than taking the first N in sort order, or, more
cheaply, note in the digest that the samples are not representative of which entities appear, so the
model does not treat an absence in the sample as a fact about the data.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: an aggregative question against a >5-entity result set
      produces a model answer that omits or miscounts an entity present only in older rows (N runs
      stated)
- [ ] After the fix, the digest for an aggregative typed-tool question carries a populated
      `breakdown` (or equivalent grouped count), and the model's answer matches it rather than a
      hand-count of sample rows
- [ ] The 26-critical-findings reproduction above is re-run post-fix and both the rule counts (13/13)
      and the agent count (3, including wazuh-aio-5) are correct
- [ ] `MAX_SAMPLES` remains 5 (no token-cost regression from simply raising the cap)
- [ ] N>=10 post-fix runs, failure rate stated
- [ ] Colocated unit test in `digest.test.ts` asserting `breakdown` is populated for a typed-tool
      aggregative case, and that `samples` length stays capped at 5
- [ ] Regression gate: 16-question set passes; zero blank answers, zero raw upstream error text

### Notes

- The sample-bias defect (newest-5 sort order hiding an agent) is deterministic: since it depends
  only on sort order, not randomness, it will not surface again in repeat testing of the same data,
  and should not be treated as fixed just because a re-run looks clean.
- Wave 2. Touches `digest.ts`, which #8868 (zero-row hint) and #8869 (clamp disclosure) also
  touch — all three land in the same file; sequence to avoid merge conflicts, order does not
  otherwise matter.

### Update — 2026-08-06 validation gate: the breakdown must aggregate MATCHED rows, not returned rows

A candidate fix adding a synthetic breakdown was measured live and revealed a hard requirement
this issue must carry: **a breakdown computed over the rows the tool *returned* (`limit:N`) is
worse than no breakdown when `returned < total`, because it is presented as authoritative.**

Observed (Bedrock, gpt-oss-120b): on a query with `counts = {"total":26,"returned":20,
"truncated":true}`, the synthetic breakdown contained `web-prod-01: 13, web-prod-02: 7` and rule
counts `11/9` — each pair summing to exactly 20 (the returned page), while the true 26-row
distribution is **16/8/2 and 13/13**. The third affected agent (`wazuh-aio-5`, oldest rows) was
absent from the breakdown entirely, so the model could not have named it — the same
sort-window-bias defect as the original 5-sample case, at page size instead of sample size. A
second question showed the same mechanism with larger errors (reported 12/4/4 vs true 18/11/13).
The accompanying digest note instructed the model to *trust* the breakdown ("Use
counts/breakdown, not an absence from these rows…"), which converts the bias into a confident
wrong answer. With `limit:500` (page ≥ matched set) the same mechanism produced exactly correct
answers — confirming the breakdown approach itself is right when fed the full matched set.

Additional acceptance criteria therefore:

- [ ] The breakdown is derived from the **matched** set (a real aggregation), or is explicitly
      labeled as covering only the returned page — never presented as the population
- [ ] The reproduction includes a `limit` smaller than the matched total (e.g. `limit:20` on a
      26-row set) and the per-entity counts still match the true distribution

yenienserrano:
## Description

Closes #8868
Closes #8870

The AI assistant's `search_wazuh_data` tool never validated that model-supplied
DSL field names existed against the live index mapping: a wrong field name (e.g.
`agent.name.keyword` instead of `agent.name`) produced zero aggregation buckets
with no error, indistinguishable from a true zero (#8868). Separately, aggregative
questions ("which agents are affected", "which rules fired most") were answered
by the model reasoning over `MAX_SAMPLES = 5` sample rows drawn from the newest
end of a timestamp-descending sort, so an agent whose only matching rows were
older was invisible regardless of how many times the question was asked (#8870).

## Proposed Changes

- Added `field-validation.ts`: extracts every field name it can attribute with
  certainty from a parsed DSL body (filter clauses, agg `field` params, sort,
  `_source`) and checks it against the target index pattern's live mapping via
  `_field_caps` (cached 5 min per pattern), gated on a new
  `ToolDefinition.validateFieldNames` flag so typed tools (which cannot guess a
  field wrong) pay no round trip. A miss throws a descriptive error naming
  near-miss alternatives, which the orchestration loop already turns into a
  bounded self-correction retry.
- `digest.ts` now attaches a `hint` naming the filters applied whenever a query
  returns 0 rows with 2+ filter clauses.
- Attached a real `terms` aggregation per `FINDING_BREAKDOWN_DIMENSIONS` to every
  finding-hits tool's own Indexer request (`catalog/common.ts`'s
  `FINDING_BREAKDOWN_AGGS`), so `buildBreakdown` gets a population-true
  distribution for these 8 tools instead of one derived from sample rows.
- `buildSyntheticBreakdown` remains as the fallback for callers with no real
  aggregation attached; `buildDigest` now tags a truncated-page fallback with a
  `Digest.breakdownNote` caveat instead of letting the model treat it as
  authoritative, and `buildSamplesNote`'s wording no longer tells the model to
  trust a breakdown that isn't population-true.
- `executor.ts` passes `privacy.ts`'s `applyFieldPolicy` an identity `aggFields`
  map for the synthesized dimensions so synthetic buckets get scrubbed exactly
  like a real aggregation's would.

### Results and Evidence

Colocated unit tests cover both fixes:

- `field-validation.test.ts` — field extraction from filter/agg/sort/`_source`
  clauses, and `validateQueryFields` rejecting an invented field name while
  accepting a real one.
- `digest.test.ts` / `common.test.ts` — the zero-row hint appearing only at ≥2
  filters, and the breakdown being derived from the matched set (via a real
  aggregation) rather than the returned page.

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (server-side tools only: `catalog/`, `digest.ts`,
  `executor.ts`, `field-validation.ts`, `guardrails.ts`, `types.ts`)

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `plugins/wazuh-ai-assistant/server/tools/field-validation.test.ts` (new)
- `plugins/wazuh-ai-assistant/server/tools/catalog/common.test.ts` (extended)
- `plugins/wazuh-ai-assistant/server/tools/digest.test.ts` (extended)

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


rodrigofez (approved):
## Review

**Method.** A language model does not let you control the arguments of a tool
call, so I used the test provider in `eval/mock_provider.js`. This provider makes
the exact tool call that the prompt specifies, and it uses no language model. The
tool calls still go through the normal chat process. I read each digest from the
test provider's `/debug/requests` endpoint.

Index `wazuh-findings-v5-*`, 4599 findings
from 2 agents.

| Case | Result |
| --- | --- |
| Breakdown over all the rows that the query finds (#8870) | 🟢 |
| The plugin refuses an unknown field name (#8868) | 🟢 |
| The hint for zero rows with 2 filters, and no hint with 1 filter | 🟢 |

For #8870, the breakdown gives `rpm.local 4459` and `deb.local 140`. These counts
agree with a direct query to the indexer. The query returned only 20 rows, and
all 20 rows come from `rpm.local`. Therefore these counts can come only from a
true aggregation, not from the returned page.

For #8868, I sent the same query body directly to the indexer. The result is
`hits.total = 0` with no error. This change replaces that silent result with an
error.

### Problem: the list of alternative field names is not useful

A `term` filter on `agent.name.keyword` gives this error:

```
Field "agent.name.keyword" does not exist on "wazuh-findings-v5-*"; available:
"log.syslog.severity.name", "process.saved_group.name", "faas.name",
"process.entry_leader.attested_groups.name", "process.parent.elf.sections.name".
```

The correct name is `wazuh.agent.name`. It is not in the list.

`findNearMisses` keeps each field with the same last part of the name (`name`).
This index has 201 such fields. The function keeps the first five in the sequence
of the mapping, and does not put the best names first. `wazuh.agent.name` is at
position 106.

Only the language model reads this list. Five incorrect names make a correct
second attempt less probable.

### One item I could not test

I could not confirm that a language model corrects the field name after it reads
this error. Of the models available to me, `llama3.2` does not make right tool calls,
and `gemma4:12b` fails in the same manner with a correct field name, so I cannot
attribute its failure to the error message. However this change is supposed to only add a new reason for a rejection.

PR #8887, opened 2026-08-06
Add finding artifact filters, scope/vocabulary, and consolidated agent inventory tool

### LINKED ISSUE #8871: Tool descriptions lack negative scope, so the model silently substitutes a neighbouring tool
## Description

The findings tools' descriptions do not say what they exclude, so the model silently substitutes a
narrower tool for a broader question instead of saying so. Asked for "everything that happened" on an
agent, the model called a tool that reads only rule-matched detections, got zero rows, and reported that
nothing happened — when the honest answer is that no rule fired and the raw event stream was never
checked. This is the cheapest coverage change in the whole set: adding scope and vocabulary costs
roughly 15 tokens per tool description. Descriptions ride in every stage-2 request for their routed
category, so the recurring cost is ~15 tokens x the tools offered that turn (~180 tokens when all 12
findings-category schemas are offered) — still an order of magnitude below a single new tool's
~400-500 token schema, and it buys behaviour no new tool can: the model *saying* what it did not
check.

### Evidence

Asked *"Give me everything that happened on web-prod-01 in the last hour"*, the model called:

```
search_findings_by_agent { agent_name: "web-prod-01", limit: 500, time_range_gte: "now-1h" }
```

`search_findings_by_agent` reads `wazuh-findings-v5*` — rule-matched detections only. The result was
`{"returned":0,"truncated":false}`. **`qwen/qwen3.6-27b` made the identical tool choice on the same
question.** Two different model families converged on the same substitution, which is evidence this is a
description/catalog problem rather than a model quirk.

`search_findings_by_agent`'s current description (`search-findings-by-agent.ts`), verbatim:

> "Searches security findings for findings from one named agent, within a time range. Optional severity
> narrows to exactly that severity, or to a floor/ceiling via severity_comparison."

Nothing in it states what the tool does **not** cover. The other 7 finding-hits tools that share the same
baseline table/digest shape (`get_critical_findings`, `get_findings_by_time`, `get_brute_force`,
`get_suspicious_powershell`, `search_findings_by_multiple_agents`, `search_findings_by_rule_title`,
`search_findings_by_rule_tag` — enumerated together in `common.ts`'s comment above
`STANDARD_FINDING_TABLE_COLUMNS`) have the same gap.

**The system prompt already carries the instruction this failure violates, and the model did it anyway.**
`server/prompts.ts:48-51`, verbatim:

> "If no tool matches the question exactly, try search_wazuh_data with a minimal, correct query; if you
> cannot express it within its rules, say plainly what you can and cannot check with the available tools
> — never silently answer a narrower question than the one asked."

The instruction exists and was not followed, on two model families. That is why this fix belongs in the
tool descriptions the model consults when *choosing* a tool, not in the prompt it is already meant to
follow once a tool is chosen — the prompt-level fix has already been tried, in effect, and failed to
prevent this exact case.

Separately, the plugin's own source already states the events/findings distinction, just not where the
model sees it when picking a tool. `find-document-by-field.ts`'s `index_pattern` parameter description
says, verbatim: "ALL normalized events, matched or not -- most raw log/telemetry documents live ONLY
here, not in findings."

### Code references

- `plugins/wazuh-ai-assistant/server/tools/catalog/search-findings-by-agent.ts` — `spec.description`,
  the tool implicated by the measured failure
- `plugins/wazuh-ai-assistant/server/tools/catalog/common.ts` — the comment enumerating the 8
  finding-hits tools that share the same description shape and the same gap
- `plugins/wazuh-ai-assistant/server/prompts.ts:48-51` — the existing "never silently answer a narrower
  question" instruction that did not prevent this
- `plugins/wazuh-ai-assistant/server/tools/catalog/find-document-by-field.ts` — `index_pattern`'s
  description, which already states the findings-vs-events distinction the other tools lack

### Suggested treatment

Two additive changes to `spec.description`, both cheap and neither requiring a new tool or a schema
change:

**1. Negative scope.** Add one clause to each finding-hits tool's description: it covers rule-matched
detections only, and does not include raw events that matched no rule. `search_findings_by_agent` first,
since it is the one with measured evidence; extend to the other 7 finding-hits tools sharing the gap,
since the underlying index and failure mode are identical for all of them.

**2. Synonym vocabulary, not a keyword map.** Analysts phrase the same request many ways: boxes, hosts,
machines, endpoints, servers, assets, systems all mean *agent*; alerts, detections, hits, incidents,
signals usually mean *finding*, but sometimes literally mean *event*. A keyword map is a lookup table
that fails closed — any phrasing not in the table falls through unchanged. A description enriched with
this vocabulary generalises instead, because the model matches on meaning against text it already reads
on every turn, not against a fixed list. Seed each finding-hits tool's description with its synonym set;
the estimated cost is ~15 tokens per tool.

Neither change requires the new typed events tool that would fully close this gap (no typed tool reads
`wazuh-events-v5*` today) — that remains a separate, larger change (a new schema, ~400-500 tokens) and is
out of scope here. The negative-scope clause is what lets the model say "no rules fired; I did not check
raw events" in the meantime, instead of silently reporting nothing happened.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: the L08 phrasing above, **n=1 on each of two model
      families** (`openai/gpt-oss-120b`, `qwen/qwen3.6-27b`), both silently answering findings-only with
      no caveat
- [ ] `search_findings_by_agent` (and the other 7 finding-hits tools sharing `common.ts`'s baseline
      shape) state in `spec.description` that they cover rule-matched detections only
- [ ] The same descriptions carry the synonym vocabulary for agents and findings
- [ ] **N>=10** post-fix runs of the L08 phrasing and at least one paraphrase from a different register
      (e.g. "what happened on web-prod-01 in the last hour"), stating how many responses now name the
      rule-matched-only scope when the result is empty
- [ ] Colocated unit test asserting each affected tool's `spec.description` contains the negative-scope
      clause (new `*.test.ts` beside the affected catalog files, or an addition to `common.test.ts`)
- [ ] Regression gate: 16-question set passes; zero silent scope narrowing without a stated caveat

### Notes

- **Cheapest coverage change in this set.** ~15 tokens per tool description versus ~400-500 tokens for a
  new tool schema.
- Touches `server/tools/catalog/*.ts` (the findings family), the same files #8872 (artifact filters)
  also touches. Land this one first — it is smaller, independent of the new filter parameters, and does
  not change any tool's `parameters` schema, only `description`.
- Does not add a typed events tool. That remains a separate, costlier change; this issue only makes the
  existing gap speak rather than close it.
- The prompt-level instruction at `prompts.ts:48-51` is not being removed or weakened — it stays as a
  second line of defense. This issue adds the fix at the layer the model actually consults first when
  choosing which tool to call.

### LINKED ISSUE #8872: Findings tools cannot filter by IP, user or process
## Description

None of the findings tools accept an IP, username or process name as a filter parameter, even though
some of that data is already returned in the tool's own output. A threat hunter cannot ask "which
findings came from this IP" through a typed tool — the model either falls back to `search_wazuh_data`
and hand-writes OpenSearch DSL, or pulls a broad result and filters it by eye. This is a parameters
change, not a new-tool change: adding an optional filter to an existing tool costs roughly 20-40 tokens
per routed turn, against roughly 400-500 tokens for a whole new tool schema — and stage 2 was measured
at ~5,800 tokens with 12 schemas already offered, so the cheaper option matters.

### Evidence

**`source.ip` is already returned to the model but cannot be searched on.** It is one of the
`FINDING_DIGEST_EXTRA_COLUMNS` appended to every finding-hits tool's digest (`common.ts`), and the system
prompt tells the model as much (`server/prompts.ts:68-71`, verbatim): "For questions about WHICH users,
IPs, commands or programs were involved, prefer the typed finding tools: their results include
source.user.name, destination.user.name, source.ip and process.command_line." No finding tool's
`spec.parameters` accepts any of those four fields as an input. The model can see an IP in its results
and has no typed way to ask "show me more like this one."

**Measured persona impact.** Against a 40-question persona set, the threat-hunter persona has **12%**
typed coverage versus **62%** for Tier-1 triage (measured in the QA coverage pass) — the widest gap of any persona pair
measured, and it traces directly to this filter gap: "Which findings came from 198.51.100.10?", "show me
everything involving user root", "which findings involve this process?" all have a tool that reaches the
right index and no parameter to express the filter.

**The escape hatch already works around this, correctly, twice — this issue is about not needing it.**
Two hand-written `search_wazuh_data` queries were verified against an independent control: a `term` on
`source.ip` returned 46 hits, and a `should` over `source.user.name`/`destination.user.name` returned 74
hits (consistent with 36 + 64 − 26 overlap). The model recognised the missing filter and wrote correct
DSL both times. This is not a case of an unanswerable question — it is a case of the answer depending on
the model successfully hand-serialising a query every time, rather than on a schema that expresses the
filter directly.

**`process.name` is already used inside a query, just not exposed to the caller.**
`get_suspicious_powershell.ts` filters on `process.name` internally (a `terms` clause matching
`powershell.exe`/`pwsh.exe`/`powershell`) — with its own doc comment noting that whether `process.name`
is reliably populated on Wazuh 5.0 Windows agent data is unconfirmed. The field is queryable on
`wazuh-findings-v5*`; no tool lets the model supply its own value for it.

**`destination.ip` is confirmed mapped on the findings indices.** A live `_field_caps` check against
`wazuh-findings-v5*` (run 2026-08-05 during this QA pass) returned, across all eight backing indices:

```json
"destination.ip": {"ip": {"type": "ip", "searchable": true, "aggregatable": true}}
```

So a `term` filter on it is valid. Two supporting facts: `server/tools/privacy.ts`'s
`FIELD_POLICY_DEFAULTS` already carries `{ field: 'destination.ip', action: 'anonymize', kind: 'IP' }`
next to the `source.ip` entry, and the field is returned by `get_agent_ports` on the inventory index.
One residual caveat: the mapping existing does not guarantee documents populate it — unlike
`source.ip`, it is not in any finding-hits tool's digest today, so how often it carries a value in
real finding documents was not measured. A filter on a sparsely-populated field is still correct
(it narrows to the documents that have it); it is just worth knowing when interpreting results.

### Code references

- `plugins/wazuh-ai-assistant/server/tools/catalog/common.ts` — `FINDING_INVESTIGATION_ROW_FIELDS` /
  `FINDING_DIGEST_EXTRA_COLUMNS` (fields already returned) and `severityProperty()` /
  `timeRangeProperties()` (the pattern a new shared filter-property helper should follow)
- `plugins/wazuh-ai-assistant/server/tools/catalog/search-findings-by-agent.ts` — representative
  finding-hits tool with no artifact filter today
- `plugins/wazuh-ai-assistant/server/tools/catalog/get-suspicious-powershell.ts` — `process.name` already
  used in a `buildRequest` filter clause, confirming the field is queryable on `wazuh-findings-v5*`
- `plugins/wazuh-ai-assistant/server/prompts.ts:68-71` — the instruction telling the model these fields
  are output-available on finding tools, with no mention that they can be filtered on because they cannot
- `plugins/wazuh-ai-assistant/server/tools/privacy.ts:70-73` — `FIELD_POLICY_DEFAULTS` entries for
  `source.ip`, `destination.ip`, `source.user.name`, `destination.user.name`

### Suggested treatment

Add optional filter parameters to the finding-hits tools, following the existing `severityProperty()` /
`timeRangeProperties()` pattern in `common.ts` (a small shared helper returning the property definitions,
each becoming an additional `bool.filter` term/match clause in `buildRequest` when supplied):

- `source_ip` → `term` on `source.ip`
- `process_name` → `term`/`terms` on `process.name`
- `source_user_name` / `destination_user_name` → `term` on `source.user.name` / `destination.user.name`
- `destination_ip` → `term` on `destination.ip` (mapping confirmed via `_field_caps`, see Evidence;
  population rate in real finding documents unmeasured)

Start with `search_findings_by_agent`, the tool with measured evidence, then extend to the other
finding-hits tools sharing `common.ts`'s baseline shape, since they read the same index and have the same
gap.

No new tool, no new index, no guardrails change — every field involved is already inside
`wazuh-findings-v5*`, already allowlisted, and (except `destination.ip`) already returned to the model.
This is the same reasoning #8873 applies to inventory: prefer a parameter over a schema whenever the
data is already in reach.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: the two verified escape-hatch queries above (46 hits on
      `source.ip`, 74 hits on the user-name pair), stating that both required hand-written DSL because no
      typed parameter existed
- [ ] `search_findings_by_agent` accepts `source_ip`, `process_name`, `source_user_name` and
      `destination_user_name` as optional filters; each, when supplied, narrows the result the same way
      the equivalent hand-written DSL clause did
- [ ] `destination_ip` filter added (mapping already confirmed via `_field_caps`); note in the PR how
      often the field is populated in the seeded data so reviewers can interpret zero-hit results
- [ ] Persona coverage re-measured against the 40-question set: threat-hunter typed coverage rises from
      12% (state the new percentage)
- [ ] A tool call with no artifact filter supplied is unchanged (regression)
- [ ] Colocated unit test in `search-findings-by-agent.test.ts` (new file) covering each new filter
      parameter, alone and combined with the existing `severity`/time-range filters
- [ ] Regression gate: 16-question set passes

### Notes

- **Not an impossible question.** The model already answers two of these correctly via
  `search_wazuh_data`. This issue removes the reliance on that escape hatch working every time, not on a
  gap that has no answer.
- Shares `server/tools/catalog/common.ts` and the findings-tool family with #8871 (negative scope and
  synonym vocabulary in the same descriptions). The two are independent — #8871 changes
  `spec.description`, this one changes `spec.parameters` — and can land in either order, though #8871
  is smaller and has no schema risk, so doing it first is reasonable.
- All five fields are confirmed: four in this codebase's own field lists, `destination.ip` by live
  `_field_caps` (see Evidence).

### LINKED ISSUE #8873: Consolidate the four inventory tools into one with a `kind` enum
## Description

Four separate tools — `get_agent_os`, `get_agent_packages`, `get_agent_ports`, `get_agent_processes` —
each cover one of the 13 `wazuh-states-inventory-*` sub-indices, and each costs its own schema on every
routed turn. Replacing them with one tool, `get_agent_inventory(agent_id, kind, limit)`, is the best
token trade in this set: it takes the inventory category from **4 schemas to 1** — a net token
*reduction* per routed turn, not an increase — while coverage rises from 4 covered surfaces to all 13.
The nine currently uncovered surfaces include hotfixes, the one that correlates with
`get_vulnerabilities` and is invisible to the assistant today.

### Evidence

**The four tools, and the exact index each reads, verified in source:**

| Tool | Index |
|---|---|
| `get_agent_os` | `wazuh-states-inventory-system*` |
| `get_agent_packages` | `wazuh-states-inventory-packages*` |
| `get_agent_ports` | `wazuh-states-inventory-ports*` |
| `get_agent_processes` | `wazuh-states-inventory-processes*` |

Note `get_agent_os` reads the `system` index, not an `os`-named one — an enum value of `os` and an index
suffix of `system` would need to map to each other in the new tool, not be assumed identical.

**All four are registered individually in `server/tools/registry.ts`** (imports at lines 26-29, `CATALOG`
entries at lines 84-87), each contributing its own schema to every stage-2 tool offer that includes the
inventory category.

**13 real `wazuh-states-inventory-*` index-pattern constants are defined in this repo**, in the sibling
`main` plugin (`plugins/main/common/constants.ts:118-140`, `WAZUH_IT_HYGIENE_*`), where the IT Hygiene
dashboards already query all 13 live:

```
wazuh-states-inventory-hardware*
wazuh-states-inventory-hotfixes*
wazuh-states-inventory-interfaces*
wazuh-states-inventory-networks*
wazuh-states-inventory-packages*
wazuh-states-inventory-ports*
wazuh-states-inventory-processes*
wazuh-states-inventory-protocols*
wazuh-states-inventory-system*
wazuh-states-inventory-users*
wazuh-states-inventory-groups*
wazuh-states-inventory-services*
wazuh-states-inventory-browser-extensions*
```

Only 4 of these 13 have a `wazuh-ai-assistant` tool today. The other 9 — hardware, hotfixes, interfaces,
networks, protocols, users, groups, services, browser-extensions — have none. There is no mention of
"hotfix" anywhere in `plugins/wazuh-ai-assistant` (zero matches), while `get_vulnerabilities`,
`get_critical_vulnerabilities`, `get_vulnerabilities_by_agent` and `get_vulnerability_by_cve` all exist
and read `wazuh-states-vulnerabilities*`. A question like "which of these critical vulnerabilities have a
hotfix available" has a tool for one half of the correlation and none for the other.

**No guardrails change is needed to reach the missing 9.** `guardrails.ts`'s `INDEX_ALLOWLIST_RE`
already matches the whole `wazuh-states-*` prefix, so every `wazuh-states-inventory-*` sub-index —
covered or not — is already inside the allowed set.

**The token math, from the stage-2 cost measured elsewhere in this pass** (~5,800 tokens with 12 schemas
offered): collapsing 4 schemas into 1 removes roughly 3 schemas' worth of tokens from every routed turn
that includes the inventory category, while the number of reachable surfaces triples.

### Code references

- `plugins/wazuh-ai-assistant/server/tools/registry.ts:26-29,84-87` — the four imports and `CATALOG`
  entries to replace
- `plugins/wazuh-ai-assistant/server/tools/catalog/get-agent-os.ts`,
  `get-agent-packages.ts`, `get-agent-ports.ts`, `get-agent-processes.ts` — the four tools to fold in;
  each already shares `validateAgentId`/`clampLimit`/`limitProperty`/`objectSchema` from `common.ts`, and
  each doc comment records the live-mapping verification already done for its own index
- `plugins/wazuh-ai-assistant/server/tools/guardrails.ts:86-87` — `INDEX_ALLOWLIST_RE`, confirming
  `wazuh-states-*` (and therefore all 13 inventory sub-indices) is already allowed
- `plugins/main/common/constants.ts:118-140` — the 13 confirmed `WAZUH_IT_HYGIENE_*` index-pattern
  constants, verified live by the `main` plugin's own IT Hygiene dashboards

### Suggested treatment

Replace the four tools with one:

```
get_agent_inventory(agent_id, kind: enum[os|packages|ports|processes|hardware|hotfixes|
                    services|users|interfaces|networks|protocols|groups|browser_extensions], limit)
```

`kind` resolves to the matching `wazuh-states-inventory-<kind>*` index, with the one naming exception
noted above (`os` → `system*`). Recommend two steps:

1. **Fold the 4 already-built kinds in as a pure refactor.** `os`, `packages`, `ports` and `processes`
   keep their existing `_source`/`tableSpec`/`digest` field lists unchanged, copied from the four current
   tool files. This alone drops the schema count from 4 to 1 with no coverage change and no new-mapping
   risk — worth doing on its own regardless of how quickly the remaining 9 kinds follow.
2. **Add the remaining 9 kinds one at a time**, each verified against a live 5.0 mapping before shipping
   — the same check each of the four existing tools' doc comments record having already done for their
   own index. This codebase has already hit a case where a field did not survive the 4.14→5.0 port
   unchanged (`get-agent-processes.ts` documents dropping the process-owner column entirely rather than
   faking it); the same care applies to the 9 new surfaces.

Prioritise `hotfixes` among the 9: it is the one that pairs with the existing vulnerability tools, and it
closes the sysadmin/patch-management gap most directly.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: 4 separate schemas offered for inventory today, and zero
      tool coverage for hotfixes/hardware/services/users/interfaces/networks/protocols/groups/
      browser-extensions (confirmed by source inspection — no runtime reproduction needed for an absent
      tool)
- [ ] `get_agent_inventory` replaces all four in `server/tools/registry.ts`; the four individual tool
      files are removed (nothing outside `registry.ts` references them by name)
- [ ] Schema count for the inventory category drops from 4 to 1 — assert this on the outbound tool list
      in a test
- [ ] The 4 previously-covered kinds (`os`, `packages`, `ports`, `processes`) return the same fields for
      the same `agent_id`/`limit` inputs as today's individual tools (regression)
- [ ] At least the `hotfixes` kind is added, with its `_source`/table/digest fields verified against a
      live 5.0 mapping, not assumed from the 4.14 equivalent
- [ ] **N>=10** post-change runs across the previously-covered kinds, confirming routing still reaches the
      right kind for each
- [ ] Colocated unit test `get-agent-inventory.test.ts` covering the `kind` enum, rejection of an invalid
      `kind`, and reuse of `validateAgentId`
- [ ] Regression gate: 16-question set passes

### Notes

- **Best token trade in this batch**: coverage goes from 4 to 13 surfaces at a *negative* marginal schema
  cost — fewer schemas offered per turn than today, not more.
- Touches `server/tools/registry.ts`, `server/tools/catalog/common.ts` **and
  `server/tools/router.ts`**: `TOOL_CATEGORY` maps each of the four tool names to the `inventory`
  category, and `assertRegistryConsistency()` runs at module load and **throws on either direction of
  drift** — a registry tool with no category entry, or a stale entry naming a removed tool. So the four
  old entries must be removed and `get_agent_inventory: 'inventory'` added in the same change, or the
  plugin fails to start. This is deliberate fail-loud behaviour working as designed; noted so the
  implementer expects it rather than debugs it.
- Shared with no other issue in this batch; can proceed independently of #8867, #8871 and #8872.
- Precedent: #8831 already performed the same consolidation pattern on the agent-status tools
  (`get_active_agents` + `get_disconnected_agents` + 2 statuses → `get_agents`). This issue applies
  the accepted pattern to the four inventory tools, which #8831 did not touch.
- The 9 new kinds each need their own live-mapping verification pass before shipping — do not assume a
  4.14 field name or shape survives into the 5.0 index unchanged.
- No relation to `plugins/main`'s IT Hygiene indices beyond reading the same data: `plugins/main` is cited
  here only as independent, already-live confirmation that all 13 index names are real; this issue makes
  no change to that plugin.

yenienserrano:
## Description

Closes #8871
Closes #8872
Closes #8873

Three cheap, additive improvements to the AI Assistant's tool catalog, all confined to
`plugins/wazuh-ai-assistant/server/tools/`:

- **#8871** — The 8 finding-hits tools' descriptions did not state that they cover
  rule-matched detections only, so the model silently substituted a narrower tool for
  broader questions ("everything that happened on this host") and reported nothing
  happened instead of saying it hadn't checked raw events. Also seeds analyst-vocabulary
  synonyms (host/machine/endpoint for agent; alert/detection/hit/signal for finding).
- **#8872** — The same finding-hits tools had no way to filter by IP, user, or process,
  forcing a hand-written `search_wazuh_data` query for common threat-hunter questions
  ("which findings came from this IP"). Adds five optional artifact-filter parameters
  (`source_ip`, `destination_ip`, `process_name`, `source_user_name`,
  `destination_user_name`) shared across all 9 finding-hits tools reading `wazuh-findings-v5*`.
- **#8873** — `get_agent_os`, `get_agent_packages`, `get_agent_ports` and
  `get_agent_processes` each cost their own schema on every routed turn for one
  `wazuh-states-inventory-*` sub-index. Consolidates them into one
  `get_agent_inventory(agent_id, kind, limit)` tool, dropping the inventory category's
  schema count from 4 to 1 while adding `hotfixes` as a 5th kind (pairs with the existing
  vulnerability tools for patch-management questions). A follow-up commit also accepts an
  optional `agent_name` (matched against `wazuh.agent.name`) alongside `agent_id`, since a
  live run showed the tool never got called for deictic phrasing ("this server") when
  `agent_id` was the only way in.

## Proposed Changes

- `common.ts`: adds shared `findingArtifactFilterProperties()` /
  `findingArtifactFilterClauses()` (artifact filters) and `FINDING_SCOPE_NOTE` /
  analyst-vocabulary constants (negative scope), following the existing
  `severityProperty()` / `timeRangeProperties()` pattern.
- 8 finding-hits tools (`search_findings_by_agent`, `get_critical_findings`,
  `get_findings_by_time`, `get_brute_force`, `get_suspicious_powershell`,
  `search_findings_by_multiple_agents`, `search_findings_by_rule_title`,
  `search_findings_by_rule_tag`, `search_findings_by_os`): wire in the artifact filters
  and the negative-scope/vocabulary description text. Also state a
  current-state-only note on the 4 vulnerability tools.
- `get-agent-os.ts`, `get-agent-packages.ts`, `get-agent-ports.ts`,
  `get-agent-processes.ts`: removed, replaced by `get-agent-inventory.ts`
  (`get_agent_inventory(agent_id, kind, limit)`, kind enum: `os | packages | ports |
  processes | hotfixes`). `agent_id` OR `agent_name` is now accepted (previously
  `agent_id` alone was required); `agent_id` wins when both are given.
- `registry.ts` / `router.ts`: swap the 4 old tool registrations for
  `get_agent_inventory`, and update `TOOL_CATEGORY` accordingly (required — the registry
  throws on module load otherwise).
- `privacy.ts`: allow-lists inventory identity fields (`package.name/version/type`,
  `package.hotfix.name`, `host.os.name/version/platform`, `process.name`,
  `network.transport`, `interface.state`, `source.port`, `destination.port`) so
  `get_agent_inventory`'s `deriveColumns: true` doesn't anonymize software/config
  identity under privacy mode. Infrastructure/personal identifiers (`host.hostname`,
  `process.command_line`, IPs, usernames) keep anonymizing.

### Results and Evidence

No UI change — server-side tool catalog only. Verified via the colocated unit tests
below and `yarn typecheck` (no new errors; the plugin's pre-existing typecheck failures
are all in files this PR does not touch).

### Artifacts Affected

None (no config/executable/package changes).

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `finding-scope.test.ts` — negative-scope clause and vocabulary present on the affected
  finding-hits/vulnerability/inventory tool descriptions.
- `search-findings-by-agent.test.ts` — artifact-filter parameters resolve to the expected
  `bool.filter` clauses, and an unsupplied filter is a no-op (regression).
- `get-agent-inventory.test.ts` — kind enum, rejection of an invalid kind,
  `validateAgentId` reuse, `agent_id`/`agent_name` precedence, limit clamping, and
  regression assertions that each folded-in kind (`os`, `packages`, `ports`,
  `processes`) matches its now-deleted tool's original `_source`/index/size.
- `privacy.test.ts` — `applyFieldPolicy` against the real `FIELD_POLICY_DEFAULTS` per
  `get_agent_inventory` kind: packages/os/processes identity stays readable while
  `host.hostname`/`process.command_line`/IPs/usernames keep anonymizing.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


rodrigofez (approved):
## Review test report

Tested on a live dev stack (dashboard, indexer, agents `001` and `002`), commit `829da21b0`.

**How I tested.** I used the repo's scripted test provider (`eval/mock_provider.js`) rather than a real model, so tool arguments are exact and repeatable. For each call I compared the query the tool ran, the result sent to the provider, and a direct query against the same index. Some fields and one index were empty here, so I added 11 temporary documents to test them, then deleted them all — details at the end.

| Issue | Behaviour | Acceptance criteria |
| --- | --- | --- |
| 8873 — new `get_agent_inventory` tool | Works, all five kinds | 2 of 9 not met |
| 8873 — privacy allow-list | **Does not hold** (D1, D2, D3) | — |
| 8872 — new filters | Works | 3 of 8 not met |
| 8871 — description text | Present on all 8 required tools; effect untested | 3 of 7 not met |

Unit tests: the 4 new files pass (74 tests). Whole plugin passes (61 suites, 786 tests).

---

## What works

**The inventory tool.** For `os`, `packages`, `ports` and `processes` on agent 001, it queries the same index with the same fields, size and sort as the four tools it replaces, and returns rows identical to a direct indexer query. Counts matched exactly: 1 / 216 / 9 / 15. `agent_id` correctly wins over `agent_name`. `limit` clamps at 3, 9999, 0, -5, and is ignored for `os` as documented. An invalid `kind` and a missing agent both give clear errors.

**The `hotfixes` kind.** That index is empty here, so I added 3 documents for agent 001. The tool returned all 3 with the right values, and 0 for agent 002, so the agent filter works. The index has a strict mapping and accepted `package.hotfix.name`, which confirms the field exists in the live 5.0 mapping rather than being carried over from 4.14.

**The new filters.** All five fields were empty here, so I added 7 findings: one per filter, plus one carrying all five values.

| Filter | Expected | Got |
| --- | --- | --- |
| none | all rows | 5103 |
| `source_ip` = .10 | its row + the all-five row | 2, correct rows |
| `source_ip` = .11 | its row only | 1, correct row |
| `destination_ip` | its row + the all-five row | 2, correct rows |
| `process_name` | its row + the all-five row | 2, correct rows |
| `source_user_name` | its row + the all-five row | 2, correct rows |
| `destination_user_name` | its row + the all-five row | 2, correct rows |
| `source_ip` = .10 **and** `process_name` | the all-five row only | 1, correct row |
| `source_ip` = .11 **and** `process_name` | nothing — .11 has no process | 0 |
| all five together | the all-five row only | 1, correct row |
| a value that matches nothing | nothing | 0 |

Each filter targets the right field, filters combine as AND rather than OR, and leaving one out changes nothing.

**Most of the privacy allow-list.** OS name, version and platform, package name and architecture, network transport, interface state, process name and hotfix name all reach the provider readable. Hostname, IPs and command line stay hidden.

---

## Acceptance criteria

I checked the PR against the criteria written in each issue, not just against the PR description.

### 8871

| Criterion | Status |
| --- | --- |
| Before-fix reproduction on two model families | Recorded in the issue |
| The 8 named tools state rule-matched-only scope | Met — all 8 |
| Those descriptions carry agent and finding synonyms | Partly. Finding synonyms yes, inside the shared clause. Agent synonyms are on 3 of the 8; the other 5 do not mention agents at all, so this may be fine — worth confirming |
| Unit test asserting the clause | Met |

### 8872

| Criterion | Status |
| --- | --- |
| Before-fix reproduction (46 and 74 hits via hand-written queries) | Recorded in the issue |
| The four filters narrow the same way the hand-written query did | Met — verified live, see the table above |
| `destination_ip` added, **and the PR states how often the field is populated** | Filter added. **The note is missing.** For the record: 0 of 5256 findings here carry it — and the same is true of the other four fields |
| **Threat-hunter coverage re-measured against the 40-question set, new percentage stated** | **Not done.** Nothing in the PR |
| A call with no filter is unchanged | Met |
| Unit test per filter, alone and combined with severity and time range | Met — 12 tests, including the combinations |

### 8873

| Criterion | Status |
| --- | --- |
| Before-fix reproduction by source inspection | Met |
| Replaces all four in the registry, the four files removed, **nothing outside the registry names them** | **Not met** — five files still name them, see D7 |
| **A test asserting the inventory schema count drops from 4 to 1** | **Not done.** The drop happens, but no test asserts it. `router.test.ts` has four tests and none cover this |
| The four kinds return the same fields for the same inputs | Met — verified live and by unit test |
| `hotfixes` added, verified against a live 5.0 mapping | Met — the strict mapping accepted `package.hotfix.name` |
| Unit test for the enum, an invalid kind, and agent-ID reuse | Met |

Two cross-cutting points:

- **All three issues ask for a CHANGELOG entry linking the issue.** The PR carries `no-changelog`. One of the two is wrong — worth settling before merge.
- **8873 says coverage goes from 4 surfaces to 13.** This PR delivers 5. Its criteria only require `hotfixes`, so the criterion is met, but the PR says "Closes #8873" and the remaining 8 surfaces have no follow-up issue that I can see.

---

## Defects

D1, D2 and D4 reproduce using the plugin's own code — no HTTP, no test provider, no saved objects. See **Reproduce it** at the end.

### D1 (High) — Package versions are hidden, even though they are on the allow-list

`package.version` is listed as allowed, but the provider gets a placeholder:

```
{"package.name":"acl","package.version":"HOST_1","package.architecture":"x86_64","package.vendor":"VAL_1"}
                       ^^^^^^ real value is "2.2.53-1.el8"
```

The allow-list is not the problem. A later step scans the result for hostnames using a `word.word` pattern, and only skips text made purely of digits and dots. `2.2.53-1.el8` has letters and a hyphen, so it is treated as a hostname and replaced.

**212 of the 216 package versions on agent 001 (98%)** are affected. Only pure-numeric versions survive.

So "what packages are installed here" answers with real names and placeholder versions. That defeats the patch-management use case this PR gives as the reason for adding the `hotfixes` kind — the version is the answer there.

OS versions happen to be safe here (`8.5.2111` is digits and dots), but any OS version with a suffix would hit the same thing.

**Fix:** skip the hostname scan for values from fields the policy already allows, or widen the "this is a version" exception to cover normal package version shapes.

### D2 (High) — An allowed value is wiped if the same text appears in a hidden field

On agent 002, process 1 has `name` = `tail` and `command_line` = `tail`. The command line is hidden, so `tail` gets a placeholder — and the replacement then rewrites it everywhere, including in `process.name`, which was supposed to stay readable:

```
{"process.pid":1,"process.name":"VAL_2","process.state":"VAL_1",
 "process.parent.pid":0,"process.command_line":"VAL_2"}
```

The replacement is a plain "find this text, swap it" across the whole result, with no minimum length and no word boundary. The same cause makes a one-letter value corrupt existing placeholders: `process.state` = `S` turns `HOST_1` into `HOVAL_1T_1`, which can no longer be turned back into the real value.

**Fix:** do not re-scan a value the policy already allowed, and give the replacement a minimum length or a word boundary.

D1 and D2 both sit in code older than this PR. But this PR is the first to promise these fields stay readable, and this is why they do not.

### D4 (Medium) — Four fields the tool reads are missing from the allow-list

All four confirmed live:

| Field | Value | Sent as |
| --- | --- | --- |
| `host.os.full` | `Ubuntu 24.04.4 LTS (Noble Numbat)` | `VAL_1` |
| `host.architecture` | `x86_64` | `VAL_2` |
| `package.vendor` | `CentOS` | placeholder |
| `process.state` | `S` | placeholder |

`host.os.full` is the readable OS name, so "what OS does this host run" comes back as a placeholder. That field is empty on both real agents here, so I tested it with a temporary document.

**Fix:** add all four. They are the same kind of software information as the entries already there.

Minor: `process.pid`, `process.parent.pid`, `source.port` and `destination.port` are numbers, and numbers are never replaced, so the two port entries in the allow-list currently do nothing.

### D5 (Medium) — A bad IP value returns zero rows and no error

The two IP filters go into the query unchecked. If the value is not an IP address, one shard fails but the search still succeeds with zero hits, so the error never surfaces and the model is told "0 results". I saw this live with `source.ip: "not-an-ip"`.

This is the same failure 8871 is trying to remove in this same PR — reporting "nothing found" when nothing was checked.

**Fix:** validate the two IP values and raise a clear error, as the agent ID already does.

### D6 (Low) — `search_findings_by_os` is inconsistent

It got the new filters but not the new scope text.

**8871 does not require it.** That issue names eight tools and this is not one of them, and all eight required tools have the text. So this is not a missed criterion.

It is still worth a look. The PR description lists nine tools as getting the text, including this one, so the description and the code disagree. And the tool now takes the artifact filters while saying nothing about what it covers.

**Fix:** add the text and put the tool in the test list, or correct the PR description.

### D7 (Low, but it is a written criterion) — The evaluation scripts still name the four deleted tools

8873 asks that nothing outside the registry reference them. Five files still do:

`eval/run_plumbing.js` (lines 62, 101-104, 151-154, 193-196), `eval/run_tool_matrix.js` (102-105, 277-307), `eval/measurement-set.json` (246, 254, 262, 270), `eval/README.md` (231-238), `server/tools/field-policy-coverage.test.ts` (216).

Those names now return "Unknown tool", so the plumbing script fails on this branch.

### D8 (Low) — Worth confirming

The new allow-list entries are plain field names, not tied to one tool, so they also apply to the free-search and document-lookup tools. Fields like `process.name` and the OS fields used to be hidden there and now are not. That may be intended, but the PR does not say so.

---

## Not verified

**Whether the description text changes which tool the model picks.** The test provider ignores descriptions, and no model available here is stable enough to attribute a change to the edit. This is the 8871 criterion asking for 10 or more runs, and the 8873 criterion asking for 10 or more routing runs. Both need a real model.

Unrelated note: `search_findings_by_os` cannot match anything in this deployment, because `host.os.name` is empty on all 5256 findings. That is a data gap here, not a problem with the PR.

---

## Test data I added and removed

| Index | Added | Purpose | Removed |
| --- | --- | --- | --- |
| `wazuh-findings-v5-security` | 7 findings | The five filter fields were empty | yes |
| `wazuh-states-inventory-hotfixes` | 3 hotfixes | The index was empty | yes |
| `wazuh-states-inventory-system` | 1 host | `host.os.full` was empty | yes |

All 11 were tagged so they could be found again, and all were deleted. Counts are back to where they started: findings 5256, hotfixes 0, system 2, packages 311.

---

## Reproduce it

D1, D2 and D4 are in the plugin's own code. A tool result reaches the provider in three steps:

1. Apply the field policy — `server/tools/executor.ts:46-68`
2. Turn it into JSON
3. Scan the JSON for hostnames, then swap known text — `server/routes/chat.ts:347-350`

Step 3 undoes step 1.

- **D1:** `server/tools/privacy.ts:478` replaces anything matching `word.word`. Its only exception is line 442, `/^[0-9.]+Z?$/` — digits and dots only.
- **D2:** `server/tools/privacy.ts:297` — `out.split(value).join(placeholder)`. Plain text swap, no minimum length, no word boundary.

Drop the file below in as `server/tools/pr8887-proof.test.ts` and run `yarn test:jest --testPathPattern=pr8887-proof`. It uses this branch's own `FIELD_POLICY_DEFAULTS` and the same two steps `chat.ts` uses. All checks pass, which means the defects are present.

<details>
<summary>pr8887-proof.test.ts</summary>

```ts
import assert from 'node:assert/strict';
import {
  FIELD_POLICY_DEFAULTS,
  Pseudonymizer,
  applyFieldPolicy,
  prescanAndMintToolContent,
} from './privacy';
import { Digest } from './digest';

const TOOL = 'get_agent_inventory';
// get-agent-inventory.ts sets `deriveColumns: true`, so executor.ts passes isEscapeHatch = true.
const IS_ESCAPE_HATCH = true;

/** The exact two steps chat.ts applies to a tool message, in order. */
function whatTheProviderReceives(digest: Digest, pz: Pseudonymizer): string {
  const scrubbed = applyFieldPolicy(
    digest, FIELD_POLICY_DEFAULTS, pz, undefined, TOOL, IS_ESCAPE_HATCH,
  );
  return pz.applyToText(prescanAndMintToolContent(JSON.stringify(scrubbed), pz));
}

test('D1: package.version is allowed but the provider still gets a placeholder', () => {
  assert.deepEqual(
    FIELD_POLICY_DEFAULTS.find(e => e.field === 'package.version'),
    { field: 'package.version', action: 'allow' },
  );
  const digest: Digest = {
    tool: TOOL,
    counts: { total: 216, returned: 50, truncated: true },
    samples: [{
      'package.name': 'acl',
      'package.version': '2.2.53-1.el8',
      'package.architecture': 'x86_64',
      'package.vendor': 'CentOS',
    }],
    columns: ['package.name', 'package.version', 'package.architecture', 'package.vendor'],
  } as Digest;

  const pz = new Pseudonymizer([]);
  const sent = whatTheProviderReceives(digest, pz);
  console.log('\nD1 — provider receives:\n  ' + sent);
  console.log('  placeholders: ' + JSON.stringify(pz.newEntries()));

  assert.ok(sent.includes('"package.name":"acl"'));
  assert.ok(!sent.includes('2.2.53-1.el8'), 'expected the real version to be gone');
  assert.match(sent, /"package\.version":"HOST_\d+"/);
});

test('D1b: real package versions are affected; pure-numeric ones are not', () => {
  const real = [
    '2.2.53-1.el8', '3.0-0.17.20191104git1c2f876.el8', '8-3.el8',
    '5.3.28-42.el8_4', '0.63.0-3.el8',
  ];
  const hidden = real.filter(v => {
    const pz = new Pseudonymizer([]);
    const d: Digest = {
      tool: TOOL, counts: { returned: 1, truncated: false },
      samples: [{ 'package.version': v }], columns: ['package.version'],
    } as Digest;
    return !whatTheProviderReceives(d, pz).includes(v);
  });
  console.log('\nD1b — hidden: ' + JSON.stringify(hidden));
  assert.equal(hidden.length, real.length);

  const pz = new Pseudonymizer([]);
  const ok: Digest = {
    tool: TOOL, counts: { returned: 1, truncated: false },
    samples: [{ 'package.version': '8.5.2111' }], columns: ['package.version'],
  } as Digest;
  assert.ok(whatTheProviderReceives(ok, pz).includes('8.5.2111'));
});

test('D2: an allowed value is wiped when the same text is in a hidden field', () => {
  assert.deepEqual(
    FIELD_POLICY_DEFAULTS.find(e => e.field === 'process.name'),
    { field: 'process.name', action: 'allow' },
  );
  assert.equal(
    FIELD_POLICY_DEFAULTS.find(e => e.field === 'process.command_line')?.action,
    'anonymize',
  );
  // Real row: agent 002, pid 1 — name and command_line are both "tail".
  const digest: Digest = {
    tool: TOOL,
    counts: { total: 15, returned: 15, truncated: false },
    samples: [{
      'process.pid': 1, 'process.name': 'tail', 'process.state': 'S',
      'process.parent.pid': 0, 'process.command_line': 'tail',
    }],
    columns: ['process.pid', 'process.name', 'process.state',
              'process.parent.pid', 'process.command_line'],
  } as Digest;

  const pz = new Pseudonymizer([]);
  const sent = whatTheProviderReceives(digest, pz);
  console.log('\nD2 — provider receives:\n  ' + sent);

  assert.ok(!sent.includes('"process.name":"tail"'), 'expected process.name to be wiped');
  assert.match(sent, /"process\.name":"VAL_\d+"/);
});

test('D2b: a one-letter value corrupts an existing placeholder', () => {
  const digest: Digest = {
    tool: TOOL, counts: { returned: 1, truncated: false },
    samples: [{ 'host.hostname': 'web-prod-01', 'process.state': 'S' }],
    columns: ['host.hostname', 'process.state'],
  } as Digest;

  const pz = new Pseudonymizer([]);
  const sent = whatTheProviderReceives(digest, pz);
  console.log('\nD2b — provider receives:\n  ' + sent);

  assert.ok(!/HOST_\d+/.test(sent), 'expected the placeholder to be corrupted');
  assert.match(sent, /HOVAL_\d+T_\d+/);
});

test('D3: four fields the tool reads are absent from the allow-list', () => {
  const listed = new Set(FIELD_POLICY_DEFAULTS.map(e => e.field));
  for (const f of ['host.os.full', 'host.architecture', 'package.vendor', 'process.state']) {
    assert.equal(listed.has(f), false, `${f} unexpectedly listed`);
  }
});
```

</details>

Output:

```
D1  — {"package.name":"acl","package.version":"HOST_1","package.architecture":"x86_64","package.vendor":"VAL_1"}
      placeholders: [{"value":"CentOS","pseudonym":"VAL_1"},{"value":"2.2.53-1.el8","pseudonym":"HOST_1"}]
D1b — hidden: ["2.2.53-1.el8","3.0-0.17.20191104git1c2f876.el8","8-3.el8","5.3.28-42.el8_4","0.63.0-3.el8"]
D2  — {"process.pid":1,"process.name":"VAL_2","process.state":"VAL_1","process.parent.pid":0,"process.command_line":"VAL_2"}
D2b — {"host.hostname":"HOVAL_1T_1","process.state":"VAL_1"}

Tests: 5 passed, 5 total
```

PR #8888, opened 2026-08-06
Add Discover handoff and fix AI usage accounting

### LINKED ISSUE #8874: Offer a Discover query when the assistant cannot answer
## Description

When the assistant cannot answer a question — the index is outside what the model is allowed to read,
or no tool covers the question — the user gets prose or nothing. The machinery to hand them a working
query instead already exists in the codebase; it is just wired to the wrong trigger, so it never
appears on the failure path where it would matter most.

### Evidence

`common/discover-url.ts` already exports `buildDiscoverUrl` and `extractTimeRange`, and
`discover-link.tsx` already resolves an index name to an index-pattern saved-object id and renders an
"Open in Discover" button. Today that affordance is attached to `TableSpec.discover` — so it only
renders when a tool call **succeeded** and produced a table. The one case it cannot help with is the
one it would help with most: a question a tool could not, or was not allowed to, answer.

Three concrete surfaces are permanently out of reach for this reason. `server/tools/guardrails.ts`'s
`INDEX_ALLOWLIST_RE` (`^wazuh-(events-v5|findings-v5|states|threatintel-(rules|decoders|integrations|
policies|filters|kvdbs))[^,\s]*$|^\.opensearch-sap-detectors-config$`) does not match
`wazuh-threatintel-enrichments-a` (excluded by name, per that file's own comment: "this plugin has NO
tool touching `wazuh-threatintel-enrichments`"), nor `wazuh-metrics-*`, nor `wazuh-active-responses-*`
— none of the three alternatives in the regex cover them. So the model cannot query any of these three
index families under any circumstance, typed tool or escape hatch.

**That restriction is exactly where a handoff is safe to offer.** The allowlist restricts what the
**assistant** can read. It has no bearing on what the **user** can read — Discover runs under the
user's own RBAC, same as it does today for every other index. So a handoff to Discover for one of
these three surfaces serves the user data they are already entitled to see, through a route the model
itself cannot use to see it. Security posture is unchanged either way: the guardrail keeps doing its
job for the model, and the user's own permissions keep doing theirs for Discover.

### Code references

- `plugins/wazuh-ai-assistant/common/discover-url.ts` — `buildDiscoverUrl` and `extractTimeRange`,
  both already dependency-free and in `common/`
- `plugins/wazuh-ai-assistant/public/components/chat/discover-link.tsx` — `createDiscoverUrlResolver`
  and the index-pattern lookup; currently only invoked with a `TableSpec`
- `plugins/wazuh-ai-assistant/common/types.ts` — `TableSpec.discover` (only ever set for Indexer-backed
  tables, per its own doc comment) and the `StreamEvent` union, which has no failure-path variant today
- `plugins/wazuh-ai-assistant/server/tools/guardrails.ts` — `INDEX_ALLOWLIST_RE`, and the comment
  documenting `wazuh-threatintel-enrichments` as deliberately excluded

### Suggested treatment

Add a new `suggested_query` variant to the `StreamEvent` union: `{ type: 'suggested_query'; index:
string; dsl: Record<string, unknown>; reason: string }`, carrying no table. Render it as a callout with
an "Open in Discover" action, reusing `buildDiscoverUrl`/`extractTimeRange` **unchanged** — both already
live in `common/`, so nothing crosses the `public`/`server` boundary that doesn't already. The server
emits it whenever the model recognizes it cannot answer from an allowed index, or whenever a tool result
indicates the honest answer is out of scope. The bulk of the plumbing already exists; the actual new
work is the event shape, a small renderer, and the prompt/orchestration logic that decides when to emit
it.

**The wording matters as much as the mechanism.** Whatever sentence introduces the handoff must state
what the data contains and then point somewhere real — never that a feature is missing or "not yet"
available, since that reads as an unfinished product rather than a tool with a defined scope:

| Reads as unfinished | Reads as a stated boundary |
|---|---|
| "Triage state isn't available yet — it's on the roadmap." | "Wazuh doesn't record who acknowledged a finding, so I can't tell you. If your team tracks that in a ticketing system, that's where it will be." |
| "I can't see raw events yet." | "I read rule-matched findings. Raw events aren't in my toolset — here's the query, open it in Discover." |
| "Threat-intel lookup isn't implemented." | "Indicator enrichment isn't one of the sources I can query. Here's that IP in Discover against the enrichment index." |
| "I don't support ranges over 90 days." | "I query up to 90 days. Over that window: … For the full 12 months, open this in Discover." |

Every right-hand sentence explains the boundary as a fact about the data and ends with a route forward.
The word "yet" and any mention of a roadmap should not appear in an answer.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: confirm today's response for a question touching one of
      the three guardrail-blocked surfaces (`wazuh-threatintel-enrichments-a`, `wazuh-metrics-*`,
      `wazuh-active-responses-*`) carries no query-handoff affordance of any kind
- [ ] `suggested_query` added to the `StreamEvent` union in `common/types.ts`, with no `table` field
- [ ] A renderer shows the callout and a working "Open in Discover" link built from the unchanged
      `buildDiscoverUrl`/`extractTimeRange` helpers
- [ ] The introducing sentence follows the stated-boundary wording rule above (no "yet", no "roadmap")
- [ ] Each of the three guardrail-blocked surfaces produces a `suggested_query` pointing at the correct
      index once the model states the boundary, verified over **N>=10** runs per surface
- [ ] Colocated unit test asserting the new `StreamEvent` shape and that a `suggested_query` event never
      also carries a `table`
- [ ] Regression gate: 16-question set passes; zero blank answers, zero raw upstream error text

### Notes

- This is Track C ("graceful failure") of the coverage plan, and is designed to run in parallel with the
  coverage-parameter work (new filter params, the inventory-tool consolidation) without file overlap —
  it touches `common/types.ts`, a new renderer component, and the orchestration prompt, not the tool
  catalog.
- Pairs with, but does not require, the findings-tools negative-scope wording and the 90-day-clamp
  honesty fix — those two make the model *say* it hit a boundary; this issue makes that statement end in
  something clickable. Land the wording work first if sequencing matters, since this issue's prompt work
  depends on the model reliably stating the boundary in the first place.
- No new tool, no new architecture, no guardrail change: this is entirely reuse of `common/` code plus
  one new stream-event variant.
- File adjacency: #8853 (merged) reworked `message-bubble.tsx` rendering — the new callout renderer
  should expect a small rebase against that shape; no functional overlap.
- **A suggested query cannot be pre-validated by executing it** — for the guardrail-blocked indices,
  not running the query is the whole point. And this QA pass measured the model inventing field names
  in hand-written DSL (`agent.name.keyword` — matched 229 docs, returned zero buckets, no error). A
  suggested query with an invented field would silently show the user zero hits **in Discover**,
  moving the silent-wrong-answer defect from chat into the handoff. Mitigation: validate the suggested
  DSL's field names against `_field_caps` before emitting the event — a mappings/metadata read, not a
  document read, so it does not defeat the allowlist's purpose; whether the executor's allowlist should
  carve out metadata reads for this, or the suggestion path does its own scoped lookup, is an
  implementation decision to record in the PR. Alternatively (weaker), emit the link with only the
  index and time range and no field filters. Do not skip this — an unchecked handoff query inherits the
  exact defect #8868 exists to close.

### LINKED ISSUE #8875: `usage` reports the last upstream call only, understating tokens ~2x
## Description

A single question causes three sequential calls to the model provider (stage-1 router, stage-2 tool
call, tool-result summarization). The orchestration loop only forwards the token usage from the
**last** of those calls to the client — usage from every earlier round is discarded. Any cost dashboard,
quota display, or chargeback figure built on the `done` event's `usage` will understate real spend by
roughly half.

### Evidence

Measured: the plugin reported `inputTokens: 6409` for a question that actually consumed **~12,740**
input tokens across its three upstream calls — per-call sizes confirmed independently by Groq's own
rate-limit refusals, which state the exact token count of the request they rejected (`Requested 5808`
for the stage-2 call, `Requested 6179` for the tool-result call).

The mechanism: `server/routes/chat.ts`'s round loop suppresses every intermediate round's `done` event
without yielding it — when a round produced a tool call, its `done` is discarded with a bare `break`
(the branch at `if (sawToolCall) { ... break; }`) so the loop can continue with grown message history.
Only the final round's `done` — the one that actually reaches `yield event;` — is ever forwarded to the
client, carrying only that round's `usage`.

Each individual adapter call is *not* the bug: `openai-compatible.ts` yields exactly one `done` with the
`usage` its own upstream stream carried, once per `chatStream` invocation, and `anthropic.ts` does the
same. The bug is one level up, where three such calls are stitched into one turn and only the last
survives.

### Code references

- `plugins/wazuh-ai-assistant/server/routes/chat.ts:939-967` — the round loop's `done` handling; the
  `break` inside the `if (sawToolCall)` branch (around line 950-953) suppresses every non-final round's
  `done` (and its `usage`) without forwarding it; `yield event;` (line 966) is the only point a `done`
  event ever reaches the client
- `plugins/wazuh-ai-assistant/server/providers/openai-compatible.ts:146-159` — yields `done` with
  `usage: { inputTokens: parsed.usage.prompt_tokens, outputTokens: parsed.usage.completion_tokens }`
  once per call; correct for that one call
- `plugins/wazuh-ai-assistant/server/providers/anthropic.ts:192-200` — same per-call shape
  (`message_delta` carrying `usage.input_tokens`/`usage.output_tokens`)

### Suggested treatment

Accumulate usage in the round loop itself, not in either adapter. Keep a running total in
`server/routes/chat.ts`, add each round's `usage` to it before that round's `done` is suppressed, and
emit the sum on the final `done` that actually reaches the client. Both adapters already report correct
per-call usage in the same shape (`inputTokens`/`outputTokens`), so the accumulation logic reads that
shape once and needs no adapter-specific handling — no change to `openai-compatible.ts` or
`anthropic.ts` is required for this fix.

**One call lives outside the round loop and must not be forgotten:** the stage-1 router call is a
separate adapter stream read (the `route_question` pre-step), whose `done` — and usage — is consumed
inside that helper and never reaches the round loop at all. An accumulator placed only in the loop
captures stage 2 and the summarisation round but still drops stage-1's ~760 input tokens (~6% of the
total). Either thread stage-1's usage into the same accumulator, or explicitly document the residual
undercount — silently fixing "half missing" into "6% missing" would leave the same class of bug at a
smaller size.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: **N=10** runs of a multi-call question, reported
      `inputTokens` compared against the sum of per-round usage captured by test instrumentation (or
      against Groq's rejected-request token counts, where available)
- [ ] The round loop in `chat.ts` accumulates `inputTokens`/`outputTokens` across every round and emits
      the sum once, on the final `done`
- [ ] **N>=10** post-fix runs confirming the reported total matches the sum of per-round usage
- [ ] A single-round turn's reported usage is unchanged (regression) — accumulation must be a no-op when
      there is only one round
- [ ] Colocated unit test beside `server/routes/chat.ts` (no `chat.test.ts` exists today) asserting
      usage sums across at least two simulated rounds
- [ ] Regression gate: 16-question set passes; zero blank answers, zero raw upstream error text

### Notes

- The fix lives in `chat.ts`'s round loop, not in either provider adapter — `openai-compatible.ts` and
  `anthropic.ts` are cited above only as evidence of the correct per-call shape the accumulation reads
  from.
- `chat.ts`'s round loop is also touched by the never-route-zero-tools issue (#8864) and the `tool_choice: 'none'` issue (#8863). Land after those two land, since both change control flow in
  the same loop and this issue's diff is easier to review against a settled loop shape.
- Shares `openai-compatible.ts` with the `delta.reasoning` issue (#8862),
  the `tool_choice` issue (#8863), the temperature issue (#8865) and the prompt-caching issue (#8876) —
  per #8862's stated order, this one lands after the two urgent correctness fixes (02, 03).

### Update — 2026-08-06: two additional defects in the same accounting path, measured live

**(a) OpenAI-compatible providers that gate usage on `stream_options.include_usage` report no
usage at all.** Groq volunteers a `usage` frame unprompted; Amazon Bedrock's OpenAI-compatible
endpoint does not. Because the adapter never requested the terminal usage frame, **every** turn
on Bedrock reported `usage: null`. Verified live 2026-08-06: after adding
`stream_options: { include_usage: true }`, the same turn reports
`{"inputTokens":7122,"outputTokens":132}`. Requesting the terminal frame is in scope for this
issue — without it there is nothing to accumulate on such providers.

**(b) Usage is null whenever the final round streams no deltas.** On turns where the model made
4 tool calls and then emitted nothing (the no-text fallback fired), `usage` was `null` in 4 of 4
such records across two runs — i.e. not one call in the whole turn, including stage-1 routing,
had its usage reported. Deterministic signature: `ms_to_first_delta == latency_ms`.

Fresh evidence of the headline defect (Bedrock, gpt-oss-120b, single-turn records): a
zero-tool-call turn reports `inputTokens: 6814`; a one-tool-call turn of the same question —
which adds a complete second round re-sending the schema set plus the tool digest — reports
`7105`, only +291. A genuine cross-round sum would roughly double. A two-tool-call turn even
reported *fewer* input tokens (6716) than the zero-tool turn, which is arithmetically impossible
for a sum. The reported value behaves as the final round alone, consistent with the mechanism
described above.

Additional acceptance criteria:

- [ ] `stream_options.include_usage` (or provider equivalent) is requested on OpenAI-compatible
      streams, with a unit test asserting the request body carries it
- [ ] A turn whose final round streams no deltas still reports the accumulated usage of its
      earlier rounds (target: `usage` never null on a turn that made at least one provider call)
- [ ] Post-fix verification includes a multi-round turn where reported input tokens are
      compared against the per-round sum (must be ≈ N× a single round, not +≈300)

yenienserrano:
## Description

Closes #8874
Closes #8875

## Proposed Changes

Two related fixes to the AI Assistant's orchestration loop (`plugins/wazuh-ai-assistant`):

**#8874 — Discover handoff when data is out of reach**
- Added `suggest_discover_query`, a new non-data tool the model calls when a question touches an
  index it cannot query (guardrail-blocked index, unsupported filter, >90-day range) instead of
  guessing or answering silently.
- Added the `suggested_query` `StreamEvent` variant (`common/types.ts`) and a renderer in
  `message-bubble.tsx` that shows the model's stated reason plus an "Open in Discover" link, reusing
  the existing `buildDiscoverUrl`/`extractTimeRange` helpers.
- Field-level filters in the suggested query are only kept when `_field_caps` confirms every
  referenced field exists on the target index (`resolveSuggestedDsl`); otherwise the query is
  stripped down to index + time range only, so a hallucinated field name can never send the user to
  a silently-empty Discover search.

**#8875 — `usage` reported only the last upstream call, understating tokens ~2x**
- `chat.ts`'s round loop now accumulates `inputTokens`/`outputTokens` across every round (stage-1
  router + every stage-2/tool round) instead of forwarding only the final round's `usage`.
- OpenAI-compatible requests now set `stream_options.include_usage`, since some providers (e.g.
  Amazon Bedrock's OpenAI-compatible endpoint) only emit a terminal usage frame when explicitly
  asked, unlike others (e.g. Groq) that volunteer it.
- The terminal usage frame is now read even when it arrives after a tool-call finish event, so a
  round that ends in a tool call no longer reports `usage: null`.

### Artifacts Affected

- `plugins/wazuh-ai-assistant` — server orchestration (`server/routes/chat.ts`,
  `server/routes/chat-usage.ts`), providers (`server/providers/openai-compatible.ts`), new tool
  (`server/tools/suggest-discover-query.ts`), shared types (`common/types.ts`), and chat UI
  (`public/components/chat/message-bubble.tsx`, `chat-page.tsx`).

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `server/tools/suggest-discover-query.test.ts` — argument validation and the `_field_caps`
  safety decision behind `resolveSuggestedDsl`.
- `server/routes/chat-usage.test.ts` / `chat-stage1-usage.test.ts` — cross-round usage accumulation,
  including the stage-1 router call and a single-round no-op regression.
- `server/providers/openai-compatible.test.ts` — asserts `stream_options.include_usage` is sent, and
  that the terminal usage frame is read past a tool-call finish.
- `public/components/chat/message-bubble.test.tsx` — the `suggested_query` callout rendering.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


rodrigofez (approved):
🟢 Reviewed against #8874 and #8875.

**#8875 — met.** The fix is right, and it correctly identifies that accumulation was never the real bug: the adapter returned as soon as it saw `finish_reason: 'tool_calls'`, so every tool round (and the stage-1 router call, which always ends in a tool call) reported nothing to accumulate. Reading past that finish, requesting `stream_options.include_usage`, and threading stage-1 usage through `Stage1Result` together cover the issue. Tests cover the request body, the usage-after-tool-call frame, the single-round no-op, and the "no usage reported" case.

**#8874 — met structurally, weaker than intended in practice.** The event, the renderer, and the `_field_caps` check are all there. Two things to look at:

1. **The field check never runs for the indices this feature exists for.** `resolveSuggestedDsl` only calls `_field_caps` when the index passes `checkIndexAllowlist` — but `wazuh-threatintel-enrichments*`, `wazuh-metrics-*` and `wazuh-active-responses-*` are exactly the indices that fail it. So for all three, the query is always stripped to index + time range. That's the "weaker alternative" the issue mentioned, and it's safe, but the user gets a Discover link with no filter — asking about one IP opens the whole index for the last 24h. Worth deciding explicitly rather than by side effect.

2. **The link can vanish or point at the wrong index.** `DiscoverLink` renders nothing when no index pattern matches, and otherwise falls back to the first fuzzy title hit. The `index` value comes from the model and is only checked for being non-empty. The shipped patterns are titled `wazuh-metrics-agents*`, `wazuh-active-responses*`, `wazuh-threatintel-enrichments*`, so a model writing `wazuh-metrics-*` or `wazuh-threatintel-enrichments-a` won't match exactly. Result is either a callout with the reason and no link at all — and the link is the whole point here, unlike on a table where it's a bonus — or a link to a different index than the reason names. Resolving the index pattern before emitting the event, and skipping the callout if it can't resolve, would close this.

**No critical issues and no regressions found.** Three smaller notes:

- Not returning immediately on `finish_reason: 'tool_calls'` means the round now stays open until a usage frame, `[DONE]`, or the stream ends. There's no stream-level timeout, only client abort, so a provider that finishes and holds the socket would now block where it used to return. Unlikely, but new.
- `stream_options` now goes to every OpenAI-compatible provider. Most ignore unknown fields; a few strict gateways reject them.
- The Anthropic path is untouched, and `message_delta` carries `output_tokens` — `input_tokens` arrives on `message_start`. Worth confirming input tokens aren't still counted as zero there, otherwise the undercount this issue is about still stands for Anthropic providers.

PR #8891, opened 2026-08-06
Strip inline reasoning markup from AI Assistant answers

### LINKED ISSUE #8866: Reasoning models leak `<think>` and raw tool-call markup into the answer
## Description

Some reasoning models put their chain-of-thought **inline in `delta.content`** wrapped in `<think>`
tags, rather than on the separate `reasoning` channel. The adapter passes `content` through verbatim,
so the user's answer bubble starts with the model's private deliberation — and, worse, sometimes with
raw `<tool_call>` / `<function=…>` markup that the model emitted as text instead of as a structured
tool call.

Measured on `qwen/qwen3.6-27b`: **6 of 8 successful answers leaked this markup.**

This is the mirror image of the reasoning-channel fix (#8862): that one recovers text the adapter
was *discarding*; this one suppresses text the adapter should not be *showing*.

### Evidence

Verbatim from a captured answer (question: "How much RAM and what CPU does this host have?"):

```
<think>
The user wants to know the RAM and CPU of "this host" (implying the manager itself, agent 000, ...
Given the errors, I should try the correct field names for Wazuh states index.
Agent ID is `wazuh.agent.id`.
...
</think>

<tool_call>
<function=search_wazuh_data>
<parameter=index_pattern>
wazuh-states-*
</parameter>
```

Leak rate across the same run set:

| Model | Successful answers | Leaked markup |
|---|---|---|
| `qwen/qwen3.6-27b` | 8 | **6 (75%)** |
| `openai/gpt-oss-120b` | 1 | 0 |

`gpt-oss` does not exhibit this — it uses the dedicated `reasoning` channel (`delta.reasoning` with
`channel: "analysis"`), which the adapter now handles correctly as a fallback. Qwen 3.x inlines the
same content into `delta.content`, where nothing filters it.

Two distinct leaks, same root cause:

1. **`<think>…</think>`** — private deliberation shown as the answer. Cosmetic but severely
   unprofessional, and it exposes the model second-guessing itself in front of the analyst.
2. **`<tool_call>` / `<function=…>` / `<parameter=…>`** — the model *attempting a tool call as text*.
   This is the same underlying behaviour that community reports associate with Groq's
   "Failed to call a function" 400 (see the existing tool-call-failure issue): the model emits
   XML-ish markup instead of a structured `tool_calls` array. When Groq's validator catches it we get
   a 400; when it does not, the markup reaches the user.

### Code references

- `plugins/wazuh-ai-assistant/server/providers/openai-compatible.ts` — the `delta.content` pass-through
  (`if (delta) { sawContent = true; yield { type: 'delta', content: delta }; }`); no filtering of
  inline reasoning markup
- `plugins/wazuh-ai-assistant/server/tools/markdown-table-filter.ts` — **existing precedent**: this
  plugin already strips a class of unwanted model output (hand-built Markdown tables) from the stream,
  including the streaming/chunk-boundary handling such a filter needs

### Suggested treatment

**Recommended: strip inline reasoning markup in the adapter's delta path, reusing the streaming-filter
pattern `markdown-table-filter.ts` already establishes.** That file solves the hard part — a tag can
straddle two SSE chunks, so a naive per-chunk regex misses it; it already buffers a trailing margin
for exactly this reason.

Scope for a first pass:

- Suppress everything between `<think>` and `</think>` (inclusive), including an unclosed `<think>`
  that runs to the end of the stream.
- Suppress `<tool_call>…</tool_call>` blocks and bare `<function=…>` / `<parameter=…>` markup.
- **Do not** silently drop the whole answer: if stripping leaves nothing, the no-text fallback
  (`NO_ANSWER_MESSAGE`) should fire, so the user gets a sentence rather than an empty bubble.
- Log at debug when stripping occurs, so the rate is observable rather than invisible.

Alternative considered and rejected: filter in the UI (`message-bubble.tsx`). Rejected because the raw
markup would still be persisted into saved conversations and still count toward the answer for any
other consumer of the stream — the server is the right boundary.

## Acceptance criteria

- [ ] Reproduction recorded **before** the fix: **N=10** runs on `qwen/qwen3.6-27b` of a question that
      triggers inline reasoning, leak count stated (observed 6 of 8 on an 8-run sample)
- [ ] After the fix, no `<think>`, `<tool_call>`, `<function=` or `<parameter=` markup appears in any
      answer text
- [ ] **N>=10** post-fix runs, leak count stated — target zero
- [ ] A tag straddling two SSE chunk boundaries is still stripped (unit test with a deliberately
      split fixture)
- [ ] An unclosed `<think>` running to end-of-stream is stripped rather than shown
- [ ] A normal answer containing legitimate angle brackets (e.g. discussing `<script>` in a rule
      description, or a comparison like `size < 500`) is **not** damaged (regression)
- [ ] If stripping leaves the answer empty, the no-text fallback fires instead of a blank bubble
- [ ] `openai/gpt-oss-120b` answers are byte-identical before and after (it never emits these tags)
- [ ] Colocated unit test beside the adapter, following `markdown-table-filter.test.ts`'s style
- [ ] Regression gate: 16-question set passes; zero blank answers

### Notes

- **Found during the night-shift release gate**, after the wave-1/C/D fixes had already been
  integrated — it was invisible earlier because the earlier long-tail runs on Qwen mostly *failed*
  (tool-call rejections) before producing prose, so there were few successful answers to inspect.
  Fixing the `query_dsl` rejection is what let Qwen get far enough to expose this.
- **Reinforces the model recommendation** rather than changing it: `gpt-oss-120b` is unaffected. This
  issue is what makes Qwen unsuitable as a shipped default even though its tool selection improved
  markedly after the field-validation fix.
- Related to the tool-call-failure issue: the `<tool_call>`-as-text half of this leak is the same
  model behaviour that produces the 400 when the provider's validator catches it. Neither issue fixes
  the other; both are worth having.
- Not a blocker for a gpt-oss-only release; **is** a blocker for supporting Qwen or any other
  inline-reasoning model as a documented option.

yenienserrano:
## Description

Some reasoning models (e.g. `qwen/qwen3.6-27b`) inline their chain-of-thought directly in
`delta.content` wrapped in `<think>` tags, instead of using the separate `reasoning` channel —
and sometimes emit raw `<tool_call>`/`<function=…>`/`<parameter=…>` markup as plain text when a
structured tool call fails to form. The adapter passed `content` through verbatim, so the leaked
markup ended up in the user's answer bubble. Measured on `qwen/qwen3.6-27b`: 6 of 8 successful
answers leaked this markup.

Closes #8866

## Proposed Changes

- Added `InlineReasoningMarkupFilter` (`server/providers/inline-reasoning-markup-filter.ts`): a
  depth-tracked, streaming-safe suppressor for `<think>`, `<tool_call>`, `<function=…>`, and
  `<parameter=…>` markup. Handles a tag split across two SSE chunk boundaries, and suppresses an
  unclosed block all the way to end of stream rather than showing it.
- Wired the filter into `openai-compatible.ts`'s `chatStream`: every `content` delta is now run
  through the filter before being yielded, and the filter is drained at every stream exit path
  (tool-call, `[DONE]`, usage, loop end) — mirroring how `reasoningBuffer` is already handled.
- `sawContent` is now set from the *filtered* result, not the raw `delta` presence, so a turn whose
  entire `content` was markup and got fully stripped is correctly treated as producing no answer —
  letting the existing no-text fallback (`NO_ANSWER_MESSAGE` in `chat.ts`) fire instead of a blank
  bubble.
- A legitimate answer mentioning `<script>` or comparing `size < 500` is left untouched — the
  filter only recognizes the four known tag families, not a generic `<[^>]+>` sweep.

### Results and Evidence

- Colocated unit tests in `inline-reasoning-markup-filter.test.ts` (streaming suppressor in
  isolation) and expanded `openai-compatible.test.ts` (adapter integration), covering: a `<think>`
  block stripped before reaching the client, a tag straddling two SSE chunks, an unclosed `<think>`
  running to `[DONE]`, `<tool_call>`/`<function=>`/`<parameter=>` markup, and the `<script>`/`size <
  500` regression case (legitimate angle brackets preserved verbatim).
- Full `wazuh-ai-assistant` suite passes: 58 test suites / 764 tests.

### Artifacts Affected

- `plugins/wazuh-ai-assistant/server/providers/inline-reasoning-markup-filter.ts` (new)

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `plugins/wazuh-ai-assistant/server/providers/inline-reasoning-markup-filter.test.ts` (new):
  unit tests for the streaming suppressor in isolation.
- `plugins/wazuh-ai-assistant/server/providers/openai-compatible.test.ts`: expanded with adapter
  integration coverage for the filter (see Results and Evidence above).

## Review Checklist

- [x] Code changes reviewed
- [ ] Configuration changes documented (n/a — none)
- [ ] Developer documentation reflects the changes (n/a — none)
- [ ] ...


rodrigofez (dismissed):
🟢 Reviewed against #8866, requirement met with no regressions

**Met.** The filter is the right shape and in the right place — server-side, narrow (four known tag families, not a generic `<…>` sweep), and depth-tracked so nested `<tool_call><function=…><parameter=…>` markup stays suppressed. It's drained at all four stream exits, and `sawContent` now comes from the filtered text, so a fully-stripped answer correctly falls through to the no-text message instead of a blank bubble. Tests map onto the acceptance criteria one for one: split tag, unclosed block, `<script>`, `size < 500`, gpt-oss byte-identical, and the flush before a real tool call.

**No regressions found.** I compiled the filter and ran it standalone; clean content, `<script>`, `size < 500` and ordinary HTML-ish text all pass through byte-identical.

Three things worth a look before merge:

1. **An opening tag that never arrives isn't handled.** Many Qwen chat templates put `<think>` into the prompt itself, so the model only emits the *closing* `</think>`. I confirmed that case: input `reasoning text…</think>\n\nThe host has 8 GB` comes out completely unchanged, `didStrip` false — the whole deliberation and the stray closer both reach the user. Same for any unmatched `</think>` mid-answer. Since this is exactly the deployment shape the issue is about, it's worth deciding now whether a closer with no opener should drop everything before it, or at minimum strip the tag itself. Your captured evidence does show the opener present, so this is about other setups rather than the case you measured.

2. **Matching is exact-case and exact-spacing.** `<THINK>` and `<think >` both pass through untouched. Models are consistent enough that this probably never fires, but it's a cheap thing to harden.

3. **Quoted XML loses its contents.** If an answer quotes a config or code block containing `<parameter>` or `<function>`, everything between the tags is eaten — I saw `<parameter>-v</parameter>` disappear from an otherwise intact snippet. Acceptable trade for now, but worth knowing it's there.

**Also:** the debug-log-on-strip criterion isn't implemented — the PR explains why (adapters aren't given a logger) and exposes `didStrip` for later, which is a fair call, but it means the leak rate stays unobservable in production. The CHANGELOG entry linking #8866 is missing. And the issue asks for N≥10 post-fix runs with a leak count; the PR gives unit-test evidence and a full-suite pass, but not that number.

**One coordination note:** #8888 rewrites the same four exit paths in `openai-compatible.ts` — it stops returning at `finish_reason: 'tool_calls'` so it can read the trailing usage frame. Whichever lands second needs a deliberate check that the flush still happens exactly where it should. I traced the combination and it looks benign (`flush()` is safe to call twice), but it shouldn't be resolved as a mechanical conflict.

rodrigofez (approved):

gonzaarancibia (approved):

PR #8902, opened 2026-08-07
Harden privacy-mode pseudonymization robustness

### LINKED ISSUE #8889: Harden privacy-mode pseudonymization robustness
## Description

Privacy mode pseudonymizes sensitive values before tool-result content reaches the model provider. A review of the pseudonymization path surfaced several robustness gaps worth tightening:

1. Field-kind inference uses substring matching, which can misclassify some field names (e.g. a field name that merely *contains* a shorter token gets the wrong pseudonym kind), leading to confusing labels in the model's context.
2. The dotted-token scanner that catches hostnames/FQDNs in free text can also match non-hostname tokens — structured version strings and field-path-like identifiers — replacing them unnecessarily and, in the field-path case, breaking a query when a user names a dotted field.
3. Values in fields that are allow-listed for readability are passed through without a secondary value-shape scan, so an identifier embedded inside an otherwise-allowed field is not caught.
4. At least one commonly-surfaced field lacks an explicit policy entry and is handled by allow-by-omission rather than an intentional decision.

### Code references

- `plugins/wazuh-ai-assistant/server/tools/privacy.ts` — `inferPseudonymKind` (substring matching), the FQDN/dotted-token scan in `prescanAndMint`, `FIELD_POLICY_DEFAULTS`
- `plugins/wazuh-ai-assistant/server/tools/catalog/common.ts` — the finding sample-column set

### Suggested treatment

- Replace substring-based field-kind inference with segment/word matching.
- Narrow the dotted-token scanner to exclude known structured identifiers (version strings) and field-path-like tokens, so it does not rewrite them.
- Run the value-shape scan over allow-listed field values too, so an embedded identifier is still caught while the field stays readable.
- Add explicit policy entries for commonly-surfaced fields instead of relying on allow-by-omission.

## Acceptance criteria

- [ ] Colocated unit tests: field-kind inference segments correctly; scanner excludes version strings and field-path tokens; allow-listed field values still get an embedded-identifier scan; explicit policy entries present.
- [ ] Regression: with privacy mode on, a user referencing a dotted field name does not break the query.

rodrigofez:
## Description

Closes #8889

Privacy mode pseudonymizes sensitive values in tool results before they are sent to the model provider. A review of that path (`plugins/wazuh-ai-assistant/server/tools/privacy.ts`) surfaced four robustness gaps where the scrubbing was either too loose (an identifier reached the provider) or too aggressive (a value the user needed was rewritten). None of them is a new regression — they are pre-existing sharp edges in the pseudonymization heuristics. Two of them are user-visible: the over-aggressive dotted-token scanner silently breaks a query when a user names a dotted field (e.g. asks about `wazuh.agent.name`) or asks about a package version, and the substring-based field-kind inference makes the model narrate an unrelated field's opaque token as if it were a literal IP address.

The changes are confined to the pseudonymization layer; no tool, route, or provider contract changes.

## Proposed Changes

**1. Field-kind inference now matches whole tokens, not raw substrings** (`inferPseudonymKind`)

The old check was `field.toLowerCase().includes('ip')`, which is true for `wazuh.rule.description` — "descr-**ip**-tion" — so description-style fields were classified as IP fields and the model was told their opaque pseudonyms were IP addresses. Field names are now split into word tokens on `.`, `_`, `-`, and camelCase boundaries (`fieldNameTokens`), and a keyword must occupy a whole token or be glued to it as a prefix/suffix (`tokenMatchesKeyword`). Compound legacy Wazuh alert fields (`data.srcip`, `data.srcuser`) still classify correctly because the keyword only has to be a token suffix.

The `.name` → `HOST` alias rule was also tightened to an exact match on the final **path** segment, so `country_name` (a place, not a host) is no longer swept in while `wazuh.agent.name` still is.

**2. The FQDN/dotted-token scanner no longer rewrites version strings or field paths** (`prescanAndMint`)

Two new exclusions alongside the pre-existing all-numeric/timestamp one:

- `VERSION_LIKE_TOKEN_RE` — a semver/Debian-revision shape (`5.2.5-2ubuntu1`, `v1.2.3`). Minting a `HOST_n` here undermined the `package.version` query the user was asking about. It requires the token to start with an optional `v` plus a digit, so a real hostname (whose first label is essentially never digits-only) can't be excluded by it.
- `isFieldPathToken` — every `.`-separated segment being a known field-path word means the token reads as the user *naming a field*, not a hostname. The word set (`FIELD_PATH_WORDS`) is derived from `WAZUH_FIELD`'s values and `FIELD_POLICY_DEFAULTS`' entries, so it self-updates when a field is added — there is no parallel list to hand-maintain.

Real hostnames like `backup-vault.internal.corp` still mint as before.

**3. Allow-by-omission values now get a value-shape scan** (`applyFieldPolicy`)

A field with no `FIELD_POLICY_DEFAULTS` entry on a typed tool resolves to "allow" by omission and previously skipped any secondary scan, so an IP or FQDN embedded in otherwise free text passed through. Both the samples loop and the aggregation-breakdown loop now run `prescanAndMint` over that case. The field itself stays readable; only an embedded identifier is replaced.

> **Deviation from the issue's suggested treatment, deliberate.** The issue suggested scanning *all* allow-listed values. This is scoped to allow-**by-omission** only, and explicit `action: 'allow'` entries are left exactly as before. An explicit entry is a reviewed decision, and several of those curated values are legitimately FQDN-token-shaped without being hostnames — MITRE sub-technique IDs (`T1059.001`), compliance citations, CIS benchmark text — so scanning them would misfire and corrupt them. Those values are not left unscanned: `chat.ts`'s `scrubMessagesForProvider` runs `prescanAndMintToolContent` over every tool-result string value regardless of field policy, immediately before the request leaves the server. A test pins this scoping so it isn't "fixed" later by mistake.

**4. `wazuh.rule.title` has an explicit policy entry**

It appears in every finding-hits tool's sample columns (`catalog/common.ts`'s `STANDARD_FINDING_SAMPLE_COLUMNS`) but had no entry at all, so it reached the provider by omission rather than by decision. Reviewed and set to `'allow'`, with the reasoning and the accepted residual risk recorded inline: titles are overwhelmingly fixed strings from Wazuh's curated ruleset and the model needs the real text to name a finding, but a custom rule's `<description>` can interpolate a decoder capture group, so a title *can* echo attacker-influenced content. That case is covered by the `chat.ts` pass described above.

`catalog/common.ts` itself is unchanged — the issue referenced it only to identify which field was missing a policy.

### Results and Evidence

Unit tests (in the `osd-dev` container), full plugin suite:

```
Test Suites: 57 passed, 57 total
Tests:       755 passed, 755 total
Time:        34.863 s
```

`server/tools/privacy.test.ts` alone: **48 passed**, including all 10 new cases and the `FIELD_POLICY_DEFAULTS: wazuh.rule.title has an explicit entry, not allow-by-omission` guard.

Prettier (2.8.8, the version the plugins pin and CI installs) and ESLint are clean on all changed files.

#### How to Test

With privacy mode enabled in the AI Assistant settings:

1. Ask a question that names a dotted field, e.g. *"how many findings per `wazuh.agent.name`?"* → the field name must appear verbatim in the request and the query must succeed. Before this change the mention was replaced with a `HOST_n` token and the model replied that the field did not exist.
2. Ask about a package version, e.g. *"which agents have openssl 3.0.2 installed?"* → the version string must survive intact.
3. Ask something that surfaces a description-style field → its pseudonym must be a `VAL_n`, not an `IP_n`, and the model must not describe it as an IP address.

### Artifacts Affected

- `plugins/wazuh-ai-assistant` — server-side only (`server/tools/privacy.ts`). No packaging, executable, or default-configuration change.

### Configuration Changes

None. No new settings and no changed defaults. `FIELD_POLICY_DEFAULTS` gains one explicit `wazuh.rule.title` entry, which records the behaviour that field already had (allowed) rather than changing it.

### Documentation Updates

None required. The reasoning for each decision — including the accepted residual risk on `wazuh.rule.title` and the deliberate scoping in item 3 — is documented inline at the code it governs.

### Tests Introduced

10 new colocated cases in `plugins/wazuh-ai-assistant/server/tools/privacy.test.ts`, one per acceptance criterion:

Scanner exclusions (`prescanAndMint`):
- leaves a field-path mention (`wazuh.agent.name`) untouched
- leaves another field-path mention (`wazuh.rule.id`) untouched
- leaves a Debian/Ubuntu-style package version string untouched
- still mints a real hostname that is not field-path or version shaped *(guards against over-correcting)*

Field-kind inference:
- no longer misclassifies `description` as an IP field
- still infers IP for an actual `...ip`-shaped field *(guards against over-correcting)*

Allow-by-omission scan (`applyFieldPolicy`):
- allow-by-omission value with an embedded IP still gets scanned
- allow-by-omission value with an embedded FQDN still gets scanned
- an EXPLICIT `allow` entry is **not** scanned for embedded IPs/FQDNs *(pins the deliberate scoping)*

Policy coverage:
- `FIELD_POLICY_DEFAULTS`: `wazuh.rule.title` has an explicit entry, not allow-by-omission

One pre-existing test's fixture was adjusted: it used `web-01.corp` for a field resolved by allow-by-omission, which the new scan now pseudonymizes. Changed to a non-dotted value so the test keeps asserting what it was written for (tool-scoping isolation) instead of accidentally asserting the new behaviour.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,7 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
+| [#8889](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8889) | Fixed privacy-mode pseudonymization misclassifying some field names, over-scrubbing dotted field-path mentions and version strings in free text, missing identifiers embedded in allow-listed field values, and `wazuh.rule.title` having no explicit field policy                          |
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,7 +133,7 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
-| [#8889](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8889) | Fixed privacy-mode pseudonymization misclassifying some field names, over-scrubbing dotted field-path mentions and version strings in free text, missing identifiers embedded in allow-listed field values, and `wazuh.rule.title` having no explicit field policy                          |
+| [#8889](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8889) | Fixed privacy mode misclassifying some field names, over-scrubbing dotted field paths and version strings in free text, and missing identifiers embedded in allow-listed field values         |
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,7 +133,6 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
-| [#8889](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8889) | Fixed privacy mode misclassifying some field names, over-scrubbing dotted field paths and version strings in free text, and missing identifiers embedded in allow-listed field values         |
 
 ### Removed
 
--- a/plugins/wazuh-ai-assistant/server/tools/privacy.test.ts
+++ b/plugins/wazuh-ai-assistant/server/tools/privacy.test.ts
@@ -8,8 +8,10 @@ import {
   prescanAndMint,
   prescanAndMintToolContent,
   FieldPolicyEntry,
+  FIELD_POLICY_DEFAULTS,
 } from './privacy';
 import { Digest } from './digest';
+import { WAZUH_FIELD } from '../../common/wazuh-fields';
 
 // --- prescanAndMint (first-mention inbound pre-scan) ------------------------------------------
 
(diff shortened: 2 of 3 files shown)


yenienserrano (approved):
LGTM :green_circle:

PR #8904, opened 2026-08-07
Ask for the answer on the final tool round

### LINKED ISSUE #8893: A turn that exhausts the tool-round budget produces no model answer
| 5.0.0 | 5.0.0-00 | Chrome |

**Description**

When the AI Assistant uses up its tool-round budget, the turn ends with no
model-written answer at all — the user is left with a populated results table and
the fallback copy `No additional analysis — see the results above.`

The final round is meant to close the turn out in prose. It drops `tools` as a
structural terminator (deliberately, so no model can call a tool it was told not
to), but it adds **no instruction asking the model to answer from the results it
already gathered**. The model receives a transcript full of tool results with
nothing requesting a conclusion, so producing no text is a reasonable thing for it
to do rather than a malfunction.

The signature is deterministic: exactly 4 tool calls in the turn — the 3
tool-bearing rounds the budget allows, plus the stage-1 router's own forced
`route_question` call.

This was first observed while validating the reasoning-channel blank-answer fix,
and a comment on #8862 records that measurement. It is a **different mechanism**
from that one: there, the model did produce an answer and it arrived on a channel
nothing read; here, the model genuinely produces nothing, and the existing
`!sawAnyDelta` fallback is what puts text on screen.

**Preconditions**

1. An AI Assistant provider configured and set as default (measured on an
   OpenAI-compatible endpoint serving `openai/gpt-oss-120b`).
2. An environment with enough data that a question needs several tool calls to
   answer.

**Steps to reproduce**

1. Open the AI Assistant chat.
2. Ask a broad, multi-part analyst question — one that cannot be served by a
   single tool call. Questions phrased like a shift handover ("what should I
   focus on across the estate today?") reliably reach the budget.
3. Let the turn run to completion and watch the tool-call chips.
4. Observe a turn that made 4 tool calls and rendered at least one non-empty
   results table.

**Expected Result**

1. The assistant writes a short answer to the question, grounded in the results
   it already retrieved.

**Actual Result**

1. The turn ends with no model-written text. The user sees the results table(s)
   plus the fallback line `No additional analysis — see the results above.`,
   which is an unhelpful non-answer sitting directly above real data the user
   asked to have interpreted.

**Additional context**

Measured over a 40-question analyst-persona bank (5 personas, single run each) on
`openai/gpt-oss-120b`:

- **6 of 40** turns (15%) ended on fallback copy instead of a model answer.
- **All 6** had the same 4-tool-call signature.
- **4 of the 6** had rendered a **non-empty** table and showed
  `No additional analysis — see the results above.` — these are the ones where a
  real answer was both possible and expected.
- The remaining 2 had genuinely empty results and correctly showed
  `No matching results were found for that query.`

So roughly **1 in 10 analyst questions** currently produces a non-answer over data
the assistant successfully retrieved.

**Fix to evaluate:** on the final round, send a short explicit instruction to
answer the user's question from the results already gathered. That keeps the
structural terminator (still no `tools` offered) while giving the model the one
thing it is currently missing — a request for a conclusion — so the *model*
produces the analysis.

**Alternative already considered and rejected:** widening the tool-round budget.
It does not address the cause and simply relocates the same failure to 5 tool
calls.

**Note on the fallback copy itself:** improving the fallback wording (for example
by having the server restate the digest counts) would only make the non-answer
read better. It is worth considering separately, but it is not a fix for this
issue — the goal here is for the model to answer.

rodrigofez:
## Description

Dropping `tools` on the final round terminates the tool loop, but it only stops the model from calling anything — it never asks for a conclusion. The model receives a transcript of tool results with no outstanding request, so producing no text is a rational completion, and the turn ends with the `!sawAnyDelta` fallback copy sitting above a populated results table.

Measured on `openai/gpt-oss-120b` over a 40-question analyst-persona bank: 6/40 turns ended on fallback copy, every one with the same 4-tool-call signature (the 3 tool-bearing rounds the budget allows plus stage 1's own forced `route_question`), and 4 of those 6 had already rendered a non-empty table.

This PR appends a short instruction to the final round's outbound messages asking the model to answer from the results already gathered. It's constrained so it buys analysis without buying invention: it scopes the answer to the gathered results, forbids stating what the results do not show, and keeps an honest "this does not answer the question" reachable.

Applied to the outbound copy only, so it never enters conversation history, and gated on the turn having actually run a tool — a general-routed turn has no results to reason over. Extracted as a pure `withFinalRoundAnswerInstruction` so the decision is unit-testable without a fake orchestrate run.

Refs https://github.com/wazuh/wazuh-dashboard-plugins/issues/8893

## How to Test

- Run the AI Assistant chat and exhaust the tool-round budget on a question that requires multiple tool calls.
- Confirm the final turn produces a written answer grounded in the gathered results instead of falling back to placeholder copy.
- `yarn test:jest --testPathPattern=chat-final-round-instruction` in `plugins/wazuh-ai-assistant` (inside the OSD dev container).

## Verification

- Prettier (`--check`) on changed files: **PASS**
- ESLint on changed files: **PASS**
- `tsc -p tsconfig.typecheck.json`: fails, but identically on unrelated sibling files (`chat-stage1-usage.test.ts`, `chat-stream-limiter.test.ts`, `chat-usage.test.ts`, `chat.ts` itself) because this plugin repo standalone can't resolve `../../../../src/core/server` / `@osd/config-schema` / test globals outside the full `wazuh-dashboard` checkout it's built against — a pre-existing environment gap, not introduced by this change.
- `yarn test:jest --testPathPattern=chat-final-round-instruction` not run in this environment (needs the OSD-mounted dev container); please run it as part of review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,7 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
+| [#8893](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8893) | Fixed the AI Assistant ending a turn with no written answer after using up its tool rounds, leaving fallback text above a populated results table                                              |
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,7 +133,7 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
-| [#8893](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8893) | Fixed the AI Assistant ending a turn with no written answer after using up its tool rounds, leaving fallback text above a populated results table                                              |
+| [#8893](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8893) | Fixed the AI Assistant ending a turn with no written answer after using up its tool rounds, leaving fallback text above a populated results table                                             |
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


yenienserrano (approved):
LGTM :green_circle:

PR #8905, opened 2026-08-07
Cap aggregation limits from one source of truth

### LINKED ISSUE #8894: `get_sca_results` fails for any limit between 101 and 500
| 5.0.0 | 5.0.0-00 | n/a (server-side) |

**Description**

`get_sca_results` advertises `limit` up to 500, but any value between 101 and 500
makes the tool fail instead of returning results.

The tool clamps `limit` to a maximum of **500** and then passes that value straight
into a `terms` aggregation `size`. Every indexer request built by a tool is linted
before it runs, and that lint rejects an aggregation `size` above **100**. So the
request is refused and the assistant gets an error string back rather than SCA
policy results.

The tool's own parameter description tells the model the maximum is 500, so a model
that follows the description into the 101-500 range is guaranteed to fail.

This is an isolated oversight rather than a pattern. Four sibling tools feed their
`limit` into a `terms` aggregation in exactly the same way and all four correctly
clamp to 100, matching the lint cap:

- `get_compliance_summary` — clamps to 100 ✅
- `get_mitre_summary` — clamps to 100 ✅
- `get_security_summary` — clamps to 100 ✅
- `get_top_rules` — clamps to 100 ✅
- `get_sca_results` — clamps to **500** ❌

**Preconditions**

1. An AI Assistant provider configured and set as default.
2. At least one agent with SCA results available.

**Steps to reproduce**

1. Ask the assistant an SCA question phrased so it requests a large number of
   policies for one agent — e.g. asking for "all" or "up to 200" SCA policies for a
   specific agent — so the model emits `get_sca_results` with a `limit` above 100.
2. Observe the tool-call result.

**Expected Result**

1. Either SCA policy results are returned, or the parameter is clamped to a value
   the query layer accepts so the call still succeeds.

**Actual Result**

1. The tool call fails with an aggregation-size violation
   (`Aggregation size (N) exceeds the maximum of 100.`), and no SCA results reach
   the model. The assistant then has to answer without the data it asked for.

**Additional context**

Mechanism, for whoever picks this up:

- `get_sca_results` clamps `limit` with a maximum of 500, then uses that value as
  the `size` of a `terms` aggregation over `policy.id`.
- The guardrail layer defines a maximum aggregation size of 100 and rejects
  anything larger.
- That lint runs on **every** indexer request in the tool executor, with no
  per-tool exemption, so a typed tool is not exempt from it.

Suggested fix, matching what the four sibling tools already do: clamp `limit` to
100 and correct the parameter description so it no longer advertises 500. If SCA
genuinely needs more than 100 policies in one call, the alternative is to page the
aggregation, but note that no other aggregation tool does this today — aligning
with the existing 100 pattern is the smaller and more consistent change.

Found while building a tool-use benchmark for the assistant; no runtime data was
needed to confirm it, the mismatch is visible in the two constants.

rodrigofez:
## Description

`get_sca_results` clamped its `limit` parameter to a maximum of 500 and wrote
that value into a `terms` aggregation size, but `guardrails.ts` rejects any
aggregation size above `MAX_AGG_SIZE` (100), and `executor.ts` runs `lintDsl`
on every indexer request with no per-tool exemption. Every call with a limit
in the 101-500 range therefore hard-failed, and the tool's advertised
parameter description told a compliant model it could go up to 500, steering
it straight into the broken range.

Closes #8894

## Proposed Changes

Rather than just correcting the one hard-coded number, this fixes the class
of bug:

- Export `MAX_AGG_SIZE` so catalog code reaches the single enforced cap
  instead of restating it as a separate literal.
- Add `clampAggLimit` and `aggLimitProperty` to `catalog/common.ts`. The clamp
  takes its ceiling from `MAX_AGG_SIZE`, and the schema description is
  generated from that same constant, so the advertised maximum can no longer
  drift from the enforced one.
- Migrate all five aggregation-backed tools (`get_sca_results`,
  `get_compliance_summary`, `get_mitre_summary`, `get_security_summary`,
  `get_top_rules`) onto the shared helper. Four were already correct only by
  virtue of a hard-coded `100`; routing them through the shared helper makes
  them structurally correct instead of coincidentally correct.
- Hits-paging limits keep the existing `clampLimit`, since `applySafetyValves`
  clamps a large hits size rather than rejecting it outright — that asymmetry
  is why aggregation limits and hits limits need separate helpers.

### Tests Introduced

- Added `catalog/agg-size-coverage.test.ts`, driven from
  `listToolDefinitions()` so every current and future aggregation-backed tool
  is covered with nothing exempt by default. For each tool it builds a
  request with an absurd limit and runs it through the same two guardrail
  stages `executor.ts` applies, in the same order, so the invariant tracks the
  guardrails rather than duplicating their arithmetic. A second case detects
  behaviorally whether a limit actually reaches an aggregation (building twice
  and comparing sizes), so tools whose breakdown aggregations are sized
  independently of `limit` aren't falsely flagged.
- Updated `get-sca-results.test.ts` for the newly enforced cap.

## Review Checklist

- [x] Code changes reviewed


yenienserrano (approved):
LGTM :green_circle:

PR #8924, opened 2026-08-10
Add filter param to get_agent_inventory for presence checks

### LINKED ISSUE #8910: AI Assistant inventory presence questions unanswerable: get_agent_inventory lacks a filter parameter
## Description
`get_agent_inventory` accepts only `agent_id`, `agent_name`, `kind`, and `limit` — there is no way to filter the returned rows. Presence questions ("is openssl installed on host X?", "which process is listening on port 9200?") are therefore unanswerable by construction: the tool returns the first `limit` rows (clamped at 500, `sort: ['_doc']`) of an inventory that can be larger (691 packages on the benchmark host), and the model must scan a truncated, arbitrarily ordered sample for one specific entry.

Measured live (openai.gpt-oss-120b, N=5 per question): "Is openssl installed on <agent>?" was answered **wrong 5/5** — openssl was installed (4 matching documents in `wazuh-states-inventory-packages*`), but it was not in the 500-row sample, so runs answered "its presence cannot be confirmed" or produced a self-contradictory sentence. "What process is on port 9200?" failed the same way 3/5.

There is no workaround path left, by design: follow-up `search_wazuh_data` attempts are correctly rejected by field validation (the model invents field names, and the states indices carry no `@timestamp`), and `suggest_discover_query` was never invoked (0 of 65 benchmark runs) despite being offered on every tool round.

## Tasks
- [ ] Add an optional `filter` parameter to `get_agent_inventory` that narrows results on the kind's primary name field (package name, port number, process name, hotfix id) — e.g. a case-insensitive match/wildcard applied server-side before the limit.
- [ ] Advertise the parameter in the tool description so the model routes presence questions through it instead of scanning samples ("to check whether one specific package/port/process is present, pass `filter`").
- [ ] When results are truncated (`truncated: true`), have the digest note suggest using `filter` for targeted lookups.
- [ ] Colocated tests: schema accepts/omits `filter`; the outbound Indexer query carries the filter clause per kind; description text pins the routing guidance.

yenienserrano:
## Description

Closes #8910

`get_agent_inventory` accepted only `agent_id`, `agent_name`, `kind`, and `limit` — there was no way to filter the returned rows. Presence questions ("is openssl installed on host X?", "which process is listening on port 9200?") were unanswerable by construction: the tool returns the first `limit` rows (clamped at 500) of an inventory that can be larger, and the model had to scan a truncated, arbitrarily ordered sample for one specific entry. Measured live, this caused wrong answers in the majority of presence-question benchmark runs.

## Proposed Changes

- Added an optional `filter` parameter to `get_agent_inventory` that narrows results on the kind's primary field (package name, port number, process name, hotfix id) — a case-insensitive match applied server-side before the limit.
- Sanitized the filter value so a caller-supplied `*`/`?` cannot produce a leading wildcard.
- Advertised the parameter in the tool description so the model routes presence questions through it instead of scanning samples.

### Results and Evidence

Colocated unit tests cover the schema, per-kind filter behavior, and sanitization. No user-facing UI changes.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

`plugins/wazuh-ai-assistant/server/tools/catalog/get-agent-inventory.test.ts`: schema accepts an optional `filter`; per-kind filter narrowing (packages/hotfixes/processes/os/ports); numeric-equality fallback for ports; sanitization against wildcard injection; blank/whitespace filter treated as omitted; description text pins routing guidance.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,7 +133,6 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
-| [#8910](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8910) | Fixed AI Assistant `get_agent_inventory` tool being unable to answer presence questions (e.g. "is a package/port/process present") by adding an optional `filter` parameter                  |
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


jcorredor-spec (approved):
LGTM

01-real-llm-openssl-presence-check

felipegonzalezmv (approved):

PR #8927, opened 2026-08-10
Fix AI Assistant privacy-mode over-redaction of allowlisted inventory fields

### LINKED ISSUE #8917: Privacy mode over-redacts get_agent_inventory package names and versions, making them unusable
## Description
With privacy mode on, `get_agent_inventory`'s `packages` results are over-redacted to the point of being unusable: both the package name and the package version are replaced with fresh pseudonyms for the whole value, even though both fields carry explicit policy entries (`package.name` is `allow-scan`, `package.version` is `allow`).

Observed on a deployed build, captured from the actual outbound provider payload:

```
{"package.name":"HOST_3","package.version":"VAL_6","package.vendor":"VAL_7"}
```

where the real values were an ordinary package name and version (`adduser`, `3.118ubuntu5`). Reproduced on a first turn of a fresh conversation with an empty pseudonym dictionary, so it does not depend on accumulated state. The model therefore reasons over placeholder names and versions and cannot answer package or version questions at all while privacy mode is on. Direction of the fault is over-redaction, not leakage — no real identifier escapes.

`get_vulnerabilities_by_agent` does not behave this way: its `package.version` values pass through intact.

Likely mechanism, from reading the code: `executor.ts` passes each tool's `deriveColumns` flag as the "is escape hatch" argument to the digest's field-policy pass. `get_agent_inventory` sets `deriveColumns: true` (it must, since one tool serves five different kinds with different columns), so its digest is treated like the untyped escape hatch, where any field lacking a policy entry is fail-closed pseudonymized. What still needs confirming is why the explicit `package.name` / `package.version` entries do not take effect on this path — the fail-closed branch should only apply to fields with no entry at all.

## Tasks
- [ ] Determine why explicit field-policy entries are not honored for a `deriveColumns: true` tool's derived columns, and fix it so an explicit entry wins over the fail-closed default.
- [ ] Reconsider conflating `deriveColumns` with "is escape hatch": a typed tool that derives its columns per kind is not the same risk as the untyped search tool. If the two need different treatment, give them separate flags.
- [ ] Colocated tests: with privacy on, a `deriveColumns: true` tool's rows keep values whose fields have an explicit `allow` entry, still scan `allow-scan` fields, and still fail closed for fields with no entry.
- [ ] Verify live: a packages question under privacy mode returns readable names and versions while agent/host identifiers remain pseudonymized.

yenienserrano:
## Description

Closes #8917

With privacy mode on, `get_agent_inventory`'s `packages` results were over-redacted to the point of being unusable: both the package name and version were replaced with fresh pseudonyms, even though both fields carry explicit policy entries (`package.name` is `allow-scan`, `package.version` is `allow`). Root cause: `executor.ts` passed each tool's `deriveColumns` flag as the "is escape hatch" argument to the digest's field-policy pass, so `get_agent_inventory` (which must set `deriveColumns: true` since one tool serves five kinds with different columns) was treated like the untyped escape hatch, where any field lacking a policy entry is fail-closed — and the explicit entries weren't winning over that fail-closed default. Separately, a persisted privacy settings object never picked up field policy entries added by a later release, which was the root cause of the reported over-redaction on deployed builds.

## Proposed Changes

- Decoupled the AI Assistant privacy fail-closed default from `deriveColumns`: an explicit field-policy entry (`allow`/`allow-scan`) now always wins, regardless of whether the tool derives its columns.
- Reconciled the persisted AI Assistant field policy with the shipped defaults on read, so new policy entries from a later release are picked up while preserving every admin customization, including deliberate field removals.

### Results and Evidence

Colocated unit tests cover both the field-policy precedence fix and the persisted-settings reconciliation, including a regression test for the empty-pseudonym-dictionary first-turn case from the report. No user-facing UI changes.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `plugins/wazuh-ai-assistant/server/tools/catalog/get-agent-inventory.test.ts`, `find-document-by-field.test.ts`, `search-wazuh-data.test.ts`, `executor.test.ts`, `privacy.test.ts`: an explicit `allow`/`allow-scan` policy entry wins over the fail-closed default for `deriveColumns: true` tools; fields with no entry still fail closed.
- `plugins/wazuh-ai-assistant/server/routes/field-policy-reconciliation.test.ts`: persisted field policy is reconciled with shipped defaults on read while preserving admin customizations and deliberate removals.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


jcorredor-spec (dismissed):
Privacy toggled On, asked "What packages and their vendor are installed on agent 002? Show the first 5.":
the answer and table show real names/versions/architectures matching the raw indexer
data exactly (acl 2.2.53-1.el8, etc.) , confirms the allowlisted fields reach the user unredacted end-to-end. 

Image

jcorredor-spec (approved):

Issue #8912, opened 2026-08-08
Harden privacy mode for free-text allow-listed fields (allow-scan policy action)


MiguelCasaresRobles:
## Description
Privacy mode's field policy supports three actions — `allow`, `anonymize`, `never`. An `allow` field's value is forwarded verbatim. The message-boundary scan catches identifier *shapes* (IPs, dotted FQDNs) embedded in text, but some allow-listed fields can carry free text, and free text can contain identifiers that have no matchable shape (e.g. a bare, dotless machine name). Internal review confirmed such a value in an allow-listed field reaches the provider unmodified while privacy mode is on; the hardening merged in #8902 explicitly scopes its scan to allow-by-omission fields and does not cover this case.

Shape-based scanning cannot close this by construction — a bare identifier looks like any other word. What the pseudonymizer does have is knowledge: identifiers it has already minted pseudonyms for elsewhere in the conversation (agent names, hostnames, IPs). A dictionary lookup catches what shape-matching cannot, with near-zero false positives.

## Tasks
- [ ] Add a fourth field-policy action `allow-scan`: the value passes, but first runs (a) the existing shape scan and (b) an exact word-boundary match against the pseudonymizer's known-entity dictionary for the conversation; matches are replaced with their existing pseudonyms.
- [ ] Audit `FIELD_POLICY_DEFAULTS` and reclassify `allow` fields that can carry free or third-party-influenced text to `allow-scan` (starting with `package.name`), keeping the per-field rationale comments. Curated-vocabulary fields (MITRE names, compliance ids, `check.id`) stay `allow`.
- [ ] Accept the new action in the assistant-settings saved object's `fieldPolicy` validation so per-field overrides keep working.
- [ ] Colocated tests: dictionary hit in an `allow-scan` field is pseudonymized; the same value in a curated `allow` field passes; unknown identifiers (never seen by the pseudonymizer) pass — documenting the residual limitation honestly.

MiguelCasaresRobles:
## Cross-reference: #8920 adds a `package.vendor` entry that should become `allow-scan` when this lands

Heads-up for whoever merges this issue's branch (`fix/8912-privacy-allow-scan`), since the two touch
the same `FIELD_POLICY_DEFAULTS` table.

Live testing on a build carrying all eleven open AI Assistant branches found that
`get_agent_inventory` sets `deriveColumns: true`, which makes `applyFieldPolicy` **fail-closed** for
[...]

PR #8930, opened 2026-08-10
Fix privacy mode substring corruption via word boundaries

### LINKED ISSUE #8916: Privacy mode corrupts values by substring pseudonym replacement (e.g. package versions)
## Description
In privacy mode, pseudonym substitution is applied as a plain substring replacement over field values, so a pseudonymized token that happens to occur inside an unrelated string corrupts that string. Observed on a live deployment: a package version `7.81.0-1ubuntu1.14` was sent to the provider as `7.81.0-1VAL_21.14`, because the word `ubuntu` had been pseudonymized to `VAL_2` from `host.os.platform` earlier in the same conversation.

This is not a data leak — the direction is over-replacement, not under-replacement — but it corrupts values the model then reasons over. Version strings are the clearest case: a mangled version can make the assistant state the wrong package version, or fail to match a version against a vulnerability range, and the user cannot tell the value was altered.

Related, already documented: the known-entity dictionary scan added for allow-scanned fields (#8912) deliberately uses word-boundary matching to avoid exactly this class of damage. The generic text substitution path does not.

## Tasks
- [ ] Apply the same word-boundary discipline used by the known-entity scan to the generic pseudonym substitution over field values, so a pseudonym only replaces a whole token rather than any substring.
- [ ] Leave genuinely embedded identifiers (IPs, FQDNs) still covered — they are matched by shape, not by dictionary substring, and must keep working.
- [ ] Consider excluding version-shaped values from generic substitution entirely (a version-like token already has precedent in this code).
- [ ] Colocated tests: a pseudonymized word inside a version string leaves the version intact; a standalone occurrence of the same word is still replaced; an embedded IP/FQDN is still pseudonymized.

yenienserrano:
## Description

Closes #8916

In privacy mode, pseudonym substitution was applied as a plain substring replacement over field values, so a pseudonymized token that happened to occur inside an unrelated string corrupted that string. Observed on a live deployment: a package version `7.81.0-1ubuntu1.14` was sent to the provider as `7.81.0-1VAL_21.14`, because the word `ubuntu` had been pseudonymized to `VAL_2` from `host.os.platform` earlier in the same conversation. This is over-replacement, not a data leak, but it corrupts values the model then reasons over — a mangled version can make the assistant state the wrong package version or fail to match it against a vulnerability range.

## Proposed Changes

- Applied word-boundary discipline to the generic pseudonym substitution over field values, so a pseudonym only replaces a whole token rather than any substring — matching the discipline the known-entity dictionary scan (#8912) already uses.
- Left genuinely embedded identifiers (IPs, FQDNs) still covered, since those are matched by shape, not by dictionary substring.

### Results and Evidence

Colocated unit tests cover the word-boundary fix. No user-facing UI changes.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

`plugins/wazuh-ai-assistant/server/tools/privacy.test.ts`: a pseudonymized word inside a version string leaves the version intact; a standalone occurrence of the same word is still replaced; an embedded IP/FQDN is still pseudonymized.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,7 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
+| [#8916](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8916) | Fixed the AI Assistant's privacy mode corrupting values (e.g. package versions) by substring-replacing a pseudonymized word wherever it appeared, instead of only at whole-token boundaries   |
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


jcorredor-spec (approved):
LGTM

jcorredor-spec (approved):
LGTM

```
(node:1343) Warning: Setting the NODE_TLS_REJECT_UNAUTHORIZED environment variable to '0' makes TLS connections and HTTPS requests insecure by disabling certificate verification.
(Use `node --trace-warnings ...` to show where the warning was created)
CHAT RESPONSE STATUS 200
MOCK RECEIVED 2 requests
ASSISTANT MSG SEEN BY PROVIDER: "openssh version is 7.81.0-1ubuntu1.14 on this host."
ASSISTANT MSG SEEN BY PROVIDER: "openssh version is 7.81.0-1ubuntu1.14 on this host."
```

PR #8932, opened 2026-08-10
Narrow get_agent_inventory ports filter to listening-state sockets

### LINKED ISSUE #8914: Narrow get_agent_inventory ports filter to listening-state sockets
## Description
`get_agent_inventory`'s numeric `filter` for `kind: ports` (#8910) matches the port on either side of the socket (`source.port` OR `destination.port`), so "what is listening on port 9200?" returns every connection touching 9200 — dozens of rows — instead of the listening socket(s). The answer is still derivable (and was correct in live runs), but the result set is much broader than the question, and data-fidelity scoring against a listening-state control query fails.

The ports inventory documents carry connection state; a "listening on port N" question should prefer state-aware matching.

## Tasks
- [ ] Inspect the live `wazuh-states-inventory-ports*` mapping for the socket-state field and add it to the kind's `_source` list if absent.
- [ ] Narrow the numeric ports filter: match the port AND prefer listening-state sockets when the state field exists (fall back to the current both-sides match otherwise, so non-listening lookups still work).
- [ ] Mention the state semantics in the tool description so the model knows what the filtered result represents.
- [ ] Colocated tests for the outbound query shape with and without the state field.

yenienserrano:
## Description

Closes #8914

Depends on #8924 (branched from `fix/8910-inventory-filter-param`, not yet merged) — this PR targets that branch as its base until #8924 merges, at which point it should be retargeted to `5.0.0`.

`get_agent_inventory`'s numeric `filter` for `kind: ports` (#8910) matched the port on either side of the socket (`source.port` OR `destination.port`), so "what is listening on port 9200?" returned every connection touching 9200 — dozens of rows — instead of the listening socket(s). The answer was still derivable and correct in live runs, but the result set was much broader than the question, and data-fidelity scoring against a listening-state control query failed.

## Proposed Changes

- Narrowed the numeric ports filter to prefer listening-state sockets when the live indexer's `interface.state` field is present (matching `listening`/`listen`, case-insensitive, or falling back when the field is absent so non-listening lookups still work).
- Updated the tool description to state the listening-state semantics of the filtered result.

### Results and Evidence

Colocated unit tests cover the outbound query shape with and without the state field, and case-insensitivity of the state match. No user-facing UI changes.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

`plugins/wazuh-ai-assistant/server/tools/catalog/get-agent-inventory.test.ts`: ports numeric filter narrows on `interface.state` when present; state match is case-insensitive; a document lacking `interface.state` is still returned (fallback).

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -134,6 +134,7 @@
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
 | [#8910](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8910) | Fixed AI Assistant `get_agent_inventory` tool being unable to answer presence questions (e.g. "is a package/port/process present") by adding an optional `filter` parameter                  |
+| [#8914](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8914) | Fixed AI Assistant `get_agent_inventory` numeric `ports` filter matching every connection touching a port instead of preferring the listening socket(s), by narrowing on the `interface.state` field when present |
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


felipegonzalezmv (dismissed):

felipegonzalezmv (dismissed):

rodrigofez (approved):
🟢 LGTM

PR #8937, opened 2026-08-10
Fix AI Assistant returning wrong or incomplete answers on hunt questions

### LINKED ISSUE #8920: AI Assistant returns wrong or incomplete answers on hunt questions: sample-as-population, self-narrowed windows, sub-technique exclusion, false capability denials, dropped metric aggregations, silent host substitution
## Description

A manual SOC-analyst session against build `integration/wave-all-v13` (upstream `a7cca1e71` + the 11
open AI-Assistant branches) produced 30 realistic hunt/triage questions. Every answer was verified
against the live indexer. The assistant fabricated nothing — but it returned **wrong or materially
incomplete answers** to six classes of question, and rendered a populated results table as empty.

These were invisible to the existing 143-cell QA campaign by construction: that harness reads the SSE
stream directly (so it never executes the rendering layer) and scores tool selection, parameters and
grounding — all of which *pass* on most of these defects. The defect is in the truth of the answer
about the **population**, not in the tool call.

Each item below is a **class**, not an instance. The fix should close the class, with a
registry-wide coverage test in the style of `agg-size-coverage.test.ts` /
`field-policy-coverage.test.ts`, so an equivalent corner cannot regress silently.

### 1. The model narrates its sample as if it were the population (HIGH)

"What's listening on this host that shouldn't be?" → *"only three sockets are in a listening state"*
(22/sshd, 443/nginx, 53/systemd-resolve) — exactly the first three visible table rows. Ground truth:
**14 listening sockets**, including `9200 java` (the indexer itself), `1514/1515 wazuh-manager`,
`55000 python3`, `33773 containerd`.

Same class, two more witnesses: an SSH-hardening audit question named 2 of **10** failed SSH checks and
missed `28638 Ensure sshd PermitRootLogin is disabled`; a per-host summary said "no high-severity
alerts" for a host with **2 critical + 2 high** findings.

Root cause: the model sees only the digest's ~5-20 sampled rows while the table shows the population.
The 8 finding-hits tools mitigate this with population-true `FINDING_BREAKDOWN_AGGS`
(`server/tools/catalog/common.ts`); the **inventory and SCA tools have no breakdown aggregation at
all**. Fix structurally, not by asking the model to be careful.

### 2. Absence claimed inside a window the model narrowed itself (HIGH)

"Someone told me there were failed logins from localhost. What's the story?" → *"I couldn't find any"*.
Ground truth: `source.ip: 127.0.0.1` = **0 in the last 24h, 66 in the last 90d**. The findings tools
default to `now-90d`; the model passed `now-24h`, inherited from earlier turns in the same
conversation, then reported absence without disclosing that the window was its own choice.

This changed a recommendation, not just a number: asked to pick a host to isolate, the assistant saw 0
findings in 24h for `web-prod-01` (which has **91 in 90d — 16 critical, 48 high**) and recommended
isolating the other host on vulnerability state alone.

Fix deterministically, tool-side: when a query returns 0 rows in a **model-narrowed** window, re-count
over the tool's default window and return "0 in your window, N in the default window". Extends the
existing zero-row hint. A prompt-only rule is not sufficient here — see #8913, where three
prompt-level attempts failed before a code-level fix worked.

### 3. MITRE technique matching excludes sub-techniques, hiding a real attack (HIGH)

"Show me everything mapped to MITRE T1059." → 3 **informational SCA check *passed*** findings, plus the
assertion *"these are the only T1059-related detections currently recorded."* Ground truth:

```
T1059      3 docs  (informational SCA passes — what it showed)
T1059.001  9 docs  ALL high severity, ALL on win-dc-01:
                   "Suspicious PowerShell encoded command executed"
```

`get_mitre_findings` matches `technique_id` exactly. Sub-technique rollup is standard ATT&CK semantics —
no analyst asking about T1059 means "exclude T1059.001". The fix must apply wherever a technique id is
accepted, and must not silently over-broaden either (disclose exact vs rolled-up counts).

### 4. The assistant denies capabilities that exist (HIGH)

| Question | Claim | Truth |
| --- | --- | --- |
| "How badly are we failing CIS?" | "CIS compliance checks performed via SCA are **not covered by the current tools**" | FALSE — `get_sca_results` exists and the assistant called it on the *next* question. Answer: 102 failed / 95 passed / 10 N-A |
| "What Windows patches are missing?" | "vulnerability data is not filtered by OS", "no Windows-only vulnerability list is available from the tools at hand" | FALSE — vulnerability docs carry `host.os.name` (Ubuntu 570, **Microsoft Windows Server 2019 4**). Its query failed on a wrong field name (`agent.os.name`), which it misdiagnosed as "invalid time range" and escalated into a missing capability |
| "What browser extensions are installed?" | "no available tool reports browser extensions" | TRUE — `get_agent_inventory` supports exactly 5 kinds |

Pattern: **a failed query, or a tool the router did not offer this turn, is reported to the user as a
capability the product does not have.** This is the mirror image of #8919 (which stopped the model
*naming* an unoffered tool; nothing stops it *denying the capability exists*). A false "the product
can't do that" is more damaging than a wrong number — the user stops asking.

### 5. Top-level metric aggregations are dropped, and the tool recommends them (HIGH)

"How many distinct hosts do you actually have data for?" → the model built the correct query **twice**
(`cardinality` on `wazuh.agent.name`), OpenSearch computed it, and the tool reported **0 rows**, so the
model gave up. Ground truth: **6 distinct hosts**.

Root cause: `server/tools/digest.ts` — `bucketsToRows` reads `aggregations[firstKey].buckets` and
returns `undefined` when that is not an array. A top-level metric aggregation is shaped `{value: N}`
with no `buckets`, so the computed answer is discarded. Metric *sub*-aggregations inside buckets do
work.

Aggravating: `server/tools/catalog/search-wazuh-data.ts`'s own description instructs the model to "use
a `cardinality` aggregation" for "how many DISTINCT X" questions — the tool documents the one pattern
its digest cannot return. Every "how many distinct" question is a guaranteed dead end.

### 6. Silent entity substitution on a near-miss agent name (HIGH)

"What's going on with `wazuh-aio-05`?" → the assistant called
`search_findings_by_agent{agent_name: "wazuh-aio-5"}`, silently rewriting the name, and reported
**1,801 findings from a different host** with no mention of the substitution. Ground truth:
`wazuh-aio-05` is its own agent (**id 998**) with 1 finding and **2 vulnerabilities including
CVE-2024-3094 Critical xz-utils**.

Related: "is **our** openssl affected, and what version are we on?" answered for one host and said "on
this host", when openssl vulnerabilities exist on **three** agents. #8913 fixed the single-active-agent
deictic case; the multi-candidate case still guesses silently instead of aggregating or asking.

### 7. A populated results table is overwritten by an empty one (4 sightings)

Prose correctly described 918 findings; the table rendered **0 rows**. Verified the server is
blameless — driving the same tool sequence through a local mock provider shows the server emitting
`rows=20, columns=5` with real values.

Root cause: `public/components/chat/chat-page.tsx`, the stream's `finally` block calls
`flushPendingTable()` and then `flushPendingEmptyTable()`. A message carries a single `message.table`,
so the buffered empty table is committed last and wins. The `error` branch ~60 lines earlier does it in
the opposite (correct) order, which shows this is an ordering slip rather than intent.

Deterministic: it breaks only when a turn's **last** table event is empty (rows→empty = clobbered;
empty→rows = fine). Triage questions trip it because the model pairs a broad query with a
severity-filtered one that returns nothing.

Known residual, tracked separately: a message can only ever display one table, so a multi-tool turn
still shows one of several.

### 8. Privacy mode mints package versions as hostnames

Wire capture in privacy mode: `package.version` `3.118ubuntu5` → `HOST_2` and
`3.20191218.1ubuntu2.3` → `HOST_3`, while `0.21-4ubuntu4` and `11.4.0-1ubuntu1~22.04.3` passed
through intact.

Root cause: the message-boundary FQDN prescan skips tokens matching `VERSION_LIKE_TOKEN_RE`, which
requires every dotted label to be digits-only. A Debian/Ubuntu version whose middle label contains
letters fails that guard, matches the FQDN shape, and is minted as a hostname. The explicit
`package.version: allow` policy entry cannot help — that pass runs at the message boundary over
serialized text, where field context is gone.

Data fidelity, **not** leakage. Naively loosening the regex is how a genuine hostname would escape, so
the fix needs a shape test that is provably hostname-safe.

### 9. Graceful-failure Discover handoffs point at fields that do not exist

Three declines handed over Discover queries referencing `wazuh.module = software_inventory` /
`active-response` (4.x fields, absent in 5.0) and `wazuh.threat_intel` (does not exist; the real field
is `threat.enrichments.*`). In one case the generated DSL contained **only** an `@timestamp` range —
the field named in the prose was not filtered on at all.

The suggested query must be validated against its own reason text and against the target index
mapping, or the handoff hands the user an empty result while claiming to search for something specific.

### 10. Wasted rounds

A duplicate `get_agents` call in one turn with effectively identical arguments (`agent_ids: null` vs
omitted), and a turn that asked the user for permission to run `get_sca_checks` — a tool it had just
named — instead of chaining it, turning "give me the three failed checks that matter most" into a
non-answer.

## Tasks

- [ ] Close each numbered class above, with a registry-wide coverage test per class so an equivalent
      corner cannot regress silently.
- [ ] Population disclosure for every row-returning tool that lacks a breakdown aggregation (inventory,
      SCA), so absence/completeness claims can be made from counts rather than from a sample.
- [ ] Deterministic zero-row + narrowed-window disclosure, tool-side.
- [ ] ATT&CK sub-technique rollup wherever a technique id is accepted.
- [ ] Remove the "capability does not exist" failure mode: a failed query or an unrouted tool must never
      be reported as a missing product capability.
- [ ] Pass top-level metric aggregations through the digest, and keep the tool description and the
      digest capability in sync (a test should fail if the description advertises a shape the digest
      drops).
- [ ] Never substitute a near-miss entity name silently; surface the resolution or ask.
- [ ] Fix the table flush ordering, with a colocated `chat-page` test pinning the rows→empty case.
- [ ] Hostname-safe version-shape handling in the privacy prescan.
- [ ] Validate suggested Discover queries against their reason text and the target mapping.

## Evidence and verification

All ground-truth numbers above were taken from the live indexer during the session
(`wazuh-findings-v5*`, `wazuh-states-*`) and are reproduced in the session findings document. Data
caveat: every vulnerability document in that lab is `[QASEED-SYNTHETIC]`, and 550 of 574 belong to one
bulk-seeded package — vulnerability counts characterize the product on seeded data.

yenienserrano:
## Description

Closes #8920

The AI Assistant (`plugins/wazuh-ai-assistant`) was returning wrong or incomplete
answers to hunt questions because several server-side tools and the digest/prompt
layer silently narrowed, truncated, or mislabeled the data they were reporting on.
This PR closes those correctness gaps across query semantics, population
disclosure, capability honesty, digest metric aggregations, and privacy
pseudonymization, and hardens the test suite that guards them.

## Proposed Changes

- **Sample-as-population conflation**: added population-true breakdowns for
  vulnerability, SCA, inventory, CVE, events and FIM tools so a truncated sample
  is never narrated as the whole result set (`ea6733ee6`, `8d9a6d2eb`,
  `d3f4f525d`, `b8ccecaf6`, `bca2a061b`, `43b46aef3`).
- **Self-narrowed windows**: `suggest_discover_query` and the near-miss entity
  probe no longer silently execute against a narrower time window than the one
  requested; narrowed-window 0-row results are now disclosed instead of reported
  as if the default window was used (`ab480bbec`, `2310f2879`, `6f9184dad`,
  `a5fb2bab6`, `7fbfd6f98`, `fa7872081`).
- **Sub-technique exclusion**: `get_mitre_findings`' `technique_id` filter is
  rolled up to include sub-techniques instead of only exact matches, and the
  `technique_ids` aggregation is scoped back to the requested id
  (`ff58e8bd5`, `93fd98738`, `fd086bad5`, `812b42dbb`).
- **False capability denials**: added a capability-denial guard so a failed tool
  call is retried/routed instead of the model claiming the product lacks a
  capability it actually has; closed the resulting capability-honesty
  test-integrity gaps (`4e2437266`, `e49e35a78`, `0cff06aa3`, `18d6c87c5`).
- **Dropped metric aggregations**: `digest.ts` no longer drops top-level metric
  aggregations, bucket truncation is now reported per dimension instead of
  summed, and `Digest.columns` is restated as the authoritative row-schema hint
  (`030358764`, `3756eb0be`, `582a4a952`, `60a1070c7`, `281f3b0e3`).
- **Silent host/agent substitution**: agent-name near-misses are now disclosed
  to the user instead of silently substituting the closest match, and
  user-supplied identifiers are preserved verbatim in the prompt
  (`5c17d6c87`, `33edb92ab`).
- **Privacy pseudonymization false positives**: fixed the privacy version-shape
  regex incorrectly pseudonymizing Debian package versions and MITRE
  sub-technique ids as hostnames, and classified `package.vendor` correctly
  (`103a5811a`, `cffb469c8`, `8ba016f4b`).
- **UI table flush**: fixed an empty result table clobbering a previously
  populated one at the end of a streamed answer (`132ade3fe`, `14856b8a9`).
- **Probe-window integrity**: the near-miss probe used for entity resolution no
  longer inherits the executed range, and its invariant is asserted directly
  (against the range bound) instead of via a loose `match_all` check
  (`6f9184dad`, `a5fb2bab6`, `7fbfd6f98`).
- **Test-integrity hardening**: fixed tautological/registry-sweep assertions
  that could pass without exercising the code they claimed to cover, and added
  missing coverage for `discover-url.ts` time-range branches and
  technique-rollup coverage (`9854bce4d`, `812b42dbb`, `3b826c516`, `34a5ebe53`,
  `82d5a7640`).

All changes are confined to `plugins/wazuh-ai-assistant` (`server/tools/*` and
their colocated `*.test.ts` files) plus the `CHANGELOG.md` entries for #8920.

### Results and Evidence

This is **not a UI change** — every commit on this branch touches
`plugins/wazuh-ai-assistant/server/tools/*` (prompt construction, tool
execution, digest building, privacy pseudonymization) and their colocated
Jest tests; no `public/` React/EUI code was modified. Accordingly no
screenshot or video is attached.

Evidence so far is the expanded/added Jest test suites colocated with each
fix (`digest.test.ts`, `executor.test.ts`, `privacy.test.ts`,
`suggest-discover-query.test.ts`, `window-recount.test.ts`,
`entity-resolution.test.ts`, `guardrails.test.ts`,
`field-policy-coverage.test.ts`, `router.test.ts`). `yarn test:jest` and
`yarn lint` for `plugins/wazuh-ai-assistant` were **not run as part of this
PR prep** — the docker dev container available on this host has a different
checkout mounted, and `node_modules` are not installed in this worktree.
**Flag for the developer**: run `yarn test:jest --runInBand` and `yarn lint`
inside `docker/osd-dev` for `plugins/wazuh-ai-assistant` before moving this
out of Draft, and add before/after transcript evidence of the AI Assistant's
answers if reviewers want to see the behavioral fix directly (e.g. a hunt
question that previously returned a truncated-sample-as-population or a
false capability denial, now answering correctly).

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (server tools only): `digest.ts`,
  `entity-resolution.ts`, `executor.ts`, `guardrails.ts`, `privacy.ts`,
  `router.ts`, `suggest-discover-query.ts`, `technique-rollup.ts`,
  `window-recount.ts`, `types.ts`, and their colocated test files.

### Configuration Changes

None.

### Documentation Updates

None beyond the `CHANGELOG.md` entries for #8920.

### Tests Introduced

- New/expanded Jest suites colocated with each fix: `digest.test.ts`,
  `executor.test.ts`, `privacy.test.ts`, `suggest-discover-query.test.ts`,
  `entity-resolution.test.ts` (new), `window-recount.test.ts` (new),
  `guardrails.test.ts`, `field-policy-coverage.test.ts`, `router.test.ts`.
- Coverage added for previously untested branches: `discover-url.ts`
  time-range handling and `technique-rollup` coverage
  (`3b826c516`, `812b42dbb`).
- Test-integrity fixes so existing assertions actually exercise the guarded
  behavior instead of passing tautologically (`9854bce4d`, `18d6c87c5`,
  `82d5a7640`, `34a5ebe53`).

## Review Checklist

- [ ] Code changes reviewed
- [ ] `yarn lint` and `yarn test:jest` run inside `docker/osd-dev` for
      `plugins/wazuh-ai-assistant`
- [ ] 9 merge commits on this branch are missing DCO `Signed-off-by` trailers
      and need remediation before merge (see PR description/comments)


rodrigofez (approved):
🟢 CR

1. Sample-as-population conflation: ✅ Implemented, with coverage test.
2. Self-narrowed window disclosure: ✅ Implemented, with coverage test.
3. MITRE sub-technique rollup: ✅ Implemented, with coverage test.
4. False capability denials: ✅ Implemented, with coverage test.
5. Dropped metric aggregation: ✅ Implemented, with coverage test.
6. Silent entity substitution: ⚠️ Partial. Single near-miss name case is implemented, but the multi-candidate case (e.g. same package found on multiple agents) isn't.
7. Table flush ordering (empty table clobbering results): ✅ Implemented, with coverage test.
8. Privacy mode mis-tagging package versions as hostnames: ✅ Implemented, with coverage test.
9. Discover handoff field validation: ✅ Implemented, but no coverage test (unlike the others).

rodrigofez (approved):

PR #8938, opened 2026-08-10
Redesign AI Assistant result-table columns for analyst value

### LINKED ISSUE #8921: AI Assistant result tables: design columns for analyst value (bucket/sample mispairing states falsehoods, id-first ordering, missing severity, no column budget)
## Description

The AI Assistant's result tables were derived from each tool's query shape — the projection list and the
aggregation form — and never designed for the analyst reading them. A manual SOC-analyst session on
build v13 surfaced three concrete consequences, one of which makes the assistant state something false.

This issue covers the **column design**; the answer-correctness defects found in the same session are
tracked separately.

### Defect 1 — a sampled label paired with a bucket count produces false statements

"Give me the five noisiest rules right now" returned:

> "Wazuh VD – Vulnerability CVE-2026-43492 remediated — 796 hits"

Verified against the live indexer, over the same window:

| Measurement | Value |
| --- | --- |
| documents carrying that exact title | **1** |
| documents under rule id `61f16843-5b00-41bf-beee-700013e62575` | **796** |
| distinct `wazuh.rule.title` values under that one rule id | **772** |

The count is real and the title is real; the **pairing** is false. `get_top_rules` runs a `terms`
aggregation on `wazuh.rule.id` with `top_hits: { size: 1 }` sampling one title per bucket. That was
sound in 4.x, where a rule id had one title. In 5.0 rules are Sigma-based and the title is templated per
CVE, so one id spans hundreds of titles.

**A sampled label may only be shown where the key determines the label.** Adopting that rule requires an
audit of every tool using the pattern — and it is not one tool:

| Tool | key → label | 1:1? |
| --- | --- | --- |
| `get_top_agents` | `wazuh.agent.id` → `wazuh.agent.name` | yes in current data, but empirical only — a renamed agent breaks it |
| `get_mitre_summary` | technique id → technique name | **NO** |
| `get_top_rules` | `wazuh.rule.id` → `wazuh.rule.title` | **NO — 1:772** |

`get_mitre_summary` is the worse case and it is in the **digest** (the surface that produces prose, not
just the table): `wazuh.rule.mitre.technique.{id,name}` are **arrays**. The T1005 bucket (57 docs)
samples a seven-technique document, so one count is paired with seven names; 16 of 20 buckets carry 2-7
names. The sampled `_source` omits `technique.id`, so nothing downstream can even positionally match the
right name to its bucket. This renders visibly in the UI today — one cell reads
`T1562, T1070, T1059, T1040, T1562.002, T1070.001` with six names and three tactics.

### Defect 2 — identity takes the most prominent column, meaning is demoted

Every aggregation table reads `identifier | count | name`. The reader scans left first and meets the
least actionable cell. Proposed order: **meaning → magnitude → spread → identity (last, or in the row
expander)**.

Test for whether an identifier earns a visible column: *would an analyst type it, search it, or paste it
into a ticket?* CVE id, agent id, MITRE technique id and SCA check id all pass. A Sigma rule UUID does
not — and it was reported as a suspected bug during the session ("rule id is not correct apparently"),
because analysts expect a numeric `5715`-style id.

Correction to an earlier draft of this issue, worth recording: `wazuh.rule.id` is **not** always a UUID.
Over the tool's default 90-day window, **12 of 17** rule-id buckets are classic numeric ids (`5710`,
`31169`, `92213`, …), and such documents carry no `sigma_id` at all. So the id column does carry value
most of the time. It should be **demoted to a last column or the row expander, not removed** — and it
must stay in `digest.sampleColumns`, because the system prompt directs the model to aggregate by rule
first to discover ids for `find_document_by_field`.

### Defect 3 — the noise-triage table cannot show severity

`get_top_rules` has no level column, so the one table whose job is separating noise from signal cannot
show which rules are dangerous. The model therefore editorialises ("both are routine background noise")
from the title alone. In the session's data both noisy rules are `informational` while every `high` /
`critical` rule is low-volume — severity is precisely the discriminator.

Implementation note: `max` on `wazuh.rule.level` **does not work** — it is a keyword field, and a live
attempt returns `illegal_argument_exception: Field [wazuh.rule.level] of type [keyword] is not supported
for aggregation [max]`. Scripted aggregations are banned by the guardrails. A `terms` sub-aggregation on
the level (allowlisted) with the ordering computed in Node from the severity list is the workable route.

### Defect 4 — wide tables force a horizontal scrollbar and clip data

The ports table declares **8 visible columns** (`source.ip`, `source.port`, `destination.ip`,
`destination.port`, `network.transport`, `interface.state`, `process.name`, `process.pid`), which
overflows the chat bubble: a scrollbar appears and the leading column is clipped mid-value (`0.0.0.0`
renders as `.0.0.0`), which reads as corrupt data rather than as truncation.

A uniform column count across all tools is the wrong remedy — a vulnerability row and a FIM row
genuinely need different fields. The invariant should be that **no table needs a horizontal scrollbar**,
achieved with a per-tool column budget plus the existing `tableSpec.rowFields` row-expander mechanism.

Proposed for ports — 5 visible (`source.port`, `interface.state`, `process.name`,
`network.transport`, `destination.ip`), with `source.ip`, `destination.port`, `process.pid` in the
expander. Note `destination.*` must be **demoted, not deleted**: it is `0.0.0.0`/`0` noise on listening
rows, but on established connections it carries the peer (40 connections to 9200, several to external
addresses) — exactly what a hunter pivots on.

### Defect 5 — absent values render as blank cells

`interface.state` is absent for UDP sockets and `host.os.full` is absent on one OS record; both render as
an empty cell, which reads as a rendering glitch rather than as "no value". An explicit placeholder
(`—`) removes the ambiguity.

## Tasks

- [ ] Adopt and document the rule: a sampled bucket label may only be displayed where the key determines
      the label; otherwise carry the spread (a `cardinality` sub-aggregation) and mark the label as a
      sample.
- [ ] `get_top_rules`: add distinct-title count and a severity column; relabel the title as a sample;
      demote the rule id.
- [ ] `get_mitre_summary`: stop pairing a bucket count with an arbitrary member of a multi-value array —
      at minimum include `technique.id` in the sample and match the bucket's own key server-side.
- [ ] Cheap cardinality guard on `get_top_agents`, whose 1:1 relationship is empirical rather than
      guaranteed.
- [ ] Column-budget rule (no horizontal scrollbar at chat-bubble width) using `tableSpec.rowFields` for
      overflow; apply to the ports table and the wide vulnerability tables.
- [ ] Placeholder rendering for absent values.
- [ ] Keep `digest.sampleColumns` intact where the prompt's aggregate-then-lookup workflow depends on it.

## Explicitly out of scope

- **A derived SCA "Score %" column.** `get_sca_results` records a deliberate decision not to bake in a
  score formula because of the N/A-in-denominator ambiguity (95/197 = 48.2% vs 95/207 = 45.9%), and the
  tool description already hands the model the formula. Overturning that needs its own discussion, and
  there is no derived-column mechanism today.
- **Aggregating `get_top_rules` on `wazuh.rule.title` directly.** Permitted by the guardrails, but it
  would atomize one 796-document rule into ~772 single-document buckets and hide the "one rule is
  spamming the console" signal that motivated this issue.

## Implementation trap

`field-policy-coverage.test.ts` will **fail** for any new digest sample column whose name is in neither
`FIELD_POLICY_DEFAULTS` nor `KNOWN_SAFE_STRUCTURAL_FIELDS`. New sub-aggregation names must be added
there deliberately. That list also still carries stale 4.14 SCA entries (`policy_id`, `pass`, `fail`,
`score`), so edit it with care.

yenienserrano:
## Description

The AI Assistant's result tables (`plugins/wazuh-ai-assistant`) surfaced several
analyst-value problems: sampled labels (e.g. rule title, MITRE technique/tactic
name) were shown as if they were the bucket's one true label even though the
underlying terms aggregation only samples a single document per bucket, some
tools omitted severity entirely or pushed it out of the visible columns, column
order buried the fields analysts scan first behind less useful ones, and there
was no shared budget/placeholder convention for how many columns a result table
shows before the rest is only reachable via the row expander.

Closes #8921

## Proposed Changes

- **Sampled-label spread disclosure**: `get_top_rules` and `get_mitre_summary`
  now expose the spread behind a sampled label (a `distinct_titles`/
  `distinct_names`/`distinct_tactics` cardinality signal and, for
  `get_top_rules`, a data-driven `high_or_critical` severity signal) instead of
  presenting one sampled title/name as if it were authoritative for the whole
  bucket.
- **Registry-wide sampled-label coverage test**: a new test
  (`sampled-label-coverage.test.ts`) walks every registered tool's own
  `buildRequest` output and fails any terms/composite/multi-terms aggregation
  that samples a displayed field via `top_hits` without either a cardinality
  guard or a written, field-scoped justification in
  `SAMPLED_LABEL_ONE_TO_ONE` — so a future tool with this shape fails
  automatically unless it is explicitly justified.
- **Column budget + absent-value rendering**: introduced
  `MAX_VISIBLE_RESULT_COLUMNS` (`common/types.ts`) enforced once in the generic
  `ResultTable` renderer — only the first 6 `tableSpec` columns become visible
  table columns (demoted columns stay queried and reachable via the row
  expander; the accordion summary notes "(+N more fields per row)" whenever
  columns are hidden). Absent values (`undefined`/`null`/`''`) now render as a
  subdued `--` placeholder instead of a blank cell or empty severity badge;
  `0`/`false` are not treated as absent.
- **Registry-wide visible-column-budget coverage test**
  (`visible-column-budget-coverage.test.ts`) asserts every tool's severity
  column sits inside the budget — nothing is exempt.
- **Column reordering for the visibility budget** (identity fields demoted, not
  deleted): `get_agent_inventory` ports, the four vulnerability tools
  (`get_vulnerabilities`, `get_vulnerabilities_by_agent`,
  `get_vulnerability_by_cve`, `get_critical_vulnerabilities`),
  `get_mitre_findings`, and `get_rules` now lead with the meaning →
  magnitude → spread → identity fields analysts scan first; numeric/internal
  IDs move to the last visible or expander-only positions.
- **`renderDefaultCell` correctness fix**: booleans and arrays were previously
  passed as raw React children, which silently rendered nothing (`false`) or
  concatenated with no separator (arrays). Booleans now render `Yes`/`No`,
  arrays comma-join, and objects render as JSON; the string/number fast path
  is unchanged.
- Housekeeping: merged a duplicated `common/types` import and fixed a test
  that matched the mobile *and* desktop header cell of the same column label
  (now asserts on the header cell only).

### Artifacts Affected

- `plugins/wazuh-ai-assistant` only (server-side tool `buildRequest`/digest
  logic and the client-side `ResultTable` renderer). No other plugin is
  touched.

### Configuration Changes

None. `MAX_VISIBLE_RESULT_COLUMNS` is a fixed constant, not a user-facing
setting.

### Documentation Updates

None required; behavior is scoped to the AI Assistant's internal result-table
rendering and tool digest logic, not documented end-user configuration.

### Tests Introduced

- `get-top-rules.test.ts`, `get-mitre-summary.test.ts`: new/updated cases
  covering the sampled-label spread signals and reordered columns.
- `sampled-label-coverage.test.ts` (new, 392 lines): registry-wide walker that
  fails any tool sampling a displayed field without a cardinality guard or
  recorded justification; includes a mechanism self-test against synthetic
  offenders.
- `visible-column-budget-coverage.test.ts` (new): registry-wide assertion
  that every tool's severity column is inside the visible-column budget.
- `field-policy-coverage.test.ts`: extended for the new fields.
- Updated existing suites (`executor.test.ts`, `guardrails.test.ts`,
  `privacy.test.ts`, `router.test.ts`, `suggest-discover-query.test.ts`,
  `get-sca-results.test.ts`, `get-threat-intel-components.test.ts`,
  `search-wazuh-data.test.ts`) to match the reordered/renamed fields and the
  relaxed byte-for-byte ordering assertion (contents/index/size checked
  instead of strict order, where the order change is intentional).

## Review Checklist

- [x] Code changes reviewed
- [ ] Relevant evidence provided (screenshot/video still needed — see
  "Results and Evidence")
- [x] Configuration changes documented (none applicable)
- [x] Developer documentation reflects the changes (none applicable)
- [ ] ...


rodrigofez (approved):
Reviewed against issue #8921's task list. Most items are covered well:

- Sampled-label rule adopted and enforced with a registry-wide test.
- get_top_rules: distinct-title count, severity signal, "(sample)" label, rule ID demoted.
- Column budget (6 visible columns) with row-expander overflow, enforced registry-wide.
- Absent values now render as --.
- Ports and vulnerability tables reordered to fit the budget.

Possible gap:

1. get_mitre_summary: the issue asks to match the bucket's own key server-side, not just carry technique.id along for later. Right now technique.id is added to the sample, but nothing actually matches it to the bucket key, the table still shows the raw, unmatched sampled name.

PR #8952, opened 2026-08-12
Fix AI Assistant provider compatibility issues

### LINKED ISSUE #8947: AI Assistant providers fail or mislead with several supported vendors (temperature, Gemini tool round-trip, DeepSeek output, provider form guidance)
| Wazuh | Rev  | Browser |
|-------|------|---------|
| 5.0.0 | 5.0.0-00 (OSD 3.6.0) | N/A (mostly server-side) |

**Description**

This issue groups four separate AI Assistant provider bugs found during a live test pass on
Wazuh 5.0.0 (2026-08-11/12). Each one makes a supported provider fail, or makes the settings
form point the administrator at the wrong provider or model. Grouped together because they all
sit in the same provider layer (`plugins/wazuh-ai-assistant/server/providers/*` and the provider
settings form) and were found in the same test pass.

1. Anthropic providers never answer.
2. Gemini providers fail on any question that needs a tool.
3. DeepSeek answers contain visible internal markup.
4. The provider settings form gives wrong or unsafe guidance.

**Preconditions**

1. An AI Assistant provider configured for one of: Anthropic (`anthropic` type), Gemini or
   DeepSeek (both `openai_compatible` type), with a valid API key.
2. For part 1, the provider type must really be `anthropic`. The same base URL stored as
   `openai_compatible` fails differently (empty HTTP 404) and hides this bug.

**Steps to reproduce**

See each part below. All four were reproduced with: open AI Assistant Settings, add the
provider, press Test, make it default, open Chat, ask a real question (for example "How many
agents are active?" or "List critical findings").

**Expected Result**

1. Each of the four supported setups above answers the question correctly, using real tool
   calls, with no leftover internal text in the answer, and with settings guidance that matches
   what actually works.

**Actual Result**

See each part below.

---

### Part 1 — Anthropic providers cannot answer any question

**Symptom.** Every turn ends with no answer. The chat window shows only "Response interrupted".

**Cause.** The chat route always sends `temperature` on the request. Anthropic rejects this
parameter with HTTP 400 on Claude Opus 4.7 and later (Opus 4.7, 4.8, 5, Sonnet 5, Fable 5). A
retry-without-`temperature` already existed, but only inside `openai-compatible.ts` — it was
never applied to `anthropic.ts`. So the very first request of every turn is rejected, before any
tool can run.

**Fix.** Move the retry into a shared module,
`plugins/wazuh-ai-assistant/server/providers/temperature-fallback.ts`, used by both adapters,
with one shared rejection cache keyed by base URL + model. The check for "this is a temperature
rejection" is intentionally loose (any 4xx whose body mentions "temperature"), because different
vendors word the error differently. A false positive just costs one harmless retry; a false
negative loses the whole turn.

**Evidence** (measured live against the real Anthropic API):

| Build | Provider test | Chat turn |
|---|---|---|
| before the fix | success, 1092 ms | `HTTP 400: temperature is deprecated for this model.` |
| after the fix | success, 1395 ms | answered, with real tool calls: `get_critical_findings` 7592 ms, `get_agents` 7200 ms |

After the fix, all 8 current Claude models were verified end to end, each calling a real tool:
`claude-opus-4-8`, `claude-opus-4-7`, `claude-opus-4-6`, `claude-opus-5`, `claude-sonnet-5`,
`claude-sonnet-4-6`, `claude-haiku-4-5`, `claude-fable-5`.

Note: a green provider Test does not catch this. The test request never sends `temperature`, so
Test passes while every real chat turn fails.

---

### Part 2 — Gemini providers fail on every question that uses a tool

**Symptom.** Any question that needs a tool call fails with HTTP 400: "Function call is missing
a thought_signature." Reproduced 4 out of 4 times on both `gemini-flash-latest` and
`gemini-3-flash-preview`.

**Cause.** Gemini attaches a `thought_signature` field to its tool-call message and requires the
client to send it back unchanged on the next turn. The adapter rebuilds the conversation history
from its own internal shape and drops fields it does not know about, so `thought_signature` is
lost before the reply goes back to Gemini.

**Fix.** Add a generic vendor-extras pass-through: capture any unknown field on a streamed
tool-call message and echo it back verbatim on the next request. This is a no-op for vendors
that do not send such a field. Also relax the chat and conversations route schemas to accept
these extra optional fields — the schemas were strict and rejected the replayed history before
the pass-through fix could even take effect.

**Evidence.** Before the fix: HTTP 400 on every tool-using question, 4/4 reproducible. After the
fix: the 400 is gone and the tool round-trip completes normally on both Gemini models above.

---

### Part 3 — DeepSeek answers show internal model markup

**Symptom.** Answers from DeepSeek providers contain visible tokens that should never reach the
user, for example `<｜DSML｜function_calls` (using fullwidth vertical bar characters). These
tokens are DeepSeek's own internal function-call markup, leaking into the visible answer text.

**Fix.** Extend the existing inline reasoning markup filter (the one that already strips other
providers' internal reasoning tags) to also strip these DeepSeek tokens. The filter must stay
streaming-safe, since tokens can arrive split across chunks.

**Evidence.** Before the fix: the tokens appear in every DeepSeek answer. After the fix: zero
occurrences across repeated tests, and the visible answer text is unchanged and still correct.

---

### Part 4 — The provider settings form gives wrong or unsafe guidance

**Symptom.** The form that an administrator fills in to add a provider (`provider-form-flyout.tsx`)
contains several pieces of wrong or misleading text:

- It lists Gemini and Groq as working examples. Gemini was broken until Part 2 above. Groq
  rejects the assistant's tool definitions with HTTP 413 on common account tiers, so it cannot
  answer any real question.
- For the `openai_compatible` type it shows Claude model names as example models. This
  combination is impossible: Claude only works through the `anthropic` provider type.
- Its Anthropic model-id examples are outdated ids that no longer match current models.
- It says small models "often fail." What was actually measured is worse: models without
  tool-calling support (tested with `gemma-3-27b` and `qwen3-235b`) do not fail visibly. They
  answer fluently but invent security data — fake alerts, fake agents, fake numbers — because
  they cannot call the tools that would fetch real data. This is silent and easy to mistake for a
  working answer.

**Fix.** Correct the example text described above; show suggested model chips filtered by the
selected provider type, so Claude names never appear under `openai_compatible`; and add a clear
warning that a model without tool-calling support may produce a confident but fabricated answer,
instead of the current "often fail" wording.

**Also document:** a green provider Test does not guarantee chat works. This was measured on
three separate provider/model combinations in this test pass (see Part 1's Test-vs-Chat gap as
one example).

---

**Screenshots**

Not applicable — this is a server-side issue reproduced through the chat window; no visual UI
defect is involved beyond the form text in Part 4.

**Additional context**

Related but NOT fixed here, worth separate issues:
- `max_tokens` for `anthropic` providers is hardcoded at 4096 and cannot be changed in the UI. On
  models where thinking is on by default (Opus 5, Sonnet 5, Fable 5) this ceiling is shared
  between reasoning and the answer, so an answer can be cut short. Not observed in this run, but
  unguarded.
- The adapter never surfaces `stop_reason: max_tokens`, so a truncated answer looks like a short
  one instead of a cut-off one.

### Verified support matrix (for reference)

| Provider / model | Status | Reason |
|---|---|---|
| Anthropic, all 8 current Claude models | Supported | Recommend `claude-haiku-4-5` for speed, `claude-opus-4-8` as the balanced default |
| `openai_compatible`: `openai/gpt-oss-120b` | Supported | Verified via a Bedrock gateway endpoint |
| `openai_compatible`: `mistral-large-3` | Supported | |
| `openai_compatible`: `qwen3-32b` | Supported | |
| `openai_compatible`: `qwen3-coder` | Supported | |
| `openai_compatible`: `deepseek v3.2` | Supported | After the Part 3 markup fix |
| `openai_compatible`: `gemini-flash-latest`, `gemini-3-flash-preview` | Supported | After the Part 2 tool-call fix |
| Groq (any model) | Not supported | HTTP 413 on the assistant's tool schemas, common account tiers |
| Claude via any `openai_compatible` gateway | Not supported | Impossible combination; use the `anthropic` provider type instead |
| Models without tool calling (e.g. `gemma-3-27b`, `qwen3-235b`) | Not supported | Fabricates security findings instead of failing visibly |
| OpenRouter free tier | Not supported | Too slow for real use (2-3 minutes per answer) |

yenienserrano:
## Description

Fixes several AI Assistant provider-compatibility issues (#8947):

- Shares the temperature-rejection fallback with Anthropic
- Passes through Gemini `thought_signature` on tool replays
- Strips DeepSeek DSML gateway markup from answers
- Fixes provider copy and adds per-vendor model chips
- Adds the corresponding CHANGELOG entry

## How to Test

- Configure each affected provider (Anthropic, Gemini, DeepSeek) and verify chat/tool-call round-trips complete without errors or leaked gateway markup.
- Verify the provider configuration form shows corrected copy and per-vendor model chips.


Ripdiegozz (approved):
## 🟢 Tests

Could not test the DeepSeek, Gemini, or (native) Anthropic-model round-trips end-to-end. Tested the UI-only side instead:

- Per-vendor model chips

- Provider type copy/descriptions.

https://github.com/user-attachments/assets/cebd9def-66c3-49dd-9ed6-d315b6c31de5

PR #8953, opened 2026-08-12
Fix AI UX and provider usability issues

### LINKED ISSUE #8944: AI Assistant: settings show permanent errors, key setup is unclear, and some providers fail
| Wazuh | Rev  | Browser |
|-------|------|---------|
| 5.0.0 | 5.0.0-04 | Chrome |

**Description**

Five usability problems in the AI Assistant, found during manual testing and review.

1. **Settings shows permanent error banners.** Opening Settings runs a silent connection test on every provider. Any failure becomes a page-level red banner. Dismissing it does not last: the banner returns on every visit, even for providers the user does not use.
2. **API key setup is unclear.** The provider form does not say where to get a key, what shape it should have, or which endpoint URL to use for each provider type.
3. **The welcome screen is cut off.** On shorter windows the example cards and the message box overlap, and part of the content is hidden behind the input panel.
4. **The AI Assistant has its own top-level navigation section** for a single app.
5. **Some providers fail on every question.** Providers that reject the `temperature` parameter (for example `openai.gpt-oss-120b` on a Bedrock-compatible endpoint) return `HTTP 400: temperature is deprecated for this model`. Every turn fails, including the internal routing call.

**Preconditions**

1. AI Assistant enabled with at least one provider configured.

**Steps to reproduce**

1. Add a provider with an invalid API key. Open **Settings**. A red banner appears. Dismiss it, go to **Chat**, return to **Settings**. The banner is back.
2. Open **Add provider** and select type **Anthropic**. No guidance is shown on where to obtain a key or which endpoint URL to use.
3. Open **Chat** with no conversation, in a window under about 700 px tall. The message box and the welcome cards overlap.
4. Open the left navigation. A separate section exists for a single app.
5. Configure an OpenAI-compatible provider with a model that rejects `temperature`. Ask any question.

**Expected Result**

1. No page-level error banner when opening Settings. Provider status is shown per row. A banner appears only after the user clicks **Test**, or when the default provider is failing, and it can be dismissed.
2. The form states where to get the key, its expected format, and the correct endpoint URL for the selected type.
3. The message box and the welcome content never overlap, at any window size.
4. The AI Assistant appears inside an existing navigation section, with an entry point on the Home overview.
5. Optional parameters are not required. If a provider rejects `temperature`, the request is retried without it and that result is remembered for the provider and model.

**Actual Result**

1. The banner reappears on every visit to Settings and can only be cleared by reloading the page.
2. No guidance is shown, so a wrong endpoint URL or key format is only detected after saving and testing.
3. The welcome content is partially hidden behind the message box.
4. A navigation section exists for one app.
5. `HTTP 400: temperature is deprecated for this model` on every turn, so the assistant is unusable with those providers.

**Additional context**

Proposed fix branch: `fix/ai-ux-and-provider-usability`

Two related result-table fixes are on `enhancement/8921-followup-table-fixes`, a follow-up to the merged #8938: the ports table showed a horizontal scrollbar, and the top rules table had no severity column.

yenienserrano:
## Description

Fixes UX and usability issues in the AI Assistant and its provider configuration flow:

- Provider settings: clearer Anthropic API key onboarding, primary "Test" action with deduped status banners, fixed row-actions/test-button lookups, retry support for temperature-rejecting providers without resending the param, and no more permanent error banners on the AI settings page.
- Chat surface: composer/transcript layout fixes (sticky composer no longer overlaps content), manual retry on save-failed callouts, blocking mid-stream retry to prevent double-create, and onboarding shape-warning/type-switch fixes.
- Navigation: AI Assistant entry point added to the Home overview; dedicated AI nav section dropped in favor of joining the Home category.
- Assorted test fixes and Prettier formatting.

## How to Test

Run the AI Assistant provider configuration and chat flows manually; run `yarn test:jest` in `plugins/main` for the affected unit tests.


Ripdiegozz (changes requested):
### Section 1: Provider (Anthropic / Test / banners)

- ✅ API key onboarding: clear flow for a new Anthropic provider.
- ✅ Test button doesn't duplicate banners on repeated clicks.
- ✅ Row actions act on the correct provider when there are multiple.
- ✅ Error banner is closable, doesn't stay stuck permanently.

#### **This callout list could be improved**
Image

#### 🔴 Blocking bug: Test button gets stuck loading forever after creating/updating a provider

**Repro:** create (or update) any provider and save.

**Cause (traced in `settings-page.tsx`):** `handleSubmit` calls the auto-probe as `handleTest(saved)` (line 526), omitting the required second argument `{ manual: boolean }`.

https://github.com/user-attachments/assets/261e8e31-7398-495c-be29-0e9f0489e3ce

### Section 2: Chat surface

- ✅ Composer does not overlap the last message when scrolled to the bottom of a long conversation.
- ✅ Mid-stream retry is correctly blocked/handled — no duplicated turn or second conversation created.

https://github.com/user-attachments/assets/80b81aef-a7a4-4593-82ed-cf8378852078

### Section 3: Navigation

- ⬜ Home overview shows a new CTA/entry point to the AI Assistant.
- ⬜ "AI Assistant" no longer has its own dedicated nav section — it now lives under the Home category.

### Open question: is this navigation change actually wanted?

Image

Image

Ripdiegozz (approved):
Section 1: Provider (Anthropic / Test / banners)

https://github.com/user-attachments/assets/5adedf91-b69e-419b-97fd-b74375a07c84

Issue #8913, opened 2026-08-08
Resolve deictic host references ("this box") in the AI Assistant


MiguelCasaresRobles:
## Description
Questions that reference the current host deictically — "what packages are installed on this box?", "is this machine up to date?" — produce no tool call: the model asks the user for an agent name or id instead (measured live: 9 of 10 runs across two builds, N=5 each). `get_agent_inventory`'s description already tells the model to call `get_agents` first when the agent is not named, but a bare "this box" with no conversational anchor does not trigger that path.

Asking for clarification is not wrong, but in the very common single-agent (or few-agent) deployment the assistant can do better: resolve the reference itself and say what it did.

## Tasks
- [ ] Add explicit deictic guidance to the system prompt / tool descriptions: when the user says "this box/host/machine/server" with no prior agent context, call `get_agents` first; if exactly one active agent exists, use it and state the assumption in the answer; otherwise list the candidates and ask.
- [ ] Colocated test pinning the guidance text so a reword fails loudly.
- [ ] Verify live: the deictic phrasing resolves on a single-agent deployment and still asks (with candidates) on a multi-agent one.

MiguelCasaresRobles:
## Live verification: partially resolved, routing gap remains open

Two fixes were attempted and measured on a real deployment (N=5 per attempt, same question verbatim: "What software does this box have installed?", single active agent in the environment).

**Attempt 1 — prompt instruction only** (telling the model to call `get_agents` first for a deictic reference): 0/5. The instruction was verified present in the deployed bundle; the model asked the user to name an agent in 4/5 runs and used `search_wazuh_data` in the 5th.

**Attempt 2 — server-side resolution** (current branch state): `get_agent_inventory` now resolves the agent itself when no `agent_id`/`agent_name` is supplied and exactly one active agent exists, and surfaces the assumption in the tool result so the answer can state it. This is unit-tested and correct in isolation, but still 0/5 live — because `get_agent_inventory` is never called at all for this phrasing (3/5 ask the user to clarify, 2/5 reach for `search_wazuh_data`). The hardening only engages once the tool is chosen.

**Remaining work is tool selection, not parameter resolution.** A durable fix needs the routing layer to steer a bare deictic inventory question to the inventory tool, or a mechanism that does not depend on the model choosing correctly.

[...]

PR #8955, opened 2026-08-12
Fix AI Assistant answers being silently thinned on enumeration/completeness questions

### LINKED ISSUE #8935: AI Assistant: answer completeness — capability honesty and tool chaining in code, and raise the digest bucket budget so enumeration answers are complete
## Description

Follow-up to #8920 and #8921, scoped from **measured** results rather than estimates. On build v14
(v13 + both branches, jest 1205/1205, deployed, 26/26 markers) a 42-run live proof batch against a real
provider produced one clean split:

> **Every fix implemented in code passed 3/3. Both fixes implemented at the prompt layer failed 0/3.**

Deterministic wins: MITRE sub-technique rollup, agent near-miss disclosure, port 9200/java, privacy
version survival, `get_top_rules` count/title truth, narrowed-window recount, metric aggregations
answering at all. Prompt-layer losses: capability honesty, and chaining the follow-up tool.

This issue closes those two at the layer that works, and removes the remaining reason answers feel
thin. **Guiding principle: never cut the assistant off by timeframe or document count. When a bound
binds, disclose it — never let it silently thin an answer.**

Nothing here requires sending more data to the provider. The model continues to read **5 sample rows**;
completeness comes from aggregations, which are computed over every matched document and returned as a
handful of `{key, count}` pairs (~1,000 tokens for 100 buckets).

## 1. Capability honesty, in code (not in the prompt)

**Measured:** asked "How badly are we failing CIS, in plain numbers?", the assistant called **no tool at
all** and answered *"none of the available compliance-framework tools include a 'cis' tag"*. Ground
truth: **102 failed / 95 passed / 10 not-applicable** on `cis_ubuntu22-04`. 0/3 runs. It then used
`get_sca_results` successfully on the very next question.

**Why #8920's fix did not reach it:** `augmentToolError` annotates a tool **error**. A no-tool answer has
no error to annotate, so the deterministic half covers only the wrong-field-query witnesses. The headline
case rested on prompt text, and prompt text lost — the fourth time this project has learned that (#8913
needed three prompt attempts before a code-level fix worked).

Direction:

- Routing must reach `sca` for compliance/CIS/benchmark/hardening vocabulary. Verify by instrumenting
  the stage-1 decision, exactly as #8913's root cause was finally found — do not guess at prompt wording.
- A capability claim must be **checkable against the registry**, not asserted from what the model can see
  this turn. A "no tool covers this" statement should be derivable (the tool catalog plus each tool's
  parameter enums are known server-side), so a false denial becomes impossible rather than discouraged.

## 2. Tool chaining, in code

**Measured:** "My auditor wants proof of SSH hardening." The assistant returned the SCA summary —
95/102/10, **exactly right** — then said the summary does not break out SSH checks and offered to query
the details, instead of calling `get_sca_checks`, the tool it had just named. 0/3 runs.

A question that explicitly asks for specific items ("the three failed checks that matter most", "what can
you show my auditor") must not terminate on a summary while rounds remain. This is an orchestration
decision — does the turn have budget, and is there an obvious next call implied by the tool it just ran —
not a matter of prompt phrasing.

## 3. Enumeration completeness — the digest bucket budget is the binding constraint

**Measured:** the SSH-hardening answer named **2 of 10** failed SSH checks and missed
`28638 Ensure sshd PermitRootLogin is disabled` — the one an auditor actually asks about.

The constraint is **not** `MAX_AGG_SIZE` (100). It is `BREAKDOWN_BUCKET_CAP = 5` in `digest.ts`: even a
100-bucket aggregation is trimmed to five before the model reads it.

Direction, with the numbers:

| Bound | Today | Proposed | Why |
| --- | --- | --- | --- |
| `MAX_SAMPLES` (rows the model reads) | 5 | **unchanged** | this is the "don't send everything" guarantee; completeness must not come from rows |
| `BREAKDOWN_BUCKET_CAP` (buckets in the digest) | 5 | **up to ~50 when the aggregation IS the answer**, 5 when it is a side-disclosure | a bucket is ~60 chars, so 50 ≈ 3,000 chars — half of `DIGEST_CHAR_CAP` (6,000), which `capDigest` already enforces by popping buckets, so it degrades instead of bursting |
| `MAX_AGG_SIZE` (buckets requested) | 100 | **unchanged** | 100 is enough once the aggregation is SCOPED (below) |
| `MAX_LOOKBACK_MS` (span) | 90 days | **unchanged** — add disclosure only | cross-quarter comparison is explicitly out of scope; what is in scope is saying the window was capped instead of answering thin |
| `track_total_hits` | already `true` | unchanged | #8909 removed the 10,000 clamp, so counts are already EXACT at any volume — this issue is not about document volume |

## 4. The aggregations, concretely

- **Enumeration ("which ones"):** a `terms` aggregation on the answering field, **scoped with an
  `include` pattern** when the question names a subject — `ssh` → the ~10 matching checks rather than
  competing with 102 for bucket slots. Precedent shipped in #8920: `get_mitre_findings`' `technique_ids`
  aggregation is scoped exactly this way, and the guardrails permit it (`checkAggs` validates `field` and
  `size` only, never `include`). Reuse the instrument, do not invent one.
  Note 102 failed checks exceeds `MAX_AGG_SIZE`, which is precisely why scoping beats raising: a filtered
  aggregation answers the question asked, an unfiltered one cannot enumerate the population anyway.
- **Counts:** `value_count` / `filter` `doc_count` rather than `hits.total`. Both are exact and unbounded.
- **Distinct counts:** `cardinality` with the `precision_threshold` #8920 already raises to the engine
  maximum, and the `Digest.metrics[].approximate` flag it added, so an estimate is never presented as a
  count.
- **Trend (only if a wider window is ever allowed):** `date_histogram` with a calendar interval — 6-12
  buckets, ~250 bytes.

## 5. Acceptance criteria — the live proofs this issue must pass

Each replaces a measured v14 failure. Re-run on the deployed build, N>=3, with the ground truth already
recorded in `qa/proofs-v14-questions.json`:

- [ ] "How badly are we failing CIS, in plain numbers?" → **102 failed / 95 passed / 10 N-A**, no
      capability denial. (v14: 0/3, no tool called.)
- [ ] "My auditor wants proof of SSH hardening" → names the failed SSH checks **including
      `PermitRootLogin`**. (v14: 2 of 10.)
- [ ] "Give me the three failed hardening checks that matter most" → chains `get_sca_checks` in the same
      turn rather than asking permission. (v14: non-answer.)
- [ ] "What Windows patches are missing?" → uses `host.os.name`; **4 documents on win-dc-01**, OS
      `Microsoft Windows Server 2019`. (v14: honest but unanswered.)
- [ ] "What is listening that shouldn't be?" → **14 listening sockets** with the state split. (v14: 2/3.)
- [ ] A question whose window is capped states that the window was capped.
- [ ] Digest payload stays within `DIGEST_CHAR_CAP`; rows shown to the model stay at `MAX_SAMPLES`.
- [ ] No new unmarked approximation or truncation: every raised bound ships with its disclosure, and a
      registry-wide coverage test pins it.

## 6. Recorded, not filed separately

- **Model silence / dead fallback copy:** 1 occurrence in 13 runs on a zero-row two-tool turn, and
      **0/10 on a dedicated re-run**. Intermittent, not introduced by #8920 (nothing was rejected, and it
      was not the final round). Recorded here rather than dramatized as its own bug.
- **`get_top_agents` uses `clampLimit`** where `catalog/common.ts` mandates `clampAggLimit` /
      `aggLimitProperty`; harmless only because the literal `100` happens to equal `MAX_AGG_SIZE`.
- **`process.pid`, `process.parent.pid`, `host.architecture`, `host.os.full`** have no
      `FIELD_POLICY_DEFAULTS` entry while `get_agent_inventory` sets `failClosedFieldPolicy`, so they
      reach the provider as `VAL_n` — over-redaction, the #8917 class, in derived sample columns.

## Evidence

Live proof artifacts and the full split: `deliverables/20-8920-8921-handoff.md` (v14 proof section),
question set with ground truth in `qa/proofs-v14-questions.json`. Data caveat unchanged: every
vulnerability document in the lab is synthetic, and 550 of 574 belong to one bulk-seeded package, so
vulnerability counts characterize the product on seeded data.

yenienserrano:
## Description

Closes #8935

## Proposed Changes

The AI Assistant's answers were being silently thinned in several ways even
when the underlying tool call succeeded:

- **Digest bucket budget**: aggregation breakdowns were capped at a small,
  fixed bucket count regardless of how much room the answer actually had,
  truncating enumeration answers without saying so. The digest now carries
  buckets up to a character budget instead, and discloses every trim.
- **SCA scoped enumeration**: "which SSH checks failed" competed for sample
  rows against every other failed check instead of getting a dedicated,
  population-true enumeration of just the matching check names. `get_sca_checks`
  now attaches a scoped `matching_*` terms aggregation on `check.name` whenever
  `result` and/or `search` narrows the question to a named subject.
- **Tool chaining**: a deferred single-tool offer ended the model's turn
  instead of being followed automatically. The offer is now chained in code,
  so the model doesn't need to remember to ask again.
- **Bound disclosure**: an over-wide `@timestamp` lookback window is now
  clamped and disclosed as a capped-but-successful call, instead of being
  silently narrowed or rejected outright.

While reconciling this branch with the already-merged #8920 fix (PR #8937,
squash-merged to `5.0.0` after this branch's own copy of that work was
written), the full test run surfaced three merge-resolution regressions,
fixed here:

- Two `privacy.test.ts` assertions still used `extractAggFields`'s old
  string-returning shape instead of the `AggFieldSpec` object shape #8937
  introduced.
- A `get-sca-checks.test.ts` assertion asserted the pre-#8935 aggs shape for
  a `result`+`search` combination that the new scoped-enumeration feature
  deliberately changes (superseded by the dedicated tests for that
  combination).
- `executeIndexerRequest` (executor.ts) stopped forwarding `assumptionNote`
  to `buildDigest`, silently dropping issue #8913's resolved-parameter note
  for every Indexer-backed tool call (its `executeManagerRequest` sibling was
  unaffected). Added a regression test covering this end-to-end.

### Results and Evidence

- `yarn test:jest --runInBand` (plugins/wazuh-ai-assistant, inside the
  `osd-dev` container): 90 suites / 1300 tests passing.
- `eslint` and `prettier --check` clean on all changed files.

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (server tools: digest, guardrails, router,
  executor, privacy, get-sca-checks/get-sca-results, openai-compatible
  provider, chat route; public settings page).

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- New coverage tests for the bucket-budget, scoped-enumeration, tool-chaining,
  and lookback-disclosure behavior (colocated `*.test.ts` files under
  `server/tools/` and `server/routes/`).
- A new `executor.test.ts` regression test proving a `resolveParams`-minted
  assumption note reaches the digest for an Indexer-backed tool call.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


Ripdiegozz (approved):
🟢 Tests

### 🟢 Truncation disclosure

**Prompt:** "List every distinct CVE affecting my fleet"

### 🟢 SCA scoped enumeration

**Prompt:** "Which SSH-related SCA checks failed?"

### 🟢 Bucket budget under real cardinality

Prompt: "Break down my findings by MITRE ATT&CK technique"

### 🟢 Tool chaining

**Prompt:** "What's my overall SCA compliance status?"

### 🟢 Bound (time-window) disclosure

**Prompt:** "Show me all critical findings since January 2024."

https://github.com/user-attachments/assets/bb960012-3e05-4a8e-adb0-984f645c5e5f

Issue #8851, opened 2026-08-05
Implement agent configuration index


asteriscos:
## Description

The **Agent > Configuration** views (Global configuration, Policy monitoring, Log collection, Integrity monitoring, Active response, Client/buffer, etc.) currently fetch an agent's effective module configuration on demand from the Wazuh Server API, via `GET /agents/{agent_id}/config/{component}/{configuration}`.

As part of [wazuh/wazuh#38023](https://github.com/wazuh/wazuh/issues/38023), the HTTPS communication protocol introduces a manager-side `/config` endpoint: agents now proactively push their effective configuration to the manager, which persists the latest report per agent instead of the manager querying the agent directly (the legacy `getconfig` model). This persisted data lives in the new `wazuh-agent-config` index.

We need to change the data source for the dashboard's Agent > Configuration views so they read the agent's configuration from `wazuh-agent-config` (directly, or through whatever Server API/indexer access path is agreed on) instead of issuing the current per-module `GET /agents/{agent_id}/config/{component}/{configuration}` requests. This keeps the dashboard aligned with the new persisted-configuration model and avoids relying on the legacy on-demand query path once it is deprecated/removed on the manager side.

### Code references

All current calls go through a single helper that builds the legacy endpoint path and proxies it through the generic Wazuh API request mechanism:

- `getCurrentConfig` (endpoint builder + fetch): https://github.com/wazuh/wazuh-dashboard-plugins/blob/main/plugins/main/public/controllers/management/components/management/configuration/utils/wz-fetch.js#L75-L160
- Generic API proxy used by the fetch above: https://github.com/wazuh/wazuh-dashboard-plugins/blob/main/plugins/main/public/react-services/wz-request.ts#L384-L425
- Backend generic proxy route/controller (no agent-config-specific route exists today): https://github.com/wazuh/wazuh-dashboard-plugins/blob/main/plugins/main/server/routes/wazuh-api.ts#L66-L80 and https://github.com/wazuh/wazuh-dashboard-plugins/blob/main/plugins/main/server/controllers/wazuh-api.ts#L658

Views/components that trigger the fetch (per module, agent context):

- `plugins/main/public/controllers/management/components/management/configuration/configuration-switch.js` (module switch, agent-only views around L261-L349)
- `plugins/main/public/controllers/management/components/management/configuration/integrity-monitoring/integrity-monitoring.js` (syscheck)
[...]

PR #8965, opened 2026-08-13
Agent configuration improvements

### LINKED ISSUE #8962: Agent configuration improvements
## Description

We need to add the `modified_at` property to the main agent configuration view. We must also improve the sections that don't handle cases properly when a particular configuration is not available. Moreover, in case the agent has no document available in the `wazuh-agent-config` index, we must display a message suggesting that the user should check if the agent configuration sync is enabled in the agent **_ossec.conf_** file.

### Affected view

- `Endpoints Summary` → agent detail → **Configuration** tab and its sub-sections:
  `https://<host>/app/endpoints-summary#/agents?tab=configuration&agent=<agent-id>`

### Code references

- [`agent-config-service.ts`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/main/plugins/main/public/controllers/management/components/management/configuration/utils/agent-config-service.ts)

## Tasks

- [ ] Show the `modified_at` property on the main agent configuration view
- [ ] Audit each configuration section/sub-section and handle the case where a given configuration is not available (empty/undefined), instead of rendering incorrectly or breaking
- [ ] When the agent has no document in `wazuh-agent-config` index, show a message suggesting the user check whether agent configuration sync is enabled in the agent's `ossec.conf`
- [ ] Add/update unit tests covering the "no document" and "partial/unavailable configuration" cases
- [ ] Add a `CHANGELOG.md` entry linking to this issue

rodrigofez:
## Description

Agent > Configuration reads the agent's reported configuration from the
`wazuh-agent-config` index ([#8851](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8851)),
but the views were still written against the Server API, which answered for every section that was
asked for. An index document only carries the modules the agent actually reported, so a module that
did not report is `undefined` rather than an empty section, and there is no document at all until
the agent is configured to report. Neither case was handled:

- Sections gated every branch on the module key being truthy, including the "not present" branch, so
  an unreported module rendered **nothing at all** — no settings and no empty state.
- The Integrity monitoring badge compared `syscheck?.disabled === 'no'`, which is `false` for a
  module that never reported, so it claimed **DISABLED** for a module that had made no such claim.
- An agent with no document listed every module anyway, each one opening on an empty section, which
  reads as "this agent configures nothing".
- Nothing told the user how old the configuration on screen was, even though the document records it
  in `state.modified_at`.

Closes #8962

## Proposed Changes

**Read the report once, at the top of the visit.** `configuration-switch.js` reads the agent's
document on mount and on agent change, instead of each section reading it as it opens. Whether a
document exists decides what the whole page shows, so the decision has to precede the module list.
The service already caches the read for the rest of the visit, so the sections opened afterwards
cost no further request.

**Say when the agent reported.** A badge on the configuration overview shows
`state.modified_at` as "Reported 5 minutes ago", with a tooltip carrying the absolute timestamp
(through `formatUIDate`, so it honours the user's date format and timezone). The wording is
deliberately "reported" rather than "updated": this is when the agent last *sent* its configuration,
not when the configuration last changed.

**Prompt when the agent has never reported.** An `EuiEmptyPrompt` replaces the page — same shape as
"Agent is not active" and "Agent has never connected", including their **Select agent** action, so
there is a way out of a page that has nothing to show for this agent. The copy states only what this
view observes and what the data model requires, and defers the how to the documentation: naming the
setting, its default, its file or a restart requirement would be this plugin asserting agent-side
facts that go stale silently.

**Handle the module that did not report.** Every section now distinguishes *reported as absent*
from *did not report*:

| View | Before | After |
| --- | --- | --- |
| Policy monitoring → General | blank | "not present" empty state |
| Policy monitoring → Ignored | blank | "not present"; also renders when only `ignore_sregex` is set, and accepts a single entry reported as a bare value |
| Integrity monitoring | blank | "not present" |
| Active response | blank | "not present", with the "Server not ready yet" case kept ahead of it |
| Log collection | a lone **Sockets** tab claiming no socket is monitored | "not present" |

**Badges stop guessing.** `reportedEnabled(value, enabledValue)` in `utils.js` returns `undefined`
when the setting was not reported, which `WzBadge` already renders as no badge at all. All five
`badgeEnabled()` implementations route through it. Four of them were already correct by accident —
their `&&` chains short-circuit to `undefined` — which is exactly how Integrity monitoring
regressed when it was rewritten with optional chaining. The intent is now stated rather than
emergent.

A read that **fails** (permissions, missing index pattern) is deliberately not treated as "never
reported": the module list stays reachable so each section reports the failure with the detail it
has.

### Results and Evidence

https://github.com/user-attachments/assets/c66e6d02-cdf5-4cf7-82d6-ee33e1ec346f

### Artifacts Affected

- `wazuh` plugin (`plugins/main`): Agent > Configuration overview and its sections
  (Policy monitoring, Integrity monitoring, Active response, Log collection, Inventory,
  Vulnerabilities, Registration service).

No server-side, packaging or index-template change.

### Configuration Changes

None.

### Documentation Updates

None in this repository.

The 5.0 documentation has no page describing agent configuration reporting: on the `5.0.0` branch of
`wazuh/wazuh-documentation`, `source/user-manual/reference/ossec-conf/` still ships `client.rst`
(the pre-HTTPS `<client>` block) with no `config_report`, and no `agent.rst`. The prompt therefore
links the `ossec.conf` reference index — the same URL and label the help popover on this page
already uses. Worth a documentation issue so the prompt can point at the real page later.

### Tests Introduced

Colocated, 20 cases added across 5 files (all `plugins/main` configuration tests pass: 15 suites,
99 tests):

- `configuration-switch.test.tsx` — the prompt (with its Select agent action) replaces the page when
  the agent has no document; the report is read once and reused; the page survives a read that
  fails.
- `configuration-overview.test.tsx` — the badge says how long ago the agent reported; modules are
  listed with and without a report; nothing about reporting for the manager.
- `policy-monitoring-empty-states.test.tsx` — General and Ignored: "not present" when the module did
  not report and when nothing is ignored; ignored paths listed, including a single bare value and an
  `ignore_sregex`-only report.
- `integrity-monitoring.test.tsx` — "not present" for a missing module or missing `syscheck`; tabs
  when reported; and the badge receiving `true` / `false` / `undefined`.
- `active-response-agent.test.tsx` — "not present" cases, settings when reported, and "Server not
  ready yet" taking precedence.
- `utils.test.ts` — `reportedEnabled` for both conventions and for an unreported setting.
- `log-collection.test.tsx` — a case for the module that did not report.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Correct labels applied


asteriscos (approved):
CR: 🟢 
Test: 🟢 

https://github.com/user-attachments/assets/9b98d6e6-5e2d-41d1-81a5-a967740df7cc

Issue #8974, opened 2026-08-14
AI Assistant: usernames inside rule titles reach the provider in privacy mode, and the field policy options are unclear


MiguelCasaresRobles:
## Description

Privacy mode was tested by recording the real HTTP requests the plugin sends to the AI provider (a local capture endpoint registered as a provider). The UI cannot be used for this test: the server converts pseudonyms back to real values before the browser receives them, so a leak and correct behaviour look identical on screen.

Two problems were found. One is fixed; one needs a product decision.

### 1. Usernames inside rule titles reach the provider (not fixed)

With privacy mode ON, this string was sent to the provider unchanged:

```
"Successful user authentication - vagrant"
```

`vagrant` is a username. A second username, `auditbot`, leaked the same way in another title.

Why the existing protection does not catch it:
- `wazuh.rule.title` is `allow` in the field policy, on purpose: titles are mostly fixed text from the Wazuh ruleset and the model needs them to describe a finding.
- The value scan only detects IP and FQDN shapes. A bare username is neither.
- The known-entity scan only replaces values that were already pseudonymized from some other field. These findings carry no `source.user.name`, so no pseudonym is ever created for `vagrant`.

IPs are handled correctly, including inside title text (`"Failed authentication attempt - auditbot from IP_1"`), so the scan itself works. Usernames are the gap.

This needs a decision, because the obvious fix has a cost: anonymizing rule titles would replace almost every distinct finding label with an opaque token and make answers much less useful.

### 2. The "Allow (scanned)" option is not understandable (not fixed)

The field policy editor offers four actions: Allow, Allow (scanned), Anonymize, Never send.

"Allow (scanned)" describes an internal mechanism, not an outcome. An administrator cannot tell from the label what is checked or why they would pick it over "Allow". Only one field in the shipped policy uses it.

Suggested change: reduce the administrator options to three intents.

| Option | Meaning |
| --- | --- |
| Send | The provider receives the real value, always scanned first |
| Pseudonymize | The provider receives `HOST_1`, `IP_2`, ... |
| Never send | The provider receives nothing for this field |

Unscanned `allow` stays available internally for the shipped defaults, which are reviewed field by field.

Note: scanning is best-effort. It does not catch case 1 above. The label should not suggest a guarantee.

## Already fixed

Two related defects found in the same test are fixed on branch `fix/ISSUE_A-ai-assistant-catalog-defects`:

[...]

PR #8977, opened 2026-08-14
Fix AI Assistant catalog and answer defects found in QA

### LINKED ISSUE #8973: AI Assistant: catalog and answer defects found in QA (empty SCA table, unsupported offers, incomplete severity breakdown)
## Description

A QA run of the AI Assistant on 2026-08-14 found several defects in answers and result tables. Each one was reproduced on a live 5.0 stack and is listed below with what the user sees.

**1. SCA checks table renders empty.**
Ask "Which SSH-related CIS checks are failing?". The answer text is correct, but the table shows 10 rows where every cell is a dash and the row expander shows `{}`.
Cause: with a `search` fragment the query uses `post_filter`, so there are 0 hits by design and the rows come from the aggregation. The table then projects hit-document columns (`check.id`, `check.name`, ...) onto bucket rows, so no column resolves.

**2. SCA "Reason" column is always empty.**
The column reads `check.reason`. That field exists in the index mapping but is empty in every document. The real text is in `check.rationale`.

**3. The assistant offers actions it cannot perform.**
Ask about vulnerabilities. The answer offers: "Want me to check whether an updated lxd package is available?". No tool can do this. It also recommends `apt` for a package installed as a `snap`.
Cause: the vulnerability tools do not request `vulnerability.scanner.condition` (which holds the fix version, e.g. "Package less than 5.21.4") or `package.type`.

**4. The assistant promises a next step it cannot take.**
Ask "Give me a security summary, then break it down over time". The answer ends with "Let me pull the same window broken down over time." and the turn stops. This happens on the final tool round, where no further tool call can run.

**5. "Findings by severity" cannot be answered completely.**
Ask "How many findings do we have by severity?". The assistant makes one tool call per severity, uses the whole round budget on three of the five levels, and reports that low and informational were not requested.
Cause: severity (`wazuh.rule.level`) is not a breakdown dimension, so no single call can return the distribution.

**6. MITRE table shows the wrong technique names.**
Ask to list MITRE techniques. Rows T1190 and T1068 show the same pair of names, and a row can list seven names for one ID.
Cause: technique ID and name are parallel arrays on the document. The sampled document is merged whole instead of picking the element matching the bucket key.

**7. Text from different tool rounds is joined without a separator.**
The assistant writes text, calls a tool, then writes the answer. Both land in the same message with no break, producing output like "...for it.The most frequent finding..." and a single message that states two different counts.

## Expected behaviour

- Tables show the data they have, never a grid of dashes.
- The assistant only offers follow-up actions it can perform.
- "By severity" is answered in one call, with all five levels.
- Round text is separated by a blank line.

## Fix

All seven are fixed on branch `fix/ISSUE_A-ai-assistant-catalog-defects`, with unit tests and a before/after check on a live stack.

MiguelCasaresRobles:
## Description

Closes #8973.

Fixes seven defects found in a QA run of the AI Assistant on 2026-08-14, and completes the tool-chaining work issue #8935 asked for but which was not part of the merged fix.

Each defect was reproduced on a live 5.0 stack, fixed, and re-checked on the same stack. The issue lists them with what the user sees; the commits carry the measurement that found each one.

**Defects fixed**

1. SCA checks table rendered every cell as a dash on a topic search (bucket rows projected onto hit-document columns).
2. SCA "Reason" column read `check.reason`, which is empty in every document; it now reads `check.rationale`.
3. The assistant offered remediation actions no tool can perform, and recommended `apt` for a `snap` package. The vulnerability tools now request `vulnerability.scanner.condition` and `package.type`.
4. On the final tool round the assistant promised a follow-up it could not run. The final-round instruction now forbids announcing further work.
5. "Findings by severity" cost one tool call per level and could not reach all five. `wazuh.rule.level` is now a breakdown dimension, so one call returns the distribution.
6. MITRE rows showed the wrong technique names on documents tagged with several techniques. The sampled document is now aligned to the bucket key.
7. Text from different tool rounds was joined without a separator, producing fused sentences in one message.

**Also included** (these were on branches with no PR):

- `enhancement/8921-followup-table-fixes` — get_top_rules severity gap, ports fixture gap (#8921)
- `fix/8921-top-agents-cardinality-guard` — sampled agent-name spread disclosure (#8921)

Those two branches can be deleted. `enhancement/8878-paraphrase-routing-gate` is deliberately NOT
included: it is a test corpus and release gate, not a fix, so it stays on its own branch for a
separate PR. This PR touches only `plugins/wazuh-ai-assistant/server/**` and the CHANGELOG.

**Privacy fixes** found by recording the real requests sent to the provider (see #8974):

- The tool assumption note sent the resolved agent hostname unpseudonymized. Confirmed fixed on the wire: it now reads `HOST_1`.
- A field set to "Never send" removed its values but still sent the field name in the schema hint.

## Results and Evidence

Before and after on the same questions, live stack:

| Question | Before | After |
| --- | --- | --- |
| "Which SSH-related CIS checks are failing?" | 10 rows, every cell a dash | Real check names |
| "How many findings do we have by severity?" | 3 tool calls, 3 of 5 levels | 1 call, all 5 levels |
| "What vulnerabilities does this box have?" | Offered an update check it cannot do; suggested `apt` | Names the fix version and the `snap` channel |

## How to Test

1. Deploy this branch and configure any tool-capable provider.
2. Ask each question in the table above.
3. For the privacy fixes, run a local OpenAI-compatible endpoint that logs request bodies, register it as a provider, enable privacy mode, ask "What software does this box have installed?", and check the logged request: the assumption note must contain `HOST_1`, not the real hostname.

Unit tests: `cd plugins/wazuh-ai-assistant && yarn test:jest --runInBand` — 101 suites, 1531 tests.

## Check List

- [x] Unit tests pass (101 suites / 1531 tests)
- [x] Typecheck introduces no new errors (38; the only file whose count differs from the base is
      `get-agent-inventory.ts`, which this branch fixes)
- [x] Scope: only `plugins/wazuh-ai-assistant/server/**` plus the CHANGELOG - no docs, no test
      tooling, no other plugin
- [x] CHANGELOG entry added, linked to the issue
- [ ] Screenshot/video — server-side changes; evidence is the before/after table above

---

### Rebase note (2026-08-21)

Rebased onto current `5.0.0` (`7e935b0f9`, past #9000/#9001/#8996/#8999). The digest-fallback-synthesis fixes in `chat.ts` now sit alongside the merged cost-unit tool-round budget instead of the old fixed-round-count model they were originally written against. Also folds in four follow-up fixes found in later QA that depend on this PR's code: multi-tool-call no-text fallback synthesis, SCA check-id-direct resolution, its registry test, and the digest-column half of the single-digest fallback naming fix.

Gate on the rebased branch: **117/117 test suites, 1967/1967 tests, lint clean**.


rodrigofez (commented):
🟢 Topic search returns real check names
image

🟢 Reason column populated from check.rationale
image

🟢 The two remediation fields reach the model
image

🟢 MITRE technique names track the bucket key
image


🟢 Severity column on get_top_rules
image

🟢 "Never send" now drops the field name too

image

rodrigofez (changes requested):
we're just missing applying prettier

rodrigofez (approved):

PR #8981, opened 2026-08-17
Document supported providers and models

### LINKED ISSUE #8948: Document the AI Assistant supported providers and models
## Description

The AI Assistant asks the administrator for two things before it can answer anything: a
**provider type** and a **model name**. Both are free-text or near-free-text choices. Today
nothing tells the administrator which combinations actually work.

A wrong choice does not look like a wrong choice. It looks like a broken product. A model that
cannot call tools does not fail visibly — it invents an answer with fake data. A model that
rejects one of the parameters we send ends the turn with no answer at all. In both cases the
administrator only sees a failure (or, worse, a confident wrong answer) in the chat window, with
no way to know if the cause was the provider or the model they typed.

We ran a live test pass on Wazuh 5.0.0 (build v23, 2026-08-11/12) across every provider type and
a wide set of models. This issue asks for documentation that states, in plain language, **which
providers and models we verified as working, which we verified as not working (and why), and
which are expected to work but are still unverified.**

This issue is for the documentation only. It does not include the documentation text, and it does
not ask for any code change. It describes what the documentation must cover.

### Code references

The two supported provider types:

https://github.com/wazuh/wazuh-dashboard-plugins/blob/cacdf22b4/plugins/wazuh-ai-assistant/common/constants.ts#L74

Which adapter each provider type uses:

https://github.com/wazuh/wazuh-dashboard-plugins/blob/cacdf22b4/plugins/wazuh-ai-assistant/server/providers/registry.ts

The two adapters. They define what we send on the wire for each provider type:

https://github.com/wazuh/wazuh-dashboard-plugins/blob/cacdf22b4/plugins/wazuh-ai-assistant/server/providers/anthropic.ts
https://github.com/wazuh/wazuh-dashboard-plugins/blob/cacdf22b4/plugins/wazuh-ai-assistant/server/providers/openai-compatible.ts

The output limit we always send to Anthropic providers:

https://github.com/wazuh/wazuh-dashboard-plugins/blob/cacdf22b4/plugins/wazuh-ai-assistant/common/constants.ts#L82

The in-product guidance the documentation must stay consistent with (base URL examples, key
format hints, and model links shown in the provider form):

https://github.com/wazuh/wazuh-dashboard-plugins/blob/cacdf22b4/plugins/wazuh-ai-assistant/public/components/settings/provider-form-flyout.tsx

## Tasks

- [x] Write a "Supported providers" page. State that we support exactly two provider types:
      `openai_compatible` and `anthropic`. For each one, write one short paragraph that says what
      it is, which services it covers, and what to enter in the base URL field.
- [x] State clearly that **the provider type must match the service, and the base URL alone does
      not select it**. Anthropic's own API (`api.anthropic.com`) only works with the `anthropic`
      provider type. A Claude model behind an `openai_compatible` provider does not work at all —
      this is a hard rule, not a degraded mode. The documentation must say so directly, so
      administrators do not try to work around it.
- [x] State that a successful **Test** on a provider does not prove it can answer a question. We
      measured this on three separate provider/model combinations during the test pass: Test
      passed while the first real chat turn failed. Tell the administrator to confirm with one
      real question, not with the Test button alone.
- [x] Add a "Verified supported" table with one row per model, confirmed live on build v23:
      - Provider type `anthropic`, base `api.anthropic.com`: all 8 current Claude models.
        Recommend `claude-haiku-4-5` for the fastest responses and `claude-opus-4-8` as the
        balanced default.
      - Provider type `openai_compatible`: `openai/gpt-oss-120b` (via a Bedrock gateway
        endpoint), `mistral-large-3` (675b), `qwen3-32b`, `qwen3-coder-480b`, `deepseek-v3.2`.
      - Gemini (`gemini-flash-latest`, `gemini-3-flash-preview`) via `openai_compatible`: works.
        Note two gotchas: new Google accounts must use the alias model id
        (`gemini-flash-latest`), not a dated id; and the free tier's 20-requests-per-day quota is
        too low for real use — a paid key is needed.
- [x] Add a "Verified NOT supported" table with the reason and the visible symptom for each:
      - Claude models through any `openai_compatible` gateway: impossible. The only correct fix is
        to create an `anthropic`-type provider pointed at `api.anthropic.com`.
      - Groq: the assistant's tool schemas are rejected with HTTP 413 on typical account tiers.
        State this as a Groq-side limit, not something the administrator can configure around.
      - Models without reliable tool calling, confirmed with `gemma-3-27b` and `qwen3-235b`:
        these do **not** fail with an error. They answer fluently but **fabricate** security data
        — invented alerts, invented agents, invented numbers — because they cannot call the tools
        that would fetch real data. This is the most important warning on the whole page: readers
        must understand tool calling is a hard requirement, and that this failure looks like a
        normal, confident answer, not a broken feature.
      - OpenRouter free tier: too slow for real use (2-3 minutes per answer). OpenRouter's paid
        models need a funded account with credits.
- [x] Add an "Expected to work, not yet verified" table for combinations that share a verified
      wire protocol but that we could not test (no key or endpoint available at test time):
      - OpenAI's own API (`gpt-5.x`, `gpt-4o`, and similar) — call this out as the top priority to
        verify next, since it is one of the providers administrators are most likely to try.
      - Llama 3.3/4 through any `openai_compatible` host.
      - Self-hosted Ollama, base URL pattern `http://<host>:11434/v1`.
      Make clear in the text that "expected to work" is not the same claim as "verified," and
      label these differently in the tables (for example, a distinct column or badge).
- [x] State the hard requirement that a model must support tool calling. Explain that the
      assistant reads Wazuh data through tools, so a model without tool calling cannot answer any
      question about the environment — and, per the finding above, will not say so; it will
      fabricate an answer instead.
- [x] Document the output limit sent to Anthropic providers (4096 tokens) and that it is fixed
      and cannot be changed in the UI. Explain the effect on models that reason before answering:
      the limit is shared between the reasoning and the answer, so the answer can be cut short.
- [x] Document that some current Anthropic models reject the `temperature` parameter, that the
      plugin detects this and retries the request without it, and that the administrator does not
      need to do anything about it.
- [x] Document the per-vendor model-id gotchas found during testing, so administrators do not
      lose time on a wrong id: Gemini's alias vs. dated model ids, the Bedrock gateway's model
      namespace prefix (for example `openai/gpt-oss-120b`), and OpenRouter's `:free` suffix, which
      means a different (rate-limited) model than the paid id of the same name.
- [x] Document any provider or model that needs account-level settings before it works at all
      (for example, a data retention setting), so the administrator checks this before reporting
      a failure.
- [x] State the support policy for models we have not tested: is an untested model "unsupported"
      or "use at your own risk," and who should the administrator contact to request support for
      a new one.
- [x] Add a short troubleshooting table: symptom in the chat window -> most likely provider or
      model cause -> what to change. Include the fabrication symptom (confident answer, no real
      data) as its own row, since it is silent and easy to miss.
- [x] Agree where this page lives (Wazuh documentation site vs. the plugin README) and link it
      from the provider form, so the administrator can reach it while filling in the form.
- [x] Keep the wording consistent with the in-product hints in the provider form. If the
      documentation and the form disagree about a base URL, a key format, or a model name, the
      form wins and the documentation must be corrected.

Desvelao:
## Description

An administrator configuring the AI Assistant picks a provider type and a model name, both close
to free text, with nothing today telling them which combinations actually work. A wrong choice
does not fail visibly: a model without reliable tool calling doesn't error, it fabricates answers
with fake data, and a passing connection Test does not prove a real chat turn will succeed. This
is documentation only — no code changes.

Closes #8948

## Proposed Changes

- Expanded the **Providers** reference page
  (`docs/ref/modules/ai-assistant/providers.md`) with:
  - A single **Verification status** table (one row per provider/model combination, with a Status
    column) showing which are **verified supported**, which are **verified NOT supported** (with
    the reason and the visible symptom), and which are **expected to work but not yet verified**.
  - The hard rule that the provider type must match the service (a Claude model never works
    through an OpenAI-compatible provider, regardless of base URL).
  - Why a successful connection **Test** does not prove the assistant can answer a question.
  - Why tool calling is a hard requirement, and why a model without it fabricates data instead of
    failing visibly (the most important failure mode documented here).
  - The fixed 4,096-token Anthropic output limit and its effect on reasoning models.
  - The automatic, silent retry for Anthropic models that reject the temperature setting.
  - Per-vendor model-name gotchas (Gemini alias IDs, Bedrock gateway namespacing, OpenRouter's
    `:free` suffix).
  - A support policy for untested providers/models.
  - A symptom → cause → fix troubleshooting table.
- Updated `docs/ref/modules/ai-assistant/README.md` and `configuration.md`, and the plugin's own
  `README.md`, so they no longer describe a third "Wazuh hosted brain" provider type as available
  — it exists in the adapter registry but isn't yet selectable in Settings — and cross-link the
  expanded Providers page.

### Results and Evidence

Documentation-only change. Verified the mdBook docs site (`docs/`) builds cleanly with `mdbook
build` and that every cross-reference/anchor resolves.

### Artifacts Affected

- `docs/ref/modules/ai-assistant/providers.md`
- `docs/ref/modules/ai-assistant/README.md`
- `docs/ref/modules/ai-assistant/configuration.md`
- `plugins/wazuh-ai-assistant/README.md`

### Configuration Changes

None.

### Documentation Updates

This PR is a documentation update in full — see **Proposed Changes** above.

### Tests Introduced

None (documentation only).

## Review Checklist

- [x] Code changes reviewed
- [ ] Tests cover the new functionality
- [ ] ...


Ripdiegozz (dismissed):
🟢 LGTM

docs/ref/modules/ai-assistant/providers.md
```diff
@@ -29,10 +37,102 @@ cause — the model, not the query — and pointing at the Model field's guidanc
...
+| ❌ Verified NOT supported | OpenRouter, free tier                                          | —                                                                                                                     | Too slow for real use — 2-3 minutes per answer.                                                                                                                                                                                                                |
+| ❌ Verified NOT supported | OpenRouter, paid models                                        | —                                                                                                                     | Need a funded account with credits.                                                                                                                                                                                                                            |
```
rodrigofez: I think OpenRouter models not being free should not be stated as a reason for not supporting them, you’d have to pay with either Anthropic or OpenAI as well

Desvelao: You have right, I will change to should work.

Desvelao: Fixed — moved the OpenRouter paid models row to "Expected, unverified" and dropped the payment-related note in 91852f5eb.

rodrigofez (approved):
🟢 CR

Ripdiegozz (approved):

PR #8982, opened 2026-08-17
Home AI assistant card

### LINKED ISSUE #8979: Home AI assistant card
## Tasks

- [x] Remove the AI assistant card 
- [x] Add a link to it in the Quick Access shortcut
- [x] Enable the AI chat icon in the top bar
- [x] In the AI chat sidecar flyout, add a "maximize" icon that navigates to the AI assistant plugin chat page

Image

rodrigofez:
## Description

The AI Assistant was surfaced on the Home page as a full overview card sitting
beside the platform metric widgets. That gave a plain navigation entry point the
same visual weight as the data tiles around it, and it was the only way into the
assistant from Home — the header button that opens the chat sidecar was still
commented out, and the sidecar itself offered no way to move the conversation to
the assistant's own full-width page.

This replaces the card with lighter, better-placed entry points: a shortcut in the
Home Quick access popover, the re-enabled header button, and a maximize control in
the sidecar that hands the conversation off to the app-shell chat page.

It also carries two fixes to the chat's own status callouts, found while testing the
above and **outside the scope of the issue**: the callouts now stick to the top of
the transcript instead of scrolling out of sight in a long conversation, and the
error callout can be dismissed. Happy to split those into their own issue and PR if
preferred.

Closes #8979

## Proposed Changes

All four tasks from the issue, plus the two out-of-scope callout fixes (5 and 6):

**1. Removed the AI Assistant card from the Home overview** (`plugins/main`)

- Deleted `home-overview/components/overview/ai-assistant-cta.tsx` and its
  colocated test, and dropped the `<AiAssistantCta />` flex item from
  `overview-section.tsx`.
- Removed the matching `export * from './ai-assistant-cta'` from the
  `components/overview/index.ts` barrel. This one mattered: `home-overview.tsx`
  imports from that barrel, so a stale re-export kept the deleted module in the
  dependency graph and failed the bundle with `Module not found: Can't resolve
  './ai-assistant-cta'` — the build never produced a new bundle, so the card
  stayed visible in the running dev server.

**2. Added an AI Assistant shortcut to the Quick access popover** (`plugins/main`)

- `quick-access-menu.tsx` — the `EuiPopoverTitle` now holds a
  `justifyContent='spaceBetween'` row: the "Quick access" title on the left, an
  `EuiLink` with the `machineLearningApp` icon and the "AI Assistant" label on the
  right, vertically centred against the title.
- It lives in the title rather than in the group grid below because the assistant
  is not one of the `Applications`/category-driven groups — it is a single
  shortcut, not a content type.
- The link sits inside the popover's existing `RedirectAppLinks` wrapper, so its
  `href` is intercepted into an in-app `core.application.navigateToApp` instead of
  a full page reload, exactly like every group item.
- `getAiAssistantUrl` (`home-overview/utils/navigation.ts`) is reused as-is; only
  its doc comment changed, since it used to point at the deleted card file.

**3. Enabled the AI chat header button** (`plugins/wazuh-ai-assistant`)

- `public/plugin.ts` — dropped the `// TODO: Uncomment this when the AI Assistant
  sidecar is fixed` guard and restored the `registerAssistantHeaderButton(core)`
  call in `start()`, so the top-bar chat icon is registered again.

**4. Added a maximize control to the sidecar flyout** (`plugins/wazuh-ai-assistant`)

- `assistant-chat-panel.tsx` — a new `EuiButtonIcon` with `iconType='expand'` in
  the panel toolbar, between the conversations toggle and the settings gear.
  `expand` is EUI's outward-arrows maximize glyph; `fullScreen` reads as browser
  fullscreen and `popout` implies a new tab or external link, neither of which is
  what happens here.
- Its `openFullPage` handler mirrors the existing `openSettings` exactly: it goes
  through `runGuarded`, so an in-flight generation prompts for confirmation first,
  then closes the sidecar and calls
  `navigateToApp(PLUGIN_ID, { path: '#/' })` — the plugin's default route, i.e.
  the app-shell chat page. Closing the sidecar first is deliberate: leaving it
  docked over the page it just navigated to would render the chat twice.

**5. Made the chat status callouts stick to the top of the transcript** (`plugins/wazuh-ai-assistant`)

Not part of issue #8979 — a separate problem found while testing the above. When a
turn failed, the "Something went wrong" callout rendered at the top of the
transcript's flow and scrolled away with the first screenful, so in a long
conversation the user got an answer that never arrived and no visible reason why.

- `chat-page.scss` — new `.wzStatusCallouts`: `position: sticky; top: 0`, with an
  opaque `--wz-surface` ground so message rows scroll behind it rather than
  through the `EuiSpacer` gaps between callouts.
- `chat-page.tsx` — the callout stack moved out of `.wzContentMeasure` and into
  that band as a **direct child of `.wzTranscriptContent`**. This placement is the
  mechanism, not a detail: a sticky element only travels inside its own parent's
  box, and the ordinary (non-welcome) `.wzContentMeasure` is a plain flow box no
  taller than the callouts themselves — sticking inside it had nowhere to go.
  `.wzTranscriptContent` is the flex column carrying `min-height: 100%` off the
  scroll container, so it spans the whole scrollable height.
- A `hasStatusCallout` flag gates the band, so an empty band never paints its
  ground over the first transcript row.

**6. Made the error callout dismissible** (`plugins/wazuh-ai-assistant`)

- `status-callout.tsx` — new optional `onDismiss` prop rendering a labelled close
  control. It does **not** use `EuiCallOut`'s own `dismissible`/`onDismiss` pair:
  that implementation renders its close icon only when
  `color !== 'warning' && color !== 'danger'` (`call_out.js`), so on the danger
  callout the props would be accepted and silently do nothing. It also hardcodes
  `aria-label="dismissible_icon"`, an untranslated internal string.
- `chat-page.tsx` — only the generic error callout passes `onDismiss`. A failed
  turn is already visible in the transcript and the user can just ask again, so
  once they have read why it failed there is nothing left for it to report. The
  session-expiry and save-failure notices deliberately pass nothing: they describe
  a condition that is still true after being read, and hiding them would leave a
  real problem invisible.
- Dismissal is tracked as the dismissed error **message**, not a boolean. `error`
  is already cleared by every path that starts fresh work, and an effect turns
  that into the dismissal being dropped — so the next failure surfaces even when
  it reports the identical text, with no reset call to keep in sync at five
  existing call sites. Comparing by value also stops a dismissed send error from
  suppressing a different, still-current `providersError` underneath it.

### Results and Evidence

Local validation:

```
Prettier (CI-equivalent 2.8.8, 14 changed files):  PASS
ESLint (changed code files):                       PASS
Jest plugins/wazuh-ai-assistant (full suite):      PASS — 99 suites / 1460 tests
Jest plugins/main (full suite):                    PASS — 211 suites / 1628 tests
```

### Artifacts Affected

- `plugins/main` — Home overview bundle (`wazuh` plugin).
- `plugins/wazuh-ai-assistant` — public bundle (`wazuhAiAssistant` plugin).

No packaging, executable, or dependency changes.

### Configuration Changes

None. No new settings, no default-value changes, no backward-compatibility
concerns — the removed card and the added shortcuts are all navigation surfaces.

### Documentation Updates

None required. No developer-facing doc under `docs/` describes the Home overview
card or the sidecar toolbar.

### Tests Introduced

Colocated unit tests, per repo convention:

- `quick-access-menu.test.tsx` — two tests: the shortcut renders in the popover
  title with the "AI Assistant" label and links to the assistant app id; and it
  stays hidden until the popover is opened. Note these query
  `[data-test-subj=...]` via `document.querySelector` rather than `getByTestId`,
  because this repo does not configure testing-library's `testIdAttribute` and
  OSD uses `data-test-subj`.
- `assistant-chat-panel.test.tsx` — one test: clicking the maximize button routes
  through the interrupt gate (`runGuarded`), closes the panel, and navigates to
  `wazuhAiAssistant` at `#/`.

No new tests cover the sticky band or the dismiss control — verified manually,
evidence above.

One existing test **changed** rather than being added: `chat-page.test.tsx`'s
"removes the old sticky/gradient mechanism from the stylesheet entirely" asserted
that the whole stylesheet contained no `position: sticky` anywhere. That was a fair
proxy while the composer was the only thing that had ever been sticky, but
`.wzStatusCallouts` is now legitimately sticky, so the file-wide form would fail on
an unrelated rule. It is now scoped to `.wzComposerRow`'s own rule block — which is
what the test actually protects: that the composer is a flow grid row and not a
sticky overlay. A brace-matching `ruleBlock` helper does the scoping.

Removed with their subject: `ai-assistant-cta.test.tsx`, and the now-dead
`getAiAssistantUrl` mock entry in `overview-section.test.tsx`.

## Review Checklist

- [x] Code changes reviewed


asteriscos (approved):
CR: 🟢 
Test: 🟢 

https://github.com/user-attachments/assets/80a7170d-d193-45a0-80ff-731f272617d4

PR #8996, opened 2026-08-19
Refine the AI Assistant UX: type scale, spacing, empty state and chat polish

### LINKED ISSUE #8984: Refine the AI Assistant UX: type scale, spacing, empty state and chat polish
## Description

A design review of the AI Assistant found visual and UX problems in three areas: the chat, the
settings page, and the add-provider form. This issue tracks the iteration that fixes them. The
work is guided by a measured audit of the live UI and by common design rules (spacing grid, type
scale, alignment) from Apple HIG, Material 3 and EUI.

### Add-provider form
- Use one type scale: two text sizes, each from one mechanism. Help text had three different
  sizes before.
- Show the getting-started steps as a numbered list. Remove the caveat sentence about the
  connection test.
- Use one shared example-chip component (monospace, click fills the field). Remove the generic
  model example chips: the combo box, the endpoint-matched suggestions and the docs link
  already cover that job, and the generic ids were only valid for one endpoint.
- Make the form a single 640px column. It was a 960px flyout around a 400px form.
- Make section legends look different from field labels (they were identical).
- Make the two provider-type cards the same size, each with its own description.

### Settings page
- Cap the page at a centered 1200px measure. It stretched to the full viewport before, which
  read as large empty areas at high resolutions.
- Clip endpoint URLs inside their table column. Long URLs painted over the Model column.
- Make section titles rank above their body text.
- Cap descriptions at a readable line length.
- Lay out the Privacy and Conversation-history cards in two columns to use their width.
- Remove the fill from disabled Save buttons (a disabled filled button was the darkest element
  on the page, at 2.2:1 contrast).

### Chat
- Empty state: center the greeting, the example cards and the composer as one group. On the
  first send, the composer moves down to its docked position (based on a frame-by-frame capture
  of Gemini's transition). Reduced motion gets a hard cut. The header sidecar keeps its current
  always-docked behavior.
- Add a "Jump to latest" button when the user scrolls up. It shares the transcript's grid cell,
  so it never changes the layout when it appears, and the scroll-pin check has hysteresis so it
  cannot flicker.
- Do not render tables with zero rows. The assistant's own answer carries the "no results"
  message; a short fallback line appears only when the answer has no text.
- Keep one left edge for answer text, also on turns that carry a table.
- Ask the model for consistent markdown emphasis (bold key figures, inline code for
  identifiers), so answers do not look formatted at random.

### Cross-cutting
- One container radius (12px). Four different radii were shipping.
- Spacing values on the 4/8/16/24 ladder.
- Inline-code chips get a small radius and tighter padding.
- Dead i18n keys removed.

Evidence: before/after screenshots attached below.

yenienserrano:
## Description

Closes #8984

## Proposed Changes

Iteration on the AI Assistant's UX based on a measured design audit against the
spacing grid, type scale and alignment rules from Apple HIG, Material 3 and EUI.

**Add-provider form**
- Collapsed help text to one shared type scale; getting-started steps now render as
  a numbered list.
- Replaced the generic model example chips with one shared example-chip component,
  fed by the endpoint-matched suggestions and the docs link.
- Capped the form to a single 640px column instead of a 960px flyout around a 400px
  form; the two provider-type cards are now equal-sized, each with its own
  description.
- Distinguished section legends from field labels.

**Settings page**
- Capped the page at a centered 1200px measure; laid out the Privacy and
  Conversation-history cards in two columns.
- Clipped endpoint URLs inside their own table column instead of overflowing into
  the Model column.
- Ranked section titles above their body text and capped description line length.
- Dropped the fill on disabled Save buttons (previously the darkest element on the
  page at 2.2:1 contrast).
- Added a tabbed shell, per-provider privacy default overrides, and fixed several
  tab-state/URL-param bugs surfaced along the way.

**Chat**
- Empty state now centers the greeting, example cards and composer as one group;
  on first send the composer docks down (reduced motion gets a hard cut).
- Added a "Jump to latest" control that shares the transcript's grid cell (no
  layout shift) with hysteresis on the scroll-pin check.
- Zero-row result tables no longer render; a short fallback line only appears when
  the answer itself has no text.
- Kept one left edge for answer text on turns that carry a table, and asked the
  model for consistent markdown emphasis.
- Replaced the padlock + tooltip privacy control with a single clickable chip plus
  a discrete info icon, and reworked the provider picker into a text-trigger
  popover with proper focus/accessibility handling.

**Cross-cutting**
- One container radius (12px) and spacing values on the 4/8/16/24 ladder throughout.
- Inline-code chips got a smaller radius and tighter padding; dead i18n keys removed.

**Standards pass**
- Ran `check-standards` over the diff: Prettier/ESLint autofixes, a missing third
  argument to `noTextFallbackMessage` (TS2554), a couple of genuinely-unused
  imports/assignments, two `require-await` test helpers, and a settings-page test
  that located the wrong `EuiSwitch` by role+name (every tab's card stays mounted
  hidden rather than unmounted, so accessible-name lookups collided on this test
  env's mocked `htmlIdGenerator`).

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (public: chat, settings, header components; server:
  routes, providers, prompts, tools)

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- Colocated unit tests updated/added alongside every touched component and route
  (chat page, message list/bubble, provider picker, result table, settings page,
  provider-form flyout, chat routes, providers, prompts, tools).
- Cypress transcript-geometry specs added under
  `plugins/wazuh-ai-assistant/test/cypress`.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


rodrigofez (commented):
- Double horizontal rule shown in per provider override setting
Screenshot 2026-08-19 at 08 04 40

- Callout looks bad with a background
Screenshot 2026-08-19 at 08 06 50

rodrigofez (dismissed):

rodrigofez (approved):
🟢 Test

https://github.com/user-attachments/assets/852713fd-3198-4aea-9869-9557d60ac1fb

PR #8999, opened 2026-08-20
Fix duplicate provider names and stale settings

### LINKED ISSUE #1512: Errors detected in the AI Assistant module
## Description

The QA team has developed a new automated end-to-end (E2E) test that checks the functionality of the AI Assistant module on the dashboard. During one of its runs, we detected the following errors:

### 1. Add provider with an already-taken name

In this test, we create two providers with the same name. Creating a second provider with an already-used name was accepted and saved instead of being rejected. The providers table then contained two rows with the same name, and the chat's Provider dropdown offered the options with the two duplicates having identical visible text with no endpoint, model, status, suffix, or tooltip to distinguish them.

#### Steps to reproduce

1. Click "Add provider" and create a first provider.
2. Click "Add provider" and create a SECOND provider using a name that is already taken by an existing row (reuse the name of any provider currently in the table).
3. Look at the providers table and at the chat's "Provider" dropdown. Two entries with the identical name and no way to tell them apart (different endpoint, a suffix, anything) means a user cannot know which provider a question was answered by. Environment is left as you found it.

#### Actual result

The providers table then contained two rows with the same name, and the chat's Provider dropdown offered the option with two duplicates that have identical visible text. The Delete confirmation dialog is equally ambiguous

#### Expected result

The AI provider form validates input before creating it.

<details><summary>Evidence</summary>

Image

</details> 

---

### 2. Admin privacy policy is not applied to the Chat tab until the page is reloaded

In AI Assistant → Settings → Privacy, an administrator can turn on "Enable privacy mode by default" and turn off "Allow users to override privacy mode from the chat page", which should lock the per-conversation privacy control ON so users cannot disable it. When the administrator saves this policy, the Chat tab does not pick it up in the current session: opening Chat and starting a new conversation still shows the privacy control as "Off" and editable, and the user can freely toggle it Off↔On. The policy is only honored after a full page reload. The staleness also happens in reverse: after the administrator relaxes the policy again, the Chat control stays locked "On" until the page is reloaded. Because privacy mode is what pseudonymizes hostnames, IP addresses, usernames, process command lines, and finding/rule text before they are sent to the configured AI provider, a user can bypass the administrator's lock within a session and send that data unpseudonymized.

#### Steps to reproduce

1. As an administrator, open AI Assistant → Settings → Privacy and note the current state (e.g. "Enable privacy mode by default" = Off, "Allow users to override privacy mode from the chat page" = On).
2. Set "Enable privacy mode by default" = On and "Allow users to override privacy mode from the chat page" = Off, then click "Save privacy settings" (a "Privacy settings saved." toast appears).
3. Without reloading the page, open the "Chat" tab and click "New conversation".
4. Look at the per-conversation privacy control in the composer: it still reads "Off" and is enabled. Click it — it flips to "On" and back to "Off" freely.
5. Reload the page and open Chat again: only now does the control read "On", become disabled/locked, and show the tooltip "Set by administrator".
6. (Reverse case) Restore "Allow users to override" = On and "Enable privacy mode by default" = Off and Save. Without reloading, Chat → "New conversation" still shows the control locked "On"; only after a page reload does it become "Off" and editable again.

#### Actual result

After the administrator saves the privacy policy, the Chat tab keeps the previous privacy state for the rest of the session. With privacy set to On-by-default and override disallowed, a new conversation still shows the privacy control as "Off" and editable, so a user can turn privacy Off despite the administrator having disallowed it (reproduced twice). The new policy is only applied after a full page reload. (The backend policy and the locked, read-only control with the "Set by administrator" attribution are correct — the only problem is that the Chat tab does not refresh the policy without a reload.)

#### Expected result

Saving the privacy policy should take effect immediately in the Chat tab within the same session, with no page reload required: once the administrator enables privacy by default and disallows override, the privacy control in a new conversation should immediately be locked "On" (and, ideally, the current conversation too), so a user cannot disable privacy or send unpseudonymized data to the AI provider.

<details><summary>Evidence</summary>

Image

Image

---

Image

Image

</details>

---

### UX report

The test also evaluates the user experience during runtime, using LLM agents. Here are the improvements it has detected:

[ai_assistant_UX_REPORT.html](https://github.com/user-attachments/files/31254976/ai_assistant_UX_REPORT.html)

yenienserrano:
## Description

Closes https://github.com/wazuh/wazuh-dashboard/issues/1512

Two independent problems with the AI Assistant's provider/privacy settings:

1. Two AI providers could share the same name, making them indistinguishable
   in the chat's provider selector and in every toast the plugin emits.
2. The Chat and Settings views stay mounted side by side behind
   `display: none` (and the header flyout holds its own, separate `ChatPage`
   / `useProviders` instance), so a privacy policy saved in Settings — or a
   provider created, edited, deleted, or made default — never reached an
   already-open chat until a full page reload.

## Proposed Changes

- **Reject duplicate provider names.** `POST /providers` and
  `PUT /providers/{id}` now list existing providers and refuse a name another
  provider already holds (HTTP 409), comparing on `trim()` +
  `toLowerCase()`. `PUT` excludes the provider being updated. The add/edit
  flyout mirrors this client-side against the loaded provider list before the
  round-trip, and surfaces a genuine race's 409 through its existing error
  callout.
- **Harden the uniqueness gate.** The gate now re-lists with the store's real
  `total` instead of one page of 200 (matching the flyout's own
  `fetchAllPages` pre-check), trims the name at write time so storage matches
  what was compared, rejects a name that trims to empty, and is now covered
  by a test that drives the real routes end-to-end and asserts no
  create/update happens on a 409. The residual read-then-write race (no
  unique constraint at the indexer) is accepted and documented.
- **Surface all provider form errors on one click.** The name and
  endpoint-URL checks ran sequentially; both are now evaluated before either
  returns, so a form with two problems doesn't need two submit attempts to
  see both.
- **Apply saved AI settings to an open chat.** Two window events, dispatched
  after a successful save and consumed by every mounted consumer:
  - `assistantSettingsChanged` — every `ChatPage` refetches (and also
    refetches when it becomes the visible tab). When `userCanOverride`
    arrives `false`, the "user already toggled" flag is cleared so the lock
    binds the current conversation, not only the next.
  - `providersChanged` — `useProviders` refreshes, fixing every consumer at
    once, including the header flyout's own instance that no prop callback
    can reach.
- **Make the chat privacy chip tell the truth.** The events above only reach
  the browser that did the saving. Two more triggers close that gap without
  polling: `handleSend` re-reads the policy before building the request body
  (bounded by a 1.5s deadline, fail-soft — never blocks sending, never
  silently flips privacy off), and a `visibilitychange` refetch corrects an
  idle chat when the user returns to the tab. Also: the settings event now
  carries the saved document as `CustomEvent` detail (applied directly,
  closing a read-after-write gap where a re-GET could still see the pre-save
  document), every refetch path is guarded against resolving after unmount,
  and a stale comment claiming tab-switch alone covered other-tab saves was
  corrected.

### Results and Evidence

Covered by the automated test suite below (1641 tests passing), plus a
manual click-through against a running dev instance (`docker/osd-dev`,
`5.0.0` base). Screenshots below; some steps only have text evidence
(accessibility-tree snapshots / network payloads) where noted.

**Duplicate name + malformed URL, one click.** Submitted the Add-provider
form with `Name: test` (already used by an existing provider) and
`Endpoint URL: not-a-valid-url` together. Both inline errors appeared from
the single submit, no second attempt needed:
- Name field: `A provider named "test" already exists. Pick a different name.`
- Endpoint URL field: `Enter a valid URL starting with http:// or https://`

invalid-url-error

**Privacy policy reaches an already-open chat, live.** With a Chat tab
already open showing the composer chip as `Privacy · Off` (clickable
`<button>`), saved `Enable privacy mode by default: On` +
`Allow users to override privacy mode: Off` on the Settings tab. Switching
back to the same (never-reloaded) Chat tab, the chip had flipped to a
locked `Privacy · On`. Confirmed in the DOM the element is no longer
clickable:
```
tagName: "SPAN"
className: "euiBadge euiBadge--hollow euiBadge--iconLeft wzPrivacyChip wzPrivacyChip--on"
```
The header flyout's independent `ChatPage`/`useProviders` instance,
opened fresh afterwards, showed the same locked `Privacy · On` and the
same selected provider — consistent with the tab that was live-updated.

chat-privacy-off-before
chat-privacy-on-live-after-save
header-flyout-consistent

**`providersChanged` reaches every mounted consumer.** With the header
flyout open and untouched, created a provider (`e2e-test-provider`) from
the main Settings tab. Re-opened the main Chat tab's provider picker
(mounted the whole time, never remounted) and it already listed the new
provider alongside the two pre-existing ones:
```
menuitemradio "test Default gemma4:12b" [checked=true]
menuitemradio "nvidia nvidia/nemotron-3.5-lightning:free" [checked=false]
menuitemradio "e2e-test-provider llama3" [checked=false]
```
The header flyout's own picker (still the same open instance) showed the
identical list. Deleted `e2e-test-provider` from Settings and re-opened the
Chat tab's picker again without any reload — the deleted provider was gone
immediately:
```
menuitemradio "test Default gemma4:12b" [checked=true]
menuitemradio "nvidia nvidia/nemotron-3.5-lightning:free" [checked=false]
menuitemradio "e2e-crud-check llama3" [checked=false]
```

provider-created-toast

**Pre-send re-read sends what the server will enforce.** Sent a real
message while privacy was locked on (`privacyDefaultOn: true`,
`userCanOverride: false`). Captured the outgoing request body via the
browser's network log:
```json
POST /api/wazuh_ai_assistant/chat
{"providerId":"...","messages":[{"role":"user","content":"hola, probando pre-send privacy re-read"}],"privacy":{"enabled":true,"map":[]}}
```
`privacy.enabled` matches the just-saved policy at send time, not a stale
in-memory value. (The provider itself failed to respond — no reachable
LLM endpoint in this local dev environment — which is unrelated to this
PR; the request payload is the thing under test.)

real-message-sent

Test providers were deleted and the privacy settings reverted to their
original values after this pass, leaving the dev instance as found.

Not yet run: the existing Cypress suite under
`plugins/wazuh-ai-assistant/test/cypress`.

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (public, server, common)

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- Server: route-wiring test proving the 409 gate is actually wired into
  `POST`/`PUT` (not just unit-tested in isolation), and a uniqueness test
  covering the `total`-aware re-list, trim-at-write, and empty-after-trim
  rejection.
- Public: coverage for cross-view propagation of `assistantSettingsChanged`
  and `providersChanged` (chat page and settings page), the pre-send policy
  re-read, the visibility-change refetch, and the provider form's combined
  name/URL validation.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,7 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
+| [#1512](https://github.com/wazuh/wazuh-dashboard/issues/1512)         | Fixed AI assistant accepting duplicate provider names, and the admin privacy policy and provider list not reaching an open chat until a page reload                                           |
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,7 +138,6 @@
 | [#8708](https://github.com/wazuh/wazuh-dashboard-plugins/pull/8708)   | Fixed Home KPI's visualization persistent filters                                                                                                                                             |
 | [#8766](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8766) | Fixed Server Management Settings crashing or showing a blank page for users without permission to read the manager configuration                                                              |
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                |
-| [#1512](https://github.com/wazuh/wazuh-dashboard/issues/1512)         | Fixed AI assistant accepting duplicate provider names, and the admin privacy policy and provider list not reaching an open chat until a page reload                                           |
 
 ### Removed
 
--- a/plugins/wazuh-ai-assistant/public/components/chat/chat-page.test.tsx
+++ b/plugins/wazuh-ai-assistant/public/components/chat/chat-page.test.tsx
@@ -2975,11 +2975,28 @@ describe('ChatPage — admin privacy policy applies without a reload', () => {
    * long per render, which tipped these over. Explicit headroom instead of a global bump, so a
    * genuinely hung test still fails rather than stalling the gate. */
   const PRIVACY_POLICY_TEST_TIMEOUT_MS = 30_000;
-  // Scoped through jest.setTimeout rather than a per-test third argument, which Prettier expands
-  // into a far noisier shape; the afterEach restores Jest's own default so no later file-level
-  // test inherits the longer budget.
-  beforeEach(() => jest.setTimeout(PRIVACY_POLICY_TEST_TIMEOUT_MS));
-  afterEach(() => jest.setTimeout(5_000));
+  /**
+   * Scoped through jest.setTimeout rather than a per-test third argument, which Prettier expands
+   * into a far noisier shape.
+   *
+   * Jest exposes no public getter for the configured test timeout, so the previous budget is read
+   * off the jasmine global (present under the jasmine2 runner, which is what the platform's Jest
+   * config uses) and restored afterwards. Nothing is hardcoded: if that global is ever absent the
+   * raised budget simply stays in effect, which is harmless because this is the LAST describe in
+   * the file — keep it last, or capture and restore explicitly.
+   */
+  const jasmineGlobal = () =>
+    (globalThis as { jasmine?: { DEFAULT_TIMEOUT_INTERVAL?: number } }).jasmine;
+  let previousTimeout: number | undefined;
+  beforeEach(() => {
+    previousTimeout = jasmineGlobal()?.DEFAULT_TIMEOUT_INTERVAL;
+    jest.setTimeout(PRIVACY_POLICY_TEST_TIMEOUT_MS);
+  });
+  afterEach(() => {
+    if (typeof previousTimeout === 'number') {
+      jest.setTimeout(previousTimeout);
+    }
+  });
 
   const settings = (overrides: Record<string, unknown>) => ({
     privacyDefaultOn: false,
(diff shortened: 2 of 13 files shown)


Ripdiegozz (approved):
🟢 LGTM

🟢 Unit Tests:

```bash
...
 PASS   jsdom  public/hooks/use-providers.test.ts

=============================== Coverage summary ===============================
Statements   : 27.23% ( 1677/6157 )
Branches     : 22.57% ( 976/4324 )
Functions    : 32.59% ( 398/1221 )
Lines        : 27.19% ( 1647/6057 )
================================================================================

Test Suites: 6 passed, 6 total
Tests:       275 passed, 275 total
Snapshots:   0 total
Time:        26.343 s
Ran all test suites matching /(provider-name|chat-page|settings-page|use-providers|provider-form-flyout)/i in 2 projects.
Done in 27.27s.
```

🟢 The case-insensitive name & endpoint form validations

https://github.com/user-attachments/assets/c04adea7-283d-471f-a5b3-aee943cb4dd8

🟢 Updated privacy settings arrive even after a chat was created

https://github.com/user-attachments/assets/8e5e7576-5ab2-404f-b9bf-d88f8266ef85

image

PR #9006, opened 2026-08-21
Fix AI Assistant privacy-mode leaks in replayed conversation history

### LINKED ISSUE #9005: AI Assistant: real hostnames, IPs and query arguments reach the AI provider in privacy mode via replayed conversation history
| 5.0.0 | rc1  | Chrome                       |

**Description**
The AI Assistant's privacy mode (Settings > AI Assistant, toggle "Anonymize sensitive data") ships OFF by default. Testing the normal usage pattern — chat for a while with privacy off, then turn privacy on — plus a request capture against a mock provider on a live 5.0 stack surfaced several ways real hostnames, IPs, usernames and query arguments still reach the AI provider, or get corrupted, once privacy is supposed to be protecting them. Listed below with what the user sees, then the cause.

**1. Replayed history bypasses masking (main defect).**
Chat with privacy off for a few turns, then turn privacy on and keep chatting. The browser resends the accumulated conversation on the next request, and the server's outbound masking on those replayed messages is a no-op, because the client-held pseudonym map is empty at that point. The result: the whole prior transcript's real hostnames, IPs and usernames, plus the real arguments of the last two tool calls, reach the AI provider verbatim — and stay unmasked for the rest of the session. Resuming a previously-saved conversation reproduces the same leak.
Cause: the assistant-message branch and the tool-call-argument branch of the outbound scrub only did string substitution against already-known pseudonyms, with no independent shape scan; and the field policy that anonymizes agent names only runs when the server itself creates a tool digest, not when a client-supplied digest is replayed back to it.

**2. Containers under an explicit policy reached the provider unscrubbed.**
A field explicitly configured as `anonymize` or `allow-scan` whose value was an array of objects, a nested array, or a mixed-type array was sent to the provider raw. An *unlisted* field's container was correctly dropped by the fail-closed default — so the explicit policy was weaker than doing nothing.

**3. Identifiers the user types were not masked.**
Text typed directly into the chat box was only scanned for IP addresses and dotted domain names. A bare hostname or username typed by the user reached the provider unmasked even with privacy on.

**4. Stale pseudonyms could be reversed to the wrong real value.**
After resuming a conversation, the pseudonym counters restarted at zero while the restored history still contained the previous session's tokens. The assistant could then show a real IP address in its answer while attached to a claim about a different, unrelated machine.

**5. Ordinary words in questions were corrupted.**
Any unlisted field ending in `.name` is treated as a hostname for masking purposes. Common words that happen to match — process names like `top`, `find`, `less`; words like `ubuntu`, `critical`, `root` — became replacement targets inside the user's own question. The model received garbled text, but the answer shown to the user had the words restored, so the only visible symptom was the assistant seeming to misread the question.

**6. Two smaller issues found alongside the above.**
Prototype-polluting object keys (e.g. `__proto__`) in tool results silently dropped data during scrubbing instead of being handled safely. Separately, the privacy toggle's help text over-promised what it actually covers.

**Preconditions**
1. AI Assistant plugin enabled, a configured provider (verified against a mock provider), and privacy mode available in Settings.

**Steps to reproduce**
1. Start a new conversation with privacy mode OFF and ask a question that causes the assistant to reference a real hostname/IP (e.g. via a tool call).
2. Turn privacy mode ON in Settings without starting a new conversation.
3. Ask a follow-up question and capture the outbound request to the provider.

**Expected Result**
1. The server never trusts client-replayed content to be pre-protected — every outbound path gets an unconditional shape scan regardless of where the content came from.
2. History captured while privacy was off is not resent to the provider once privacy is turned on.
3. A field explicitly marked `anonymize` is never weaker than an unlisted field under the fail-closed default.
4. The privacy toggle's help text states exactly what is, and is not, covered.

**Actual Result**
1. The outbound payload after step 3 contains the real hostname/IP from the pre-privacy portion of the conversation, verbatim.

**Screenshots**
N/A — verified via request capture against a mock provider, not a visual defect.

**Additional context**

Known residuals (deliberate) — please do not report these as regressions of the fix that will close this issue:
- A bare, never-before-seen identifier, or a secret pasted by the user, is still not masked. Nothing in this design classifies free text as sensitive by content alone.
- ~30 curated common words (`root`, `admin`, `prod`, `agent`, …) and 1-2 character values are deliberately left unmasked, since masking them would destroy signal for no privacy benefit.
- Opaque provider-passthrough fields on replayed tool calls must round-trip byte-identical and are intentionally not scanned.
- `file.path` has no policy entry yet, so a username embedded inside a path is not caught by this fix (tracked as a separate follow-up).

MiguelCasaresRobles:
## Description

Closes #9005.

Privacy mode ships OFF by default, and the normal usage pattern — chat with it off, then turn it on mid-conversation — replays the accumulated history back to the server. That replayed content was never re-scrutinized: masking on it was a no-op, and the outbound scrub only did string substitution against already-known pseudonyms rather than an independent shape scan. This closes that client-replay trust boundary and the smaller gaps found alongside it while proving the fix (container policy bypass, unmasked user-typed identifiers, stale-pseudonym reversal, and `.name`-field over-masking of ordinary words).

## Proposed Changes

- **Unconditional shape scan on every outbound path.** The assistant-message branch and the tool-call-argument branch of the outbound scrub (`plugins/wazuh-ai-assistant/server/routes/chat.ts`, `plugins/wazuh-ai-assistant/server/tools/privacy.ts`) now recursively scan for sensitive shapes at any nesting depth instead of relying solely on pseudonym substitution, so replayed content is scrubbed the same way freshly generated content is.
- **`privacyEnabled` stamped per turn and enforced server-side.** Each turn/tool exchange now carries a `privacyEnabled` flag (`plugins/wazuh-ai-assistant/common/types.ts`, `plugins/wazuh-ai-assistant/common/chat-history.ts`), persisted via `conversation-store.ts` and `routes/conversations.ts`. The server enforces `excludePrivacyOffHistory`: once privacy is on, history from before it was enabled — or history with no flag at all (e.g. persisted by an older build) — is excluded from what gets sent to the provider, fail-closed.
- **Container deep-scrub under explicit policy.** Fields explicitly marked `anonymize` or `allow-scan` now get the same array-of-objects/nested-array/mixed-type deep scrub as everything else, plus a terminal passthrough rule that refuses to forward any container type it doesn't have an explicit rule for (`server/tools/privacy.ts`).
- **User-typed text scanned against minted identifiers.** Chat input is now matched against already-minted pseudonym-map entries (with a kind classifier and a stop-list of common words) instead of only regexing for IPs and dotted domains, so a bare hostname/username the user retypes gets masked too.
- **Turn history cleared when the pseudonym map resets.** Resuming a conversation no longer lets restarted-at-zero pseudonym counters collide with tokens minted in a previous session (`chat-page.tsx`, `chat.ts`).
- **Prototype-safe property assignment** when rebuilding scrubbed objects, so a tool result containing a key like `__proto__` no longer silently drops data.
- **Privacy help text corrected** in `translations/en-US.json` and `translations/es-ES.json` to state what the toggle does and does not cover, instead of implying blanket protection of user-typed text.

### Results and Evidence

Verified with a request capture against a mock provider on a live 5.0 stack:
- After toggling privacy on mid-conversation, no real hostname or IP from the pre-toggle portion of the conversation appears in the outbound payload to the provider.
- A conversation conducted entirely with privacy on still replays its pseudonym-form history correctly (no over-correction / no loss of context).
- Ordinary words in user questions (e.g. `top`, `find`, `ubuntu`, `root`) pass through to the provider unchanged, confirming the `.name`-field over-masking regression from the fix's own intermediate iterations is closed.

A UI screenshot of the reworded privacy help text still needs to be added by the developer before merge.

### Artifacts Affected

`plugins/wazuh-ai-assistant` only (common, public, server, translations).

### Configuration Changes

None. The new per-message `privacyEnabled` field is internal and backward-compatible: it is absent on history persisted by older builds, and absence is treated as "privacy was off" — i.e. excluded once privacy is later enabled, per the fail-closed rule above.

### Tests Introduced

~40 new unit tests across `common/chat-history.test.ts`, `public/components/chat/chat-page.test.tsx`, `public/components/settings/settings-page.test.tsx`, `server/routes/scrub-messages-for-provider.test.ts` (new file), `server/routes/tool-call-vendor-extras-schema.test.ts` (new file), and `server/tools/privacy.test.ts`. Full gate: 113 suites / 1929 tests green, lint clean.

## Review Checklist

- [x] Code changes reviewed
- [ ] UI screenshot of reworded privacy help text attached


yenienserrano (dismissed):
LGTM :green_circle:

Desvelao (approved):

yenienserrano (approved):

PR #9016, opened 2026-08-21
Show index and time range in the AI assistant evidence popover

MiguelCasaresRobles:
## Description

The AI assistant's evidence popover did not state which index a tool call queried nor the time range the query actually ran against, and a server-side lookback clamp produced two near-identical badges instead of one clear statement. Provenance findings from the QA E2E review of the AI assistant.

Closes #9008

## Proposed Changes

- The server records an authoritative `provenance` object on each table spec at query time: the index queried, the requested and effective time ranges read from the actual DSL (before/after the lookback clamp), whether a clamp occurred, the producing tool call id, and the execution timestamp (`executedAt`).
- The evidence popover renders only these recorded facts: an `Index:` line and a `Time range:` line. No recorded field, no line. Relative (date-math) bounds resolve against `executedAt`, never against the render clock, so restored conversations show the true window.
- Clamped queries show one combined badge (`90d · requested 2y`) at the chip level; both the requested and effective queries stay visible in the popover.
- Provenance attaches only to the tool call that produced the table (matched by id) — multi-call turns no longer cross-attribute.
- The popover closes on Escape (only when open; the event is not swallowed otherwise).
- The narrow-view "Open in Discover" fallback label states it applies the default range only to range-less queries, derived from the shared `DEFAULT_TIME_RANGE` constant.

### Results and Evidence

- Full plugin gate: 113 suites / 1898 tests passing, `yarn lint` clean, Prettier clean. CI: all checks green.
- Verified live on a deployed build: the popover shows the real recorded index and absolute time range; Escape closes it; no false clamp badges on unclamped queries.

Evidence popover showing index and resolved time range

### Artifacts Affected

- `plugins/wazuh-ai-assistant` only.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `server/tools/executor.test.ts`: provenance recorded from observed facts only; `executedAt` bounded both sides.
- `server/routes/chat-provenance-toolcallid.test.ts`: real orchestrate run threading the producing call id.
- `server/routes/conversations-table-schema.test.ts`: persistence acceptance of the new field.
- `public/components/chat/tool-call-label.test.ts`, `result-table.test.tsx`, `message-bubble.test.tsx`: rendering rules (no invented ranges, no fabricated bounds, render-clock independence), badge format, name-only truncation, Escape open/closed, raw-view provenance lines.

Issue #9008, opened 2026-08-21
Show index and resolved time range in the AI assistant evidence popover


MiguelCasaresRobles:
## Description
UX findings from the QA E2E review of the AI assistant (wazuh/wazuh-dashboard#1512), provenance group.

The evidence popover for a tool call does not show:
- which index the query ran against,
- the resolved absolute time range of the query.

When the server clamps a requested time range (90-day lookback limit), the UI shows two near-identical badges instead of one clear statement.

## Expected
- Popover lists index and resolved time range for each tool call.
- Clamped queries show one badge with both windows (for example: `90d · requested 2y`); the popover keeps both the requested and the effective query.
- Popover closes with Escape.
- The narrow-view "Open in Discover" fallback label states that it applies only to range-less queries.

Follows existing EUI popover and badge patterns.

MiguelCasaresRobles:
Fix ready for review: PR #9016 (branch `fix/9008-ai-evidence-popover-provenance`). Full plugin gate: 113 suites / 1898 tests green, lint + Prettier clean. Verified live on a deployed build: the popover shows the queried index and the resolved absolute time range; clamped queries show one combined badge; Escape closes the popover. Note: merging the PR does not auto-close this issue (non-default base branch) — close manually at merge.

PR #9017, opened 2026-08-21
Fix AI assistant results table pagination, captions and narrow readability

### LINKED ISSUE #9009: Fix AI assistant results table pagination, captions and narrow-panel readability
## Description
UX findings from the QA E2E review of the AI assistant (wazuh/wazuh-dashboard#1512), results-table group.

- The table caption reports the page count, not the total ("1 rows").
- Answers with 6+ rows hide rows behind a second page; QA observed one wrong AI prose summary caused by hidden rows.
- The row expander does not announce its state to screen readers.
- The table is unreadable at the default sidecar width (~480px): cells wrap vertically.

## Expected
- Show all rows up to 10; hide the pager when rows fit (standard EUI optional pagination).
- Correct caption and row-count text; `aria-expanded` on row toggles.
- Narrow mode: show the first priority columns from the tool's column order, truncate with tooltip, no horizontal scroll; remaining fields stay in the row expander.
- Verify at ~480px sidecar width before review.

MiguelCasaresRobles:
## Description

The AI assistant's results table split 6-10 row answers behind a hidden second page (QA observed a factually wrong AI prose summary caused by hidden rows), reported the page count instead of the total in its caption ("1 rows"), never announced the row expander's state to assistive technology, and became unreadable in narrow containers where cells wrapped vertically. Findings from the QA E2E review of the AI assistant.

Closes #9009

## Proposed Changes

- Default page size raised to 10; the pagination footer is hidden when all rows fit. Once the reader engages the page-size control, the footer stays available so the choice is always reversible.
- The accessible table caption states the total row count, plus page position when paginated; row-count text uses proper ICU plurals in both locales.
- The row expander toggle carries `aria-expanded`, a state-appropriate label, and `aria-controls`.
- **Adaptive narrow mode**: the card measures its own width (ResizeObserver, quantized to 40px steps to avoid re-render storms) and switches to priority columns whenever it cannot give each would-be-visible column a readable minimum (140px per column — a 6-column table goes narrow below ~840px, a 3-column table stays full down to 420px). In narrow mode only the first columns of the tool's existing column order render, cell text truncates to one line with a tooltip, no horizontal scroll; all remaining fields stay reachable through the existing row expander ("+N more fields" note).
- No visual redesign: at ordinary widths with fitting rows the only visible changes are the absent pager and the corrected caption text.

### Results and Evidence

- Full plugin gate: 113 suites / 1916 tests passing, `yarn lint` clean, pinned Prettier clean.
- Verified live on a deployed build at full width, ~730px (previously wrapped every cell vertically), and the ~390px docked sidecar.

Full width — total-count caption, no pager when rows fit:
Table with total caption and no pager

Narrow container — priority columns, truncation with tooltip, remaining fields via the expander:
Narrow mode with priority columns

### Artifacts Affected

- `plugins/wazuh-ai-assistant` only (public layer).

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `public/components/chat/result-table.test.tsx`: pager hidden/shown matrix incl. the reversibility case, caption totals on both pages, plural forms, `aria-expanded`/`aria-controls`, the adaptive-threshold matrix (6-col/3-col at multiple widths incl. quantization boundaries), truncation classes only in narrow mode, expander completeness.
- Mechanical plural-text updates in `message-bubble.test.tsx` / `message-list.test.tsx`.

## Notes for reviewers

- Based on the #9008 branch (both touch `result-table.tsx`); review after #9016 or by this branch's own last commits.
- Known residuals (deliberate): truncation tooltip is hover-only (full value remains reachable via the expander); caption changes are not aria-live announced on page change (matches EUI's own behavior).


Ripdiegozz (dismissed):
🟢 Tests

**1. Small result — no pagination footer at all**
A two-row answer renders complete, with neither the pager nor the rows-per-page
control taking up space below it.

Two-row agents result rendered with no pagination footer

**2. Six-to-ten rows now render in full — the regression QA caught**
The default page size moved from 5 to 10, so an 8-row answer no longer hides rows
behind a second page. This is the exact shape that produced a factually wrong AI
prose summary in the QA E2E review: the caption reads `Results (8 rows)` and the card
ends with no pagination footer at all.

Eight critical findings on a single page, no pagination footer

**3. Pagination appears only once the result exceeds one page**
A 12-row result at the default page size pages as 10 + 2, and the header caption reports
the total row count rather than the current page's.

Twelve-row result paginated 10 plus 2 at the default page size

**4. Row expanders expose their state, and the page-size control cannot trap the reader**
The expander toggle now carries `aria-expanded`, where previously it announced no state
at all. The footer also survives a reader-chosen page size larger than the result itself
(12 rows at size 25, `Page 1 of 1`), so the selector that got the reader there can always
get them back to a smaller size.

aria-expanded on the row expander, footer retained at a page size larger than the result

**5. Narrow sidecar width — readable instead of wrapped**
At the QA repro width the table keeps only its first three priority columns and states how
many were demoted ("+2 more fields per row. Expand a row to see them."). Nothing wraps into
an unreadable column and there is no horizontal scroll; every demoted field stays reachable
through the row expander.

Narrow sidecar: three priority columns kept, remaining fields moved into the row expander

**6. Large results still paginate**
A 500-row answer pages as `Page 1 of 20` at the reader's chosen page size, with only the
current page's rows mounted.

Five-hundred-row result paged 1 of 20

yenienserrano (commented):
LGTM :green_circle:

yenienserrano (approved):

PR #9018, opened 2026-08-21
Add rename and bulk delete to the AI assistant conversation rail

MiguelCasaresRobles:
## Description

The AI assistant's conversation rail offered no way to rename a conversation (auto-titles duplicate), no way to delete several conversations without repeating the full per-row flow, no confirmation feedback on delete, and no list semantics for assistive technology. Findings from the QA E2E review of the AI assistant.

Closes #9010

## Proposed Changes

- **Inline rename**: a pencil control appears as a hover/focus overlay on the row (zero footprint at rest — the resting rail is pixel-equivalent to the previous design); Enter or blur commits, Escape cancels. Persisted through a new owner-scoped `PATCH` route that mirrors the existing DELETE route's authorization exactly (fail-closed 403 on unresolved identity, 404 on wrong owner), validates and trims the title, preserves `updated_at` (a rename is not activity, so the row does not jump groups), and returns the fresh document version. Renames survive subsequent auto-saves: the update path no longer recomputes titles for existing conversations.
- **Bulk delete**: entered from a select control in the CONVERSATIONS header (the Notes/Mail "Edit" pattern); rows show checkboxes and one compact toolbar with a single "Delete N conversations?" confirmation. Partial failures are reported honestly (exact success count + error banner) and never evict the user from a conversation whose own delete failed. No "clear all" affordance. The search field keeps its full width.
- **Delete toasts** confirm single and bulk deletes; focus returns to the rail afterwards.
- **List semantics**: date groups are real headings over native lists (visually unchanged).
- Both features are also available in the rail flyout and the docked side-panel popover (the popover has no header, so it keeps an inline select control beside its search field).

### Results and Evidence

- Full plugin gate: 114 suites / 1914 tests passing, `yarn lint` clean, pinned Prettier clean.
- Route-level authorization tests drive the real router with supertest: wrong owner 404, unresolved identity 403 (with no store calls), whitespace title 400, `updated_at` unchanged.
- Regression tests pin the subtlest behaviors: a rename survives the very next auto-save; clicking the row body while renaming commits without also navigating; the rail's flex layout cannot regain the free-space bug (guard class asserted).
- Verified live on a deployed build and approved by the product owner: resting rail measured pixel-equivalent to the previous design (search at full width, groups packed, no dead space).

Final rail at rest with header select control

### Artifacts Affected

- `plugins/wazuh-ai-assistant` (public + server route + common constants).

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

[...]

PR #9027, opened 2026-08-24
Fix AI assistant Discover window and results table empty-column handling

### LINKED ISSUE #9010: Add rename and bulk delete to the AI assistant conversation list
## Description
UX findings from the QA E2E review of the AI assistant (wazuh/wazuh-dashboard#1512), conversation-rail group.

- Conversations cannot be renamed; auto-titles duplicate.
- Deleting N conversations requires N click sequences; no toast confirms a delete.
- The list lacks list semantics for screen readers.

## Expected
- Inline rename: pencil icon on hover/focus, Enter commits, Escape cancels.
- Explicit select mode with checkboxes and one "Delete N conversations?" confirmation. No "clear all" button.
- Delete shows a confirmation toast.
- Proper list semantics.

Matches the conversation-rail patterns of mainstream chat products; Gmail select-mode pattern for bulk actions.

### LINKED ISSUE #9026: Fix AI assistant Open in Discover window and results table hiding populated fields
## Description

UX findings from the QA E2E review of the AI assistant (wazuh/wazuh-dashboard#1512), result-card
group. Three defects make a result card disagree with itself or hide data that is present.

**1. "Open in Discover" opens a narrower window than the answer.**
When the executed query carries no time-range filter (a catalog listing, for example), the link
opens Discover on the last 24 hours. Discover's hit count is then lower than the total stated in the
answer above it, with nothing signalling the mismatch. The same resolver also re-evaluates relative
bounds against the current clock, so reopening a saved conversation opens a different window than
the one its evidence popover states.

- Steps: ask a question answered by a listing with no time concept (rules, detectors), then click
  "Open in Discover" on the result card.
- Expected: Discover opens the window the query actually ran against — all of history for an
  unfiltered query — and the totals match. A reopened conversation opens the window the popover
  states.
- Actual: Discover's time picker reads "Last 24 hours" and under-counts silently.

**2. Whole columns of em dashes while populated fields hide behind the expander.**
A column that is empty for every row in the result still takes a full share of the table width,
pushing a field that is populated on every row out of the visible columns and into the row expander.

- Steps: ask for critical vulnerabilities on an agent whose findings carry no description.
- Expected: a column empty across the whole result set is dropped and a populated field is promoted
  into its place; the dropped field stays readable in the row expander.
- Actual: a full column of em dashes is shown and the populated score field is hidden.

**3. The "+N more fields per row" hint is inconsistent.**
The hint counts only columns demoted past the visible budget, not the fields the row expander
actually reveals. Tables whose extra content is row-only advertise nothing, so users have no reason
to expand a row that does hold more data.

- Expected: the hint appears exactly when the expander has something extra to show, and counts it.
- Actual: present on some tables with extra fields, absent on others.

## Expected

- The Discover link reproduces the executed time window (unbounded when the query had no time
  filter) and stays pinned to the query's execution time, so it never contradicts the evidence
  popover.
- Empty-for-every-row columns are dropped and populated fields promoted; nothing disappears from the
  row expander.
- The extra-fields hint reflects what the expander shows.

## Notes

The AI assistant sidecar's resize handle has no accessible label. That control is rendered by the
platform (`wazuh-dashboard`), not by a plugin in this repository, and needs its own issue there.

MiguelCasaresRobles:
> **Stacked on #9016, #9017 and #9018** — only the last 5 commits are new; the branch will be rebased onto `5.0.0` once those merge.

## Description

Three defects on the AI Assistant's result-card surface, found in the UX Wave 2 review. The
"Open in Discover" link could open Discover on a narrower window than the query it came from, so
Discover's total disagreed with the answer above it; a column empty across the whole result set
kept its header while a populated field stayed demoted; and the "+N more fields per row" disclosure
fired on some tables with extra fields and not others.

Closes #9026

## Proposed Changes

**1. The Discover link carries the executed window (`common/discover-url.ts`,
`public/components/chat/discover-link.tsx`)**

New `resolveDiscoverTimeRange` resolves the `_g` window in a fixed precedence:

1. `TableSpec.provenance.effectiveRange` — the post-guardrail `{gte, lte}` the executor recorded off
   the body it actually ran (#9008). Taking this first makes it structurally impossible for the link
   to open a window the evidence popover contradicts, since both now read the same one fact.
2. The DSL's own range clause (`extractTimeRange`'s walk), for a table persisted before provenance
   existed.
3. `UNBOUNDED_TIME_RANGE` (epoch → now) when there is neither — the query had no time filter, so
   the link reproduces it instead of narrowing to `DEFAULT_TIME_RANGE`'s 24 hours.

Cases 1 and 2 agree on every query whose DSL states a window (they are read off the same body), so
the only behavioural move is case 3. `DEFAULT_TIME_RANGE` is deliberately untouched:
`extractTimeRange` still fills a *one-sided* clause's missing edge from it, and
`server/tools/suggest-discover-query.ts` reads it for its own disclosure text.

The label follows the window: the range-less case now reads "Open in Discover (all time)" instead of
"(default range: 24h)", which described a window the link no longer opens.

**2. Empty columns dropped, populated fields promoted (`public/components/chat/result-table.tsx`)**

New `populatedColumns` filters `spec.columns` down to those with at least one non-absent value
(`isAbsentValue`, so a genuine `0`/`false` keeps its column) *before* the visible-column budget
slices. Promotion falls out of the same slice that did the demoting — no new mechanism. Two guards:
a zero-row result filters nothing (there is no evidence of emptiness to act on), and a result where
every column is empty keeps them all rather than rendering a column-less table. A dropped column is
still in every row object, so it stays visible in the row expander like any other demoted column.

**3. The extra-fields hint counts what the expander really shows
(`public/components/chat/result-table.tsx`)**

The accordion-title suffix is now keyed on `extraRowFieldCount` — the distinct keys present on any
row that are not visible columns (excluding the synthetic `__rowId`) — instead of on the demoted
spec-column count. It therefore covers row-only `rowFields` that never had a column, and the columns
`populatedColumns` just dropped.

**Deliberately NOT in this PR** (already fixed on the Wave 1 branches this stacks on, verified by
reading their diffs): the clamped-query badge collapse (#9008 G3 — one badge, "90d · requested
720d"), the provenance-popover Escape handler (#9008 G1), the row-expander `aria-label` toggle plus
`aria-expanded` (#9009 A3), and the conversation rail's `<ul>`/`<li>` list semantics, `role='heading'`
date groups and focus-visible delete reveal (#9010 E5/m7/F-3).

### Results and Evidence

Captured live on the integration build with all nine wave PRs merged (`deploy/vm-v35`,
`d2844d3b0`) installed into the packaged 5.0 dashboard, in Chromium. One question —
*"critical vulnerabilities in the last 90 days"* — exercises all three changes at once, because the
tool it routes to (`get_critical_vulnerabilities`) has no time concept.

**1. The Discover link carries the executed window.** The accordion title reads
**"Open in Discover (all time)"**, and the provenance popover on the same row shows the index and
the query that actually ran — `{"limit": "50"}`, with no `range` clause. The label and the popover
are reading the same one fact, so they cannot disagree.

Results table with caption, hint, provenance chip and Discover link

Provenance popover: index and executed query

Following the link opens Discover on the same window the label promises — epoch to now, i.e.
`UNBOUNDED_TIME_RANGE`, not the 24-hour default:

```
metadata:(indexPattern:'wazuh-states-vulnerabilities*' …)
_q=(filters:!((meta:(alias:'AI Assistant query' …),
     query:(bool:(filter:!((match:('vulnerability.severity':'Critical')))))))
_g=(… time:(from:'1970-01-01T00:00:00.000Z',to:'2026-08-24T21:17:25.917Z'))
```

Discover reports **555** hits, which is the same total the answer stated in prose above the table —
the check that the link reproduces the query rather than a differently-scoped one.

Discover opened on the same window, 555 hits

**2. Columns carry data.** Every visible column in that table — Agent, CVE, Severity, Package,
Version, Architecture — is populated; no all-em-dash column survives into the visible budget.

**3. The extra-fields hint counts what the expander really shows.** The caption reads
**"Results (50 rows) (+1 more field per row. Expand a row to see them.)"**, and opening a row shows
exactly one field beyond the six visible columns — `vulnerability.score.base: 9.8`. Hint and
expander agree.

Row expander showing the one extra field

**Not captured:** a before/after of a table where an all-empty column was dropped. Every column the
seeded data produced for this tool is populated, so there was no empty column on the VM to drop —
the filtering is covered by the unit tests under *Tests Introduced*.

### Artifacts Affected

- `plugins/wazuh-ai-assistant` only. No server behaviour, no API surface, no configuration.
- Branch is stacked on the Wave 1 heads it shares files with:
  `fix/9009-ai-results-table-readability` (which itself carries
  `fix/9008-ai-evidence-popover-provenance`) with `enhancement/9010-ai-conversation-rail` merged in.
  Rebase onto `5.0.0` once those merge.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `common/discover-url.test.ts` (+5): `resolveDiscoverTimeRange` precedence (recorded range beats
  the DSL clause), the DSL fallback, the unbounded case (asserting it is *not* `now-24h`), that the
  unbounded lower bound is an instant Discover can resolve, and that `buildDiscoverUrl` carries the
  resolved window into `_g` time.
- `public/components/chat/result-table.test.tsx` (+8, 2 rewritten): the "all time" label and its
  suppression when the server recorded an effective range; the hint firing for row-only fields and
  counting each field once across rows; the empty column dropped, the populated one promoted, the
  falsy-but-present column kept, the zero-row and all-empty guards, and the dropped column still
  reachable in the expander.
- Two existing tests updated for the new behaviour: the "absent severity" fixture now has two rows
  (a column absent in *every* row is dropped now, which that test was not about), and the
  range-less Discover-link label assertion.

## Review Checklist

- [x] Code changes reviewed
- [ ] No unresolved dependencies with other issues (stacked on #9016/#9017/#9018)
- [x] Correct labels applied


Ripdiegozz (dismissed):
🟢 Tests

Tested on `docker/osd-dev` against live data. The three defects behave as described.

### 1. The Discover link carries the executed window

Evidence popover, tool arguments, the DSL and Discover's time picker all agree to the second. The label follows the window: plain on a bounded query, "(all time)" on a range-less one.

Image

Image

### 2. Empty columns dropped, populated ones promoted

`vulnerability.description` is empty in all 1311 documents and sits inside the visible budget. The table drops it and promotes `Architecture`.

Image

### 3. The extra-fields hint counts what the expander shows

Image

```json
{
  "@timestamp": "2026-08-25T14:11:21.301Z",
  "wazuh.agent.name": "wazuh.agent.rpm.local",
  "wazuh.rule.title": "Wazuh Rootcheck - Rootkit or malware detected",
  "wazuh.rule.level": "high",
  "wazuh.integration.category": "security",
  "_id": "ij9DOaAB9VTzZzuspTS-",
  "wazuh.rule.id": "3be437bb-4211-43e6-8610-119e642d1ede",
  "wazuh.rule.tags": [
    "high",
    "wazuh-rootcheck",
    "attack.defense-evasion",
    "attack.impact",
    "attack.t1014",
    "attack.t1564",
    "attack.t1036"
  ],
  "wazuh.rule.mitre.technique.id": [
    "T1014",
    "T1564",
    "T1036"
  ]
}
```

yenienserrano (approved):
LGTM :green_circle:

PR #9032, opened 2026-08-24
Fix AI Assistant provider settings validation and copy

### LINKED ISSUE #9030: AI Assistant settings: provider form accepts invalid input silently and mislabels failures
## Description

UX findings from the QA E2E review of the AI assistant (wazuh/wazuh-dashboard#1512), provider settings group.

The AI Assistant settings page lets several bad inputs through without saying anything, and explains some failures in words that do not match what happened. Eight defects, all on **AI Assistant > Settings**: the Providers form and table, and the Conversation history field.

Preconditions: a user with permission to manage AI Assistant providers.

1. **Empty required fields.** Click "Add provider", then "Save & test" with the form empty.
   - Actual: only Endpoint URL errors. An empty Name or Model reaches the server and comes back as a generic "Something went wrong" callout at the top of the flyout.
2. **Model typed but never committed.** Fill Name and Endpoint URL, type a model id into the Model combo box and click "Save & test" without pressing Enter.
   - Actual: the field looks filled in, but the provider is saved with an empty model. The combo box input is also not marked as required, and nothing tells the user the typed text must be committed.
3. **Endpoint URL error clears too early.** Enter `not-a-url`, click "Save & test", then type one more character.
   - Actual: the first keystroke clears the error, which reads as "fixed" while nothing is fixed. The field is also never validated on blur, only on submit.
4. **Doubled scheme.** Select provider type "Anthropic" (which prefills the endpoint), click into the field and type or paste another URL.
   - Actual: the text is appended, producing e.g. `https://api.anthropic.comhttps://my-gateway/v1`, which passes the URL check and only fails later as an opaque connection-test error.
5. **Blocked endpoint.** Save a provider whose endpoint the URL policy refuses (e.g. a cloud metadata address).
   - Actual: the failure is titled "Something went wrong", which reads as a transient glitch worth retrying.
6. **Discard confirmation.** Edit any field in the provider form, then click Cancel.
   - Actual: "Unsubmitted changes" with "Yes, do it" / "No, don't do it" — the user has to work out what "it" is, and the destructive option is the affirmative one.
7. **Delete a provider** from the row menu.
   - Actual: the row vanishes silently, which looks the same as a failure. Every other mutation on this page confirms itself.
8. **Conversation history retention.** On the Conversation history tab, clear the field and retype a value; also try `-5`.
   - Actual: clearing the box snaps the value to 0, so typing `14` produces `014`; and any unusable value is silently clamped to 0 — the one value that means "keep every conversation forever".

## Expected

- One validation pass on "Save & test" covering Name, Endpoint URL and Model, each error under its own field, focus on the first invalid one.
- The Model field is announced as required, states that typed text must be committed, and blocks the save while it holds uncommitted text.
- The Endpoint URL is validated on blur as well as on submit, keeps its error until the value is actually valid, and rejects a pasted-on-top repeated scheme; a still-suggested prefilled value is replaced by the first keystroke.
- An endpoint refused by policy is titled as blocked, keeping the reason as the body.
- The discard confirmation is a titled question with named actions, and Escape or an overlay click keeps the form.
- Deleting a provider raises a confirmation toast naming it.
- The retention field may be empty while editing, and an unusable value is rejected with an inline error that blocks the save instead of being clamped to 0.

## Additional context

Getting-started step 2 in the Add provider flyout says "Paste its API key.", which is wrong for the OpenAI-compatible local runtimes (Ollama, LM Studio) that the same form advertises.

MiguelCasaresRobles:
## Description

The AI Assistant settings page let several bad inputs through silently and explained some failures
in words that did not match what had happened: an empty Name or Model produced a server round-trip
and a generic callout, a model typed but never committed saved as an empty model, an endpoint
refused by the URL policy looked like a transient glitch, deleting a provider looked the same as a
failure, and an unusable conversation-history retention value was silently clamped to `0` — the one
value that means "keep every conversation forever".

This fixes the provider form's validation, the endpoint field's behaviour, the copy on three
surfaces, the missing delete confirmation and the retention field's clamping. No layout or
information-architecture changes.

Closes #9030

## Proposed Changes

Provider form (`provider-form-flyout.tsx`)

- **Save & test validates every required field in one pass** (Name, Endpoint URL, Model), puts each
  error under its own field and moves focus to the first invalid one in form order.
- **Model combo box**: its search input is marked `aria-required`, and Save is blocked while the
  box holds uncommitted search text (which used to submit as an empty model while the field looked
  filled in). The placeholder now names the commit step ("Pick a suggestion, or type a model id and
  press Enter"). The error is wired to the input focus actually lands on: `EuiFormRow` clones its
  a11y props onto the combo box wrapper and `EuiComboBox` forwards neither `aria-invalid` nor
  `aria-describedby` to the search input, so the field announced nothing about why it was rejected.
  Any existing `aria-describedby` is saved and restored around the error rather than removed.
- **Endpoint URL**: one shared validator used by submit, blur and the keystroke re-check, so the
  error survives the first keystroke and clears only when the value is actually valid. Blur no
  longer erases a submit-raised "required" error on a still-empty field. A type switch that
  rewrites the endpoint re-derives the error instead of keeping the one that belonged to the
  replaced value.
- **Repeated scheme** is caught by two complementary checks. "Is this a URL at all" runs first, so
  junk in front of the scheme is reported as junk. Then a *glued-scheme signature*: a second
  `http(s)://` preceded by a host or path character is the paste that was never cleared, while one
  preceded by `/`, `=`, `&` or `?` is a gateway naming its upstream on purpose. Composed with an
  authority check, so gateway/passthrough endpoints such as `https://gw/v1?upstream=https://…`
  — which the server's URL guard accepts, and which every save re-validates — stay saveable.
  Typing on top of a prefilled endpoint is therefore *caught* rather than prevented: selecting the
  prefilled value on focus was implemented and then removed, because a live measurement in Chromium
  showed the click left `selectionStart === selectionEnd === 14` — no selection at all — so typing
  still inserted mid-value. Chromium places the caret across the whole mousedown-focus-mouseup
  sequence, which undoes a `select()` made from `focus`. The validation above fires on exactly the
  URL that mis-typing produces, and the component comment records the measurement so the idea is
  not reintroduced without one.
- **"Endpoint blocked"** titles a URL refused by the SSRF/URL policy (the server's own reason stays
  as the body), in the flyout and in the providers card.
- **Discard confirmation** is titled "Discard this provider?" with named actions ("Discard changes"
  / "Keep editing"), wired so the destructive action is the modal's confirm button and Escape or an
  overlay click keeps the form.
- **Getting-started step 2**: "Paste its API key (if the endpoint needs one)." — the previous copy
  was wrong for the OpenAI-compatible local runtimes the same form advertises.

Providers table and conversation history (`settings-page.tsx`)

- Deleting a provider raises a success toast naming it.
- The retention field keeps a raw-text draft, so it can be transiently empty while editing (no more
  `014`), and rejects an unusable value with an inline error that blocks the save instead of
  clamping it to `0`. Dirtiness is one shared helper measured on the text.
- Saving the Privacy tab only resyncs the retention field when that field is clean, and decides
  that inside the state updater against the current draft rather than one captured before the
  await: both tabs stay mounted, so the field can be typed into while the request is open.

Copy

- The field-policy help no longer promises the "real value" for **Allow**: with privacy mode on,
  known identifiers in an allowed field are still pseudonymized by the server-side scrub. Updated
  in both the `en-US` and `es-ES` catalogs.

### Artifacts Affected

`plugins/wazuh-ai-assistant` (browser bundle only) and its `en-US` / `es-ES` translation files.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

Colocated unit tests for each behaviour: one-pass required-field validation and focus order,
`aria-required` on the model combo box, the model error's aria wiring (including a test pinning
that EUI sets nothing there today), the uncommitted-search-text block, the custom-option row, blur
validation, a required error surviving a blur, error persistence across a keystroke, a type switch
clearing (and not clearing) the endpoint error, glued-scheme detection plus a gateway URL carrying
a scheme in its query, the "Endpoint blocked" title (component level and `isEndpointBlockedError`
unit level), the delete toast, the in-flight Testing state, retention parsing/validation/save, and
the retention draft surviving a privacy save.

Gates: `yarn test:jest --testPathPattern="components/settings"` → 189/189 (3 suites);
`eslint public/components/settings/` clean. The suite's pre-existing delete cases were made
deterministic (the modal's confirm button was queried by a name the row menu's own Delete item also
matched).

### How to Test

1. AI Assistant > Settings > Providers > Add provider; click "Save & test" on the empty form.
2. Type a model id without pressing Enter and click "Save & test".
3. Enter `not-a-url`, submit, then type one more character; then tab out of a wrong URL.
4. Switch the type to Anthropic, click into the prefilled endpoint and type over it without
   clearing it first — the glued URL that produces must be refused with the repeated-scheme error;
   then save a gateway
   URL that names its upstream in the query (`https://gw/v1?upstream=https://api.openai.com/v1`)
   and confirm it is accepted, and can be re-saved.
5. Save a provider pointed at `169.254.169.254` to see the "Endpoint blocked" title.
6. Edit a field and press Cancel; also press Escape and click the overlay.
7. Delete a provider from the row menu.
8. Conversation history: clear the field, retype `14`; then enter `-5` and save. Type into the
   retention field, switch to Privacy, save it, and confirm the retention edit survives.

One check jsdom cannot settle, verified in a real browser:

- EUI auto-commits combo box search text on blur, so clicking Save & test with typed, uncommitted
  model text commits it rather than reaching the field error. The guard stays as belt-and-braces
  for any path where that does not happen, and the copy ("…and press Enter") remains correct for a
  keyboard user.

## Review Checklist

- [x] Code changes reviewed
- [x] Correct labels applied


yenienserrano (approved):
LGTM :green_circle:

PR #9033, opened 2026-08-24
Remove the AI Assistant translation catalogs and gate i18n source strings

### LINKED ISSUE #8975: AI Assistant: en-US translation catalog is not loaded and has drifted from the source strings
## Description

`plugins/wazuh-ai-assistant/translations/en-US.json` is not used at the default locale, and it has drifted from the source strings.

Verified on a running dashboard: the page is served with `lang="en"` while the file is named `en-US`, so the catalog is never loaded and the UI renders the `defaultMessage` written in the source. Nothing compares the two files, so the drift is invisible.

A scan of the catalog against the source found:

- **7 entries whose text no longer matches the source string.** Two are misleading:
  - `settings.type.openaiCompatible`: catalog says "(OpenAI, **Gemini**, Ollama, ...)", source says "(OpenAI, **Bedrock gateway**, ...)".
  - `settings.type.anthropic`: catalog says "Anthropic", source says "Anthropic (Claude)".
- **4 keys that no longer exist in the source.**
- **59 message IDs used in the source that are missing from the catalog.**

## Impact

No user-visible effect today, because the file is not loaded. It becomes wrong the moment anyone sets `i18n.locale: en-US`.

`es-ES.json` **is** loaded for Spanish users, so the same class of drift there would be visible immediately. That file has not been scanned.

## Suggested work

1. Decide whether `en-US.json` should exist. If the default locale never loads it, deleting it removes the problem completely.
2. If it is kept, add a unit test asserting that every key present in both files matches the source `defaultMessage`. This would have caught all 7.
3. Only then fill the 59 missing keys.
4. Scan `es-ES.json` for the same drift, since that one is user-visible.

A scan script used for this report is available and can be added to the repository if useful.

MiguelCasaresRobles:
## Description

Closes #8975

The AI Assistant shipped two translation catalogs. **Neither ever reached a user**, and both had
drifted badly from the source strings with nothing in the repo comparing them to anything.

- `translations/en-US.json` could never load. `i18n.locale` defaults to `en`
  (`src/legacy/server/config/schema.js`), and `@osd/i18n`'s loader keys catalogs by **file basename**
  (`getLocaleFromFileName` → `normalizeLocale` in `packages/osd-i18n/src/loader.ts`), so
  `getTranslationsByLocale('en')` misses the `en-US` entry and returns `{ messages: {} }`. English is
  the source language: what renders is each string's `defaultMessage`.
- `translations/es-ES.json` could never load **on a packaged install** either. The platform
  discovers catalogs through each plugin's `.i18nrc.json` (`getTranslationPaths` in
  `src/legacy/server/i18n`), and `osd-plugin-helpers` builds the archive from a `serverSourcePatterns`
  list that matches no dotfile — so the installed plugin carried `translations/` with nothing to
  register it. Spanish only ever worked when running from source.

Meanwhile the catalogs had accumulated dead ids, entries translated from wording the source had
since changed, and a long tail of ids never covered at all.

Rather than repair a Spanish island in one plugin, **this PR makes the plugin English-only.**
Localizing one plugin on its own is inconsistent for users and puts a translation cost on every
unrelated PR; localization is deferred to a general, dashboard-wide effort that can settle the
locale set and the workflow at once.

## Proposed Changes

- **Removed both catalogs** and the `translations/` directory. No user loses anything: see the
  evidence below that neither was ever loaded in a packaged install.
- **`.i18nrc.json` keeps `prefix`/`paths` with an empty `translations` array.** `getTranslationPaths`
  skips an empty list, and keeping the rest leaves the plugin visible to the platform's own i18n
  tooling for whenever localization is picked up properly.
- **Added `common/i18n-strings.test.ts`**, a gate over the source strings that survives the change of
  scope, plus two checks that fail if a catalog is reintroduced outside the wider effort.
- Updated the docs that described the old two-catalog layout, and `tsconfig.json`'s now-dead
  `translations/**/*.json` include.

### Results and Evidence

Gate, run in the OSD dev container on a copy of this branch's plugin — **11/11**:

```
PASS node common/i18n-strings.test.ts
  icuPlaceholders
    ✓ names only the arguments a caller has to supply
    ✓ ignores plural and select branch bodies
    ✓ still sees arguments that follow a nested block
    ✓ does not derail on unbalanced braces
  i18n strings
    ✓ extracts every message from the plugin source
    ✓ has no i18n usage the extractor does not understand
    ✓ namespaces every id under the plugin prefix
    ✓ never uses one id for two different messages
    ✓ renders every message
  localization
    ✓ ships no translation catalogs
    ✓ registers no catalogs in .i18nrc.json

Test Suites: 1 passed, 1 total
Tests:       11 passed, 11 total
```

Mutation-checked — **8/8 detected**, each by the intended assertion:

| Mutation                                  | Caught by                                   |
| ----------------------------------------- | ------------------------------------------- |
| stray `es-XX.json` catalog added          | ships no translation catalogs               |
| `en-US.json` creeps back                  | ships no translation catalogs               |
| catalog re-registered in `.i18nrc.json`   | registers no catalogs in .i18nrc.json       |
| `prefix` no longer matches the ids        | registers no catalogs in .i18nrc.json       |
| one id used for two different messages    | never uses one id for two different messages |
| `plural` missing its `other` branch       | renders every message                       |
| id outside the plugin prefix              | namespaces every id under the plugin prefix |
| aliased `i18n` import + call              | has no i18n usage the extractor doesn't understand |

**Evidence that the catalogs were never loaded in production** — the copy step
(`osd-plugin-helpers/src/tasks/write_server_files.ts`) replayed with its own options, upstream
default patterns:

```
[upstream defaults] 111 files copied
   .i18nrc.json      : NOT COPIED
   translations/     : translations/es-ES.json
```

The archive shipped the catalog and not the file that registers it, so `/translations/es-ES.json`
served an empty message set and the UI silently fell back to English. That is the basis for removing
the catalogs rather than repairing them.

`yarn lint` exit 0; `prettier --check` clean on every touched file; `yarn typecheck` reports no error
in any file this PR touches (its pre-existing errors are all in `server/tools/privacy*.ts` and the
platform's own `src/core`).

### Artifacts Affected

- `plugins/wazuh-ai-assistant/translations/` — removed (`en-US.json`, `es-ES.json`).
- `plugins/wazuh-ai-assistant/.i18nrc.json` — `translations` is now an empty array.
- `plugins/wazuh-ai-assistant/tsconfig.json` — dropped the dead `translations/**/*.json` include.

### Configuration Changes

None. `i18n.locale` behaviour is unchanged in practice: `en` (the default) rendered
`defaultMessage` before and still does, and `es-ES` never loaded this plugin's catalog on a packaged
install, so it renders the same English text it always did.

### Documentation Updates

- `docs/ref/modules/ai-assistant/configuration.md`: the Internationalization section now states that
  the UI chrome is English-only, why localization is deferred to a dashboard-wide effort, and that
  **model answers are unaffected** — the assistant still replies in the language of the user's own
  question (Spanish question → Spanish answer) because that comes from the system prompt, not i18n.
- `docs/ref/modules/ai-assistant/architecture.md` and the plugin README: the `translations/` line is
  gone from the layout tree.

### Tests Introduced

`plugins/wazuh-ai-assistant/common/i18n-strings.test.ts` walks `public/`, `server/` and `common/`
(`.ts`/`.tsx`/`.js`/`.jsx`) with the TypeScript AST, collecting every
`i18n.translate(id, { defaultMessage })` and every `<FormattedMessage id defaultMessage>`, then
asserts:

1. every id is namespaced under `wazuhAiAssistant.`;
2. no id is used with two different `defaultMessage` values — whichever call site is scanned last
   would otherwise decide what the user sees;
3. every message **renders** through `@osd/i18n`'s `init()` + `translate()`, once with every ICU
   argument set to `1` and once to `2`, so an unbalanced brace or a `plural`/`select` missing its
   `other` branch fails here (formatting, not just parsing — a missing `other` parses cleanly and
   only throws once a value selects it);
4. no translation catalog exists on disk, and `.i18nrc.json` registers none — with a failure message
   pointing at the dashboard-wide localization decision instead of inviting a re-add.

An AST walk rather than a regex, so concatenated messages (`'a ' + 'b'`), apostrophes and JSX text
cannot desync it, and a non-static id or `defaultMessage` fails loudly instead of being skipped. For
the same reason the walk reports i18n usage it does not understand: a `.translate(`/`.formatMessage(`
call on a receiver other than `i18n`, a `Formatted*Message` element under another name, or an aliased
`@osd/i18n` import.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Correct labels applied


yenienserrano (approved):
LGTM :green_circle:

PR #9034, opened 2026-08-24
Fix sorting in the Server management security tables

### LINKED ISSUE #9023: Server management > Security > Roles table sort not working
## Task

- [x] Fix the sorting in the Server management > Security > Roles table 

https://github.com/user-attachments/assets/30c2e7de-3cf8-4f42-a1f8-903b2875d140

Ripdiegozz:
## Description

Sorting did nothing in **Server management > Security**. Clicking a column header left the rows untouched, the arrow never moved off `ID`, and no request was sent.

Closes #9023

## Proposed Changes

- Extract `formatSorting` from `table-wz-api.tsx` into `common/tables/format-sorting.ts` so every table shares one implementation of the `+field`/`-field` convention.
- Add sorting to `usePagination`: optional initial sort, controlled `sorting` prop, and an optional third `sort` argument on `fetchFunction`. Changing the sort resets to the first page. The argument is optional, so the three existing callers keep working unchanged.
- Remove the hardcoded `?sort=` from `GetRoles`, `GetUsers` and `GetRules`. Left in place it would have emitted the parameter twice and the API would reject the request.
- Feed the `sorting` prop of the four tables from controlled state, and let `roles.tsx` use `usePagination` instead of its own copy of the same broken logic.
- Drop `sortable` from the columns the API cannot sort: `policies` on Roles, `allow_run_as` and `roles` on Users, `actions`, `resources` and `effect` on Policies, `roles` on Roles mapping.

Each table keeps the default sort it shows today, except Roles mapping. That one requested `?sort=name` while its header displayed `ID`, so the header disagreed with the data; it now defaults to `id` and the arrow matches what is rendered.

### Results and Evidence

https://github.com/user-attachments/assets/f80f096c-fc12-427a-9ea3-fcb4fa5f625b

### How to Test

- **Roles, the reported bug:** Go to `/app/security?tab=roles` and click the *Name* header. The arrow must move from ID to Name, the rows must reorder alphabetically, and a second click must reverse them. Open the network tab and check that `POST /api/request` carries `sort` **once**, as `+name` then `-name`. A duplicated `sort` means the hardcoded URL parameter came back.

- **The other three tables:** Repeat on Users, Policies and Roles mapping

- **Columns that must not offer sorting**, because the Server API rejects them: `policies` on Roles, `allow_run_as` and `roles` on Users, `actions`, `resources` and `effect` on Policies, `roles` on Roles mapping. Hovering them must show no sort control.

- **Sorting composes with the rest:** Go to page 2 and sort: the table must return to page 1. Change rows per page while sorted descending: the sort must survive. Delete a row while sorted: the refresh must keep the order.

- **Check that there is no regression in the other tables built on `TableWzAPI`**, which now imports the extracted formatter:

- `/app/endpoints-summary#/agents-preview/`
- `/app/mitre-attack` → Intelligence tab
- `/app/server-status#/manager/?tab=status&tabView=nodes`

Sorting must work on each and every request must return 200.

**Unit tests**: `yarn test:jest` from `plugins/main`, inside the `osd-dev` container.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

Colocated unit tests, none of which existed before for these files:

- `format-sorting.test.ts` — the `+field`/`-field` mapping, nested fields, and the empty-input guard.
- `usePagination.test.ts` — initial sort, sort changes, reset to the first page, and backward compatibility for two-argument `fetchFunction` callers.
- One test per `get-*.service.ts` — the sort reaches `params`, is never duplicated, and the call shape used by `useApiService` is preserved.
- One test per table — the controlled `sorting` prop is forwarded, and only the API-sortable columns are marked sortable.

45 tests pass with `yarn test:jest` in `plugins/main`.

## Review Checklist

- [x] Code changes reviewed


asteriscos (dismissed):
CR: 🟢 
Test: 🟢 

https://github.com/user-attachments/assets/4b125596-1782-4b67-be6a-37c347212c00

asteriscos (approved):

PR #9036, opened 2026-08-25
Collapse the conversation rail row actions into an overflow menu

### LINKED ISSUE #9010: Add rename and bulk delete to the AI assistant conversation list

MiguelCasaresRobles:
## Description

Follow-up to #9018, from team review feedback (https://github.com/wazuh/wazuh-dashboard-plugins/pull/9018#issuecomment-5398727064): the conversation rail's per-row actions are consolidated into a single overflow ("kebab") menu instead of two icon buttons revealed side by side on hover.

The menu offers **Rename** and **Delete**, both wired to the existing flows — the inline rename editor and the delete confirm modal are unchanged; only how they are reached changed. Select mode and bulk delete are untouched. Entries are built from an array, so adding a future action is one entry rather than another icon competing with the title for the same few pixels.

This also resolves the overlap reported on the first cut of this change (previously #9035), where the hover-revealed pencil rendered on top of the row's text: with one trigger there is no overlay — it occupies the delete button's own slot in the row's flex group, so the row stays `[title][timestamp][actions]` and text cannot be covered by construction.

Two pre-existing focus defects were found and fixed while building this:

- A menu that had been opened left its entries hidden but focusable during the popover's 250 ms close transition; entries are now unmounted with the menu.
- Choosing Delete engaged the confirm modal's focus trap while the closing menu's trap was still unwinding, so focus fell to `<body>` after deletion. The modal is now deferred one frame so only one trap unwinds at a time, and the post-delete focus hand-off to the rail works again.

### Accessibility

- Trigger: `aria-haspopup`, `aria-expanded`, revealed on hover **and** keyboard focus, kept visible while its own menu is open, and reachable on the active/selected row.
- Menu entries are keyboard reachable; Escape closes the menu without escaping to an enclosing surface; focus returns to the trigger.
- Row click still opens the conversation — the trigger stops propagation.

### Artifacts Affected

`plugins/wazuh-ai-assistant` browser bundle only.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

Colocated tests cover: trigger reveal on hover and focus (including the active row), `aria-expanded` tracking, entry order, one open menu at a time across rows, rename and delete end states through the menu (editor appears; confirm modal appears; cancel and confirm outcomes), Escape closing the menu — including nested inside the chat panel's conversations popover — focus returning to the rail after deletion, and that hovering reflows nothing. One nested-popover behavior (the outer popover surviving an inner Escape) is not assertable in jsdom and is listed below as a manual check.

## How to Test

1. Open the AI Assistant and hover a conversation in the rail — a single ⋯ button appears where the trash icon used to be.
2. Click it: Rename and Delete. Rename → the title becomes an inline editor (Enter commits, Escape cancels). Delete → the confirm modal appears; after confirming, focus returns to the conversation list.
3. Tab to the trigger with the keyboard, press Enter, navigate the entries, press Escape — focus returns to the trigger.
4. In the header-panel chat, open the conversations popover, open a row's menu, press Escape — the menu closes and the conversations popover stays open (manual check; not reproducible in jsdom).
5. Check the docked rail and the flyout rail, light and dark themes.
6. Enter select mode — the trigger is gone and bulk delete behaves as before.

### Results and Evidence

No changelog entry, matching #9018 (`no changelog`).


plugins/wazuh-ai-assistant/public/components/chat/conversation-list.tsx
```diff
@@ -1146,84 +1260,92 @@ export const ConversationList: React.FC<ConversationListProps> = ({
...
+                                      // `true`, not `'menu'`: `EuiContextMenuPanel` renders plain
+                                      // buttons, not `role="menu"`/`role="menuitem"`, and naming a
+                                      // role the DOM does not carry would send a screen reader
+                                      // looking for menu semantics that are not there. EUI's own
+                                      // examples use `true` for the same reason.
+                                      aria-haspopup='true'
+                                      aria-expanded={isMenuOpen}
+                                      onClick={(event: React.MouseEvent) =>
+                                        toggleRowMenu(event, conversation)
+                                      }
```
yenienserrano, line 1349: **Accessibility bug — keyboard users can't open this menu with Enter/Space.**

This trigger has no `onKeyDown` of its own, so `Enter`/`Space` while it's focused bubbles up to the row's own `onKeyDown` (lines 1121–1131, pre-existing since #9018), which calls `onSelect(conversation.id)` and navigates to the conversation instead of opening the menu.

Reproduced live: Tab to a row's "Conversation actions" button, press Enter or Space — it opens the conversation instead of the Rename/Delete menu.

That row-level handler predates this PR, but this PR is what gives it a new focusable trigger to collide with. The inline-rename `EuiFieldText` a bit further down (~line 1207) already guards against the same row-level handler by calling `event.stopPropagation()` in its own `onKeyDown` — the same fix would work here.

plugins/wazuh-ai-assistant/public/components/chat/conversation-list.tsx
```diff
@@ -1146,84 +1260,92 @@ export const ConversationList: React.FC<ConversationListProps> = ({
...
+                                    closeRowMenu();
                                   }
-                                  onFocus={() => setFocusedId(conversation.id)}
-                                  onBlur={() =>
-                                    setFocusedId(current =>
-                                      current === conversation.id
-                                        ? null
-                                        : current,
-                                    )
+                                }}
```
yenienserrano, line 1321: **Accessibility bug — Escape closes the menu but focus lands on `<body>`, not the trigger.**

`closeRowMenu()` + `stopPropagation()` here appear to bypass `EuiPopover`'s own `ownFocus` return-focus handling (the popover is `ownFocus`, see the `<EuiPopover>` just below this block).

Reproduced live: after pressing Escape, `document.activeElement` is `<body>` — not the trigger button — on both the full-page rail and the nested header-panel popover.

This also contradicts the PR description's own stated requirement: "Escape closes the menu without escaping to an enclosing surface... focus returns to the trigger." Suggest explicitly re-focusing the trigger ref after `closeRowMenu()`, or routing through `EuiPopover`'s own Escape handling instead of intercepting it here.

plugins/wazuh-ai-assistant/public/components/chat/conversation-list.tsx
yenienserrano: Fixed in e8df1e79f. Turned out to be broader than just this trigger: the row's own `onKeyDown` was also swallowing Enter/Space from the Rename and Delete menu items (they're React-tree descendants of the row too, since `EuiPopover` only portals their DOM location, not their place in the component tree). Consolidated the guard once at the shared `EuiFlexItem` wrapper instead of duplicating it per control. Live-verified with agent-browser: Enter opens the menu, and Enter on both Rename and Delete now activates them.

plugins/wazuh-ai-assistant/public/components/chat/conversation-list.tsx
yenienserrano: Fixed in e8df1e79f: added a per-row trigger ref and explicitly refocus it (deferred one frame, same pattern this file already uses for the delete-modal focus-trap handoff) after closing on Escape, since intercepting Escape here to protect nested-popover isolation bypasses EuiPopover's own ownFocus return-focus step. Live-verified with agent-browser: document.activeElement is the trigger button after Escape now, on both the plain rail and the nested header-panel popover.

yenienserrano (approved):
LGTM :green_circle:

PR #9039, opened 2026-08-25
Remove data grid columns with missing fields

### LINKED ISSUE #9037: Remove findings data grid column if field doesn't exist
# Description

The Findings tab data grid throws an exception every time it tries to load a column using a non-existent field from the index pattern. We need to add validation when initializing the data grid, verifying that the configured columns use valid fields. In case the field doesn't exist, it should remove the column entirely.

Image

Ripdiegozz:
## Description

The Findings data grid crashed when a configured column pointed at a field missing from the index pattern. The tab showed "Something went wrong." and nothing else.

`useDataGridColumns` seeded its visible columns from `defaultColumns` without checking them against the index pattern. With no matching entry in `columnSchemaDefinitionsMap`, `{ ...columnSchemaDefinitionsMap[columnId] }` spread `undefined` into `{}`, a column with no `id` and no `schema`. `EuiDataGrid` called `.toLowerCase()` on it:

```
TypeError: Cannot read properties of undefined (reading 'toLowerCase')
  at EuiDataGrid (osd-ui-shared-deps.@elastic.js:161932:37)
```

Fixing that alone stopped the crash but left the grid with zero columns, so this PR covers all three places an unknown column id causes damage.

Closes https://github.com/wazuh/wazuh-dashboard-plugins/issues/9037

## Proposed Changes

**`use-data-grid-columns.ts`: build no malformed column.** A single derived `effectiveVisibleColumns` now feeds `columns`, `columnVisibility.visibleColumns` and the `columnsAvailable` ordering. Those came from different sources before, so filtering one of them would hand `EuiDataGrid` a visible id with no definition behind it. The mapper falls back to `{ id: columnId }`, which makes the empty-object shape unreachable.

The value stays derived and never returns to state, so a column comes back on its own once its field returns to the index pattern. It also stays outside `useMemo`: the comment above it warns that memoizing breaks the DataGrid cell filter, and `effectiveVisibleColumns` feeds that same array. Filtering waits until the index pattern loads, because `indexPattern.fields` arrives asynchronously and an early filter would drop every column.

**`use-data-grid-state-persistence-manager.ts`: drop the invalid column, keep the rest.** `validateColumns` threw on the first invalid id and the `catch` reset `state.columns` to `[]`, then persisted it. One stale id erased every persisted column. The comment above that block already described the intent, "to avoid clearing the entire state if only columns are invalid", so `sanitizeColumns()` now removes the invalid, empty and duplicated ids and keeps the others.

**`use-data-grid-columns.ts`: stop applying a reset list.** The two effects read `columns: []` differently. The first treats it as nothing persisted and returns. The second applied it through `setVisibleColumns`, skipping the guard in `setVisibleColumnsHandler`. That mismatch emptied the grid. The second effect now agrees with the first.

### How to Test

1. Start the environment: `cd docker/osd-dev && ./dev.sh up`.
2. Index one finding so the grid has a row to render. The grid does not mount without data:

```bash
curl -sk -u admin:admin -X POST "https://localhost:9200/wazuh-findings-v5-cloud-services/_doc?refresh=true" \
  -H "Content-Type: application/json" -d '{
    "@timestamp": "<now, ISO 8601>",
    "wazuh": { "cluster": { "name": "wazuh" }, "agent": { "name": "test.agent", "id": "001" },
               "integration": { "name": "gcp" }, "space": { "name": "standard" } },
    "event": { "action": "google.cloud.audit.test", "outcome": "success", "category": "configuration" }
  }'
```

3. Add a column with an id absent from the index pattern to `plugins/main/public/components/overview/google-cloud/events/google-cloud-columns.tsx`:

```ts
{ id: 'gcp.this_field_is_not_in_the_index_pattern' },
```

4. Open Google Cloud and click the **Findings** tab. Reaching it by URL renders a blank page, so use the tab.

Before the fix the tab shows "Something went wrong." with the `toLowerCase` `TypeError` under Details. After it, the grid renders and the invalid column is absent from the grid and from the field selector. Every valid column keeps its data and its persisted width, including `cloud.*` and `dns.*`, which live in the index pattern with no data behind them.

Revert the extra column when finished.

### Results and Evidence

**Before**:

image

**After**:

image

### Artifacts Affected

None. The changes are browser-side and contained to the shared data grid in `plugins/main`.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

`yarn test:jest --testPathPattern="data-grid"` passes 65 tests across 7 suites. Reverting the source change fails 5 of them.

The new cases cover the unknown column leaving both outputs, the two loading gates, unchanged output when every field exists, preserved column widths, `columnsAvailable` consistency, a column reappearing when its field returns, and the persistence manager keeping the valid ids.

Two existing tests asserted the old behaviour and now assert the corrected contract. Calling this out so it does not read as tests bent to fit the code:

- `should initialize with default columns` expected `toHaveLength(0)` and `toEqual([])`, which contradicted its own name.
- `should reset columns when invalid column ids are found` expected every column to be dropped. It is now `should drop only the invalid columns and persist the sanitized list`.

A `renderColumns()` helper replaces 18 repeated `renderHook` blocks, which accounts for the deletion count.

## Review Checklist

- [x] Code changes reviewed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,7 @@
 | [#8775](https://github.com/wazuh/wazuh-dashboard-plugins/issues/8775) | Fixed MITRE ATT&CK Framework tab not applying the date range and fetch filters                                                                                                                                                                                                        |
 | [#9015](https://github.com/wazuh/wazuh-dashboard-plugins/issues/9015) | Fixed the Controls tab showing 0 results by adding missing regulatory compliance requirement definitions for PCI DSS, GDPR, HIPAA, NIST 800-53, FedRAMP, ISO 27001, CMMC, TSC, and NIS2, and added an "Others" breakdown showing findings tagged with unrecognized requirement values |
 | [#9023](https://github.com/wazuh/wazuh-dashboard-plugins/issues/9023) | Fixed sorting not working in the Server management > Security tables (Roles, Policies, Users and Roles mapping), and removed the sort affordance from columns the server API cannot sort                                                                                              |
+| [#9037](https://github.com/wazuh/wazuh-dashboard-plugins/issues/9037) | Fixed the Findings data grid crashing and losing every column when a configured column's field does not exist in the index pattern                                                                                                                                                    |
 
 ### Removed
 
--- a/plugins/main/public/components/common/data-grid/data-grid-state-persistence-manager/use-data-grid-state-persistence-manager.test.ts
+++ b/plugins/main/public/components/common/data-grid/data-grid-state-persistence-manager/use-data-grid-state-persistence-manager.test.ts
@@ -87,7 +87,7 @@ describe('useDataGridStatePersistenceManager', () => {
     });
   });
 
-  it('should reset columns when invalid column ids are found', () => {
+  it('should remove invalid column ids and keep the valid ones', () => {
     const persistedState = {
       columns: ['col1', 'nonExistentColumn'],
     };
@@ -109,7 +109,7 @@ describe('useDataGridStatePersistenceManager', () => {
     });
 
     expect(mockStateManagement.retrieveState).toHaveBeenCalled();
-    expect(retrievedState.columns).toEqual([]);
+    expect(retrievedState.columns).toEqual(['col1']);
   });
 
   it('should reset column widths when invalid', () => {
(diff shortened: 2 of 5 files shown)


rodrigofez (approved):
🟢 Test

Reverted the two files to the parent commit and the Findings tab crashes with the `toLowerCase` error. With the fix the grid renders, and the invalid column is gone from both the grid and the field selector. The nine valid columns keep their data and widths.

image

Desvelao (approved):

Issue #1532, opened 2019-06-14
Update NIST 800-53 requirements descriptions


juankaromo:
Hi team,

Since the first iteration of the dashboards for NIST 800-53, the requirements we support have been modified and the API now returns some new ones and has eliminated others. So the changes we have to implement are:

New additions (We'll have to add your descriptions)
```
AU.5
AU.12
CM.5
SI.2
SI.3
```

Removed
```
AU.1
AU.3.1
MA.2
SI.5
```

Regards.

PR #9045, opened 2026-08-26
Fix assistant defects reported in team review

### LINKED ISSUE #1524: Privacy mode ON: a real monitored hostname reaches the AI provider inside `wazuh.rule.title`
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | `wazuhCore` / `wazuh` / `wazuhCheckUpdates` / `wazuhAiAssistant` all `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents: Ubuntu 24, Rocky 9, Amazon Linux 2023, Windows 11, Windows Server 2022) |
| Provider used | OpenAI-compatible scratch endpoint (local capture listener) — see "Why a scratch provider" |
| Area / probe | Area 6a, probes **P1** and **P3** |
| Proposed severity | **MAJOR** |
| Known-issue match | **None.** KN-01…KN-11 are all answer-quality issues; none covers privacy. |

## Summary

With privacy mode **ON**, the real hostname of a monitored Windows host is sent verbatim to the external LLM provider, inside the `wazuh.rule.title` field of a tool result. Field-scoped anonymization works (`host.hostname`, `wazuh.agent.name`, `package.vendor` are all rewritten correctly) but `wazuh.rule.title` is on the `allow` list, and Wazuh rule titles embed the hostname or account name by construction.

## What the product promises

Settings → AI Assistant → Privacy, verbatim:

> Control whether finding data is anonymized before reaching the configured AI provider. When privacy mode is off, hostnames, IP addresses, usernames, process command lines, **and finding/rule text** leave the cluster as-is.

Finding/rule text is named explicitly, so with privacy ON it is claimed to be pseudonymized.

## Question asked (verbatim)

```text
Show me the details of the most frequent finding.
```

Asked in a fresh conversation, privacy control set to **On**, against the scratch provider.

## Verbatim response excerpt — the captured provider payload

From `capture.log`, request **3**, `messages[3]` with `role: "tool"` (the tool result on its way out of the cluster). The leaked value is marked:

```text
…"doc_count":10,"wazuh.rule.title":"Secret or credential accessed from vault - >>>AI-QA-AGENT-WIN<<<$","wazuh.rule.level":"medium","distinct_titles":1,…
```

`AI-QA-AGENT-WIN` is the real NetBIOS hostname of the `ai-qa-agent-windows11` agent. 4 occurrences, in requests **3, 6, 7 and 8**.

Two sibling titles in the same payload show the embedding is routine for a whole class of rules:

```text
"Special privileges assigned to new logon - SYSTEM"
"Successful user authentication - SYSTEM"
```

## Expected vs got

**Expected** — no real hostname anywhere in the outbound payload; a `HOST_n` pseudonym in its place, including inside rule/finding text.

**Got** — the hostname passes through verbatim. And the `privacy_map` event shows that across **all five probes, in every turn**, the complete set of substitutions performed was:

```json
[{"value": "lists.ubuntu.com", "pseudonym": "HOST_1"},
 {"value": "ubuntu.com",       "pseudonym": "HOST_2"}]
```

Two **public Ubuntu mailing-list domains**, from `package.vendor`. **Not a single value belonging to the environment under test was pseudonymized** in the whole run, while a real monitored hostname left the cluster.

## Root cause

Anonymization is field-scoped; the sensitive value lives in a field the policy allows, so the pseudonymizer never looks at it.

| Field | Policy | What happened |
|---|---|---|
| `host.hostname` | `anonymize` / `HOST` | correct — but did not appear as a field in these tool results |
| `wazuh.agent.name` | `anonymize` / `HOST` | idem |
| `wazuh.agent.host.ip` | `anonymize` / `IP` | idem — 0 real IPs in the capture |
| `package.vendor` | `anonymize` / `VAL` | correctly rewritten to `Ubuntu Developers <ubuntu-devel-discuss@HOST_1>` |
| **`wazuh.rule.title`** | **`allow`** | **verbatim** |

The pseudonymizer does work on fields it is pointed at — it rewrote FQDN-shaped tokens inside `package.vendor`. The gap is *which* fields it is pointed at.

## Why this is structural rather than a single bad row

The field policy on this build has **116 rows: 12 `anonymize`, 102 `allow`, 2 `allow-scan`**. Many of the `allow` rows are free-text name/title fields where the same embedding can occur:

```text
wazuh.rule.title, wazuh.rule.mitre.technique.name, wazuh.rule.mitre.tactic.name,
document.metadata.description, check.name, policy.name, package.hotfix.name,
host.os.name, process.name, wazuh.space.name, metric.name, name, monitor_name,
queries.name, queries.query_field_names, rule.metadata.title, document.name,
document.feed.name, document.software.name
```

A per-field allow/deny list cannot close this class on its own. Two candidate directions: move `wazuh.rule.title` (and its siblings) to `allow-scan` so value-shaped scanning applies, or extend the scan to every allowed free-text field.

## Reproduction

1. Profile A AIO with at least one Windows agent reporting findings.
2. On the dashboard host, start a capture listener that speaks SSE and can issue a tool call — `privacy_capture_listener_sse.py` (attached): `python3 privacy_capture_listener_sse.py 8099 capture.log <linux-agent-name>` *(the listener shipped in `qa-package.zip` cannot be used here — see the Area 6a tracking issue, package finding 2)*
3. Settings → AI Assistant → Providers → add an OpenAI-compatible provider with base URL `http://127.0.0.1:8099/v1` and any dummy key. Select it.
4. Set the chat's privacy control to **On**.
5. In a fresh conversation ask: `Show me the details of the most frequent finding.`
6. `grep -F 'AI-QA-AGENT-WIN' capture.log` → present, inside a `role:"tool"` message's `wazuh.rule.title`.

## Why a scratch provider

The code that reverses pseudonyms runs before the browser receives the response, so the screen shows real hostnames whether or not the outgoing request was scrubbed. Only the dashboard→provider request body is a valid test, and it is TLS-encrypted to a real provider. Pointing a scratch provider at a local plaintext listener is the procedure the handoff itself prescribes (`§5b` / `§6a`).

## Evidence attached

[capture.log](https://github.com/user-attachments/files/31452247/capture.log)
[pre_test_snapshot.json](https://github.com/user-attachments/files/31452268/pre_test_snapshot.json)
[probe_events.json](https://github.com/user-attachments/files/31452280/probe_events.json)
[privacy_capture_listener_sse.py](https://github.com/user-attachments/files/31452285/privacy_capture_listener_sse.py)

**No screenshot:** this defect is not observable in the UI by construction — the Dashboard correctly shows the real hostname to the operator either way. The captured payload is the evidence, which is what the handoff's own privacy-finding rule requires ("attach the captured payload, with the leaked value marked").

## Root cause, named precisely (added after the Area 6a leftover pass)

`privacy.js` documents a compensating control for exactly this residual risk on `wazuh.rule.title`:

> The residual risk is real, not hypothetical, though: a LOCAL/custom rule's `<description>` can interpolate a decoder capture group … That is not left unaddressed — it is covered at a different layer: chat.ts's `scrubMessagesForProvider` runs `prescanAndMintToolContent` over every tool-result string value (this one included) before it reaches the provider.

**That control cannot see this value.** It mints by value *shape* — `IPV4_TOKEN_RE`, `IPV6_TOKEN_RE`, `FQDN_TOKEN_RE` — and `FQDN_TOKEN_RE` requires at least one dot. The source records the limitation and names the fallback:

> a bare single-word hostname ("webserver", no dot) is NEVER matched … A bare-word hostname is still caught later, once it appears in a typed digest field, by the existing field-policy scrub in `applyFieldPolicy` (e.g. `wazuh.agent.name`).

`AI-QA-AGENT-WIN` is a bare, single-label NetBIOS-style name — the default shape for a Windows agent — so it matches no regex, and the fallback is the field-policy scrub, which is the thing that said `allow`. The two mitigations cover for each other and the value passes both.

The second scan that could have caught it, `scrubKnownEntities`, matches bare identifiers **already minted elsewhere in the same conversation**. On a single-turn question about rules, nothing has minted the agent name yet, so its dictionary is empty for that value. It only applies to `allow-scan` fields in any case, and `wazuh.rule.title` is `allow`.

**So the fix point is one of:** make `wazuh.rule.title` `allow-scan` rather than `allow` (it is free text influenced by third parties, which is exactly what `allow-scan` is documented for), or seed `scrubKnownEntities`'s dictionary with the known agent names at the start of every turn so a bare hostname is catchable on first mention.

**DEFECT-5 is the same two-step on a different field** (`wazuh.cluster.node`, `allow` via a wildcard, bare hostname `ai-qa-aio-node`). Fixing the shared step closes both.

### LINKED ISSUE #1526: `get_sca_checks`'s `check_id` is accepted and ignored, so a single-check question returns a truncated policy-wide sample
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | all four Wazuh plugins `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents) |
| Provider used | **Anthropic** (`claude-opus-5`) |
| Where found | Area 3 supplement — `CV-053`, `CV-042` |
| Proposed severity | **MAJOR** |
| Known-issue match | **Adjacent to KN-06, not a match.** See "Relation to KN-06". |

## Summary

Two witnesses, filed as a single issue per the dedupe rule: a tool result omits exactly the detail the question is about, even though it is populated in the index, and explicit follow-up requests for it are rejected as invalid field names.

- **Witness A (SCA).** `get_sca_checks`'s `check_id` argument is **accepted and silently ignored**. The call returns a 20-row sample of the whole 279-check policy instead of the one check asked about, so the requested check is usually not in the rows at all.
- **Witness B (inventory).** `search_wazuh_data` locates 8 documents by an `exists` filter on `network.type` and returns only agent/OS metadata; the explicit `network.*` projection in the follow-up call is rejected as a bad field name.

**Scope correction (post human pass).** An earlier version of this issue said `check.remediation` is "never projected". That was wrong — it **is** returned, with full text, when the same tool is called with a `result` filter and an explicit `limit`. See "The field is reachable" below. The defect is narrower and more precise than first written: the data is retrievable, and a documented parameter silently does nothing.

**The assistant's own behaviour is correct throughout** — it states plainly what it could not retrieve and why, and never fabricates the missing text. The defect is in the tools' filtering and projection, not in the model. Please do not triage this as an answer-quality problem.

---

### Question asked (verbatim)

```text
How do I remediate failed check ID 35509?
```

Check `35509` was chosen because it exists on this deployment. (The battery's own `CV-053` names check `28500`, which does not exist here — ours start at `26000` — so the shipped item could not test this at all. See the Area 3 tracking issue.)

### Ground truth in the indexer

`GET wazuh-states-sca/_search` for `check.id: "35509"`:

```text
result:      Failed
agent:       ai-qa-agent-ubuntu24
remediation: PRESENT — "- IF - the module is available in the running kernel: - Unload the filesystem kernel modul…"
rationale:   PRESENT
fields:      compliance, condition, description, id, mitre, name, rationale,
             references, remediation, result, rules
```

### What the tool did

```text
tool_call: get_sca_checks {"check_id": "35509"}
  digest:  counts {total: 279, returned: 20, truncated: true}
  columns: ['check.id', 'check.name', 'check.result', 'check.rationale']
```

Two distinct problems in that single call:

1. **`check_id` did not filter.** 279 total, returned as a truncated 20-row sample of the whole policy (checks `35500`–`35504`). Check `35509` is not in the sample. The argument was accepted and had no effect.
2. **`check.remediation` is absent from this call's rows** — the one field the question is about, and it is populated in the document.

**Correction (added after the human pass, see "The field is reachable" below): `check.remediation` is NOT missing from the tool in general.** It is returned, with full text, when the same tool is called with a `result` filter and an explicit `limit`. The original wording of this issue said it was never projected; that was wrong, and the corrected mechanism is narrower and more precise: `check_id` does not filter, so the rows returned are a 20-row sample of the whole policy, and the requested check is simply not among them.

The follow-up escape-hatch call requesting the field explicitly was **rejected as an invalid field name**:

```text
tool_call: search_wazuh_data {"index_pattern": "wazuh-states-*",
             "query_dsl": {"query": {"bool": {"filter": [{"term": {"check.id": "35509"}}]}},
                           "_source": ["check.id", "check.name", "check…"]}}
```

### Verbatim response excerpt

```text
I can't give you the remediation text for **check `35509`** — it wasn't returned.
The check lookup resolved `35509` to agent `001` and policy `cis_ubuntu24-04`,
but the rows that came back were a truncated sample (**20 of 279**, newest-end
only) covering checks `35500`–`35504`, and a follow-up query for the check itself
failed on an invalid field name.
```

It named the gap, the sample bounds, and the rejected query. Exemplary handling of a broken tool.

---

### Question asked (verbatim)

```text
What network protocols is agent "ai-qa-agent-ubuntu24" using?
```

### Ground truth in the indexer

A document in `wazuh-states-inventory-protocols` for that agent carries:

```json
{"network": {"metric": 100, "type": "ipv4", "dhcp": false, "gateway": ["172.31.0.1"]},
 "interface": {…}, "state": {…}}
```

### What the tool did

```text
tool_call: search_wazuh_data {index_pattern: "wazuh-states-*",
             query_dsl: {… exists: network.type …}, _source: [...]}
  digest:  counts {total: 8, returned: 8}
  columns: ['wazuh.agent.name', 'checksum.hash.sha1', 'wazuh.cluster.name',
            'wazuh.agent.host.hostname', 'wazuh.agent.host.os.name',
            'wazuh.agent.host.os.type', 'wazuh.agent.host.os.version',
            'wazuh.agent.host.os.platform']
```

The rows were **located by an `exists` filter on `network.type`**, so the field is populated — yet the returned columns are agent and OS metadata only.

Precise reading of the trace, confirmed on a re-run (see "Re-run with privacy off" below): that **first** call carries **no `_source` at all**, so what came back is `search_wazuh_data`'s default projection — a fixed set of agent/OS/checksum fields rather than the content of the documents it just matched. The tool located 8 documents by a `network.type` filter and returned none of the network content. The model's **second** call then requests the projection explicitly, and *that* is the one rejected as a bad field name.

### Verbatim response excerpt

```text
I found **8 network-state records** for `ai-qa-agent-ubuntu24`, but the actual
protocol values were not returned — the rows that came back carried only agent
and OS fields (`wazuh.agent.host.os.name` = `Ubuntu` `24.04.4 LTS`), so I can't
name which protocols are in use.

- The **8 rows** are confirmed to have a populated `network.type` field; their
  values simply weren't included in the returned columns.
- Attempts to request the protocol and interface fields explicitly were rejected
  as unrecognized field names, so the mapping for them is unconfirmed.
```

---

## Expected vs got

**Expected**

- `get_sca_checks {check_id: "35509"}` returns **that one check**, with `check.remediation` and `check.rationale` among its fields — as the same tool already does when called with a `result` filter.
- `search_wazuh_data` over `wazuh-states-inventory-protocols` returns the `network.*` fields it was asked for.
- An explicit `_source` request for a field that exists in the mapping is not rejected as an invalid field name.

**Got**

- The `check_id` filter is ignored; a truncated 20-row sample of the whole 279-check policy is returned; the requested check is not in it; explicit follow-up requests for that single check are rejected as an invalid field name.
- Rows are found by an `exists` filter on a field whose value is then not returned.

## Re-run with privacy off — this is a tools defect, not a privacy artefact

Both questions were re-run against the Anthropic provider with privacy **OFF** and **ON** to rule out interference from the privacy layer, which is known to rewrite some field names into pseudonyms (**DEFECT-6**). Evidence: `ai-qa-results/area-6a/leftovers/probe_defect3.json`.

**With privacy OFF, both witnesses reproduce in full.**

`CV-053`, privacy off:

```text
tool_call: get_sca_checks {"check_id": "35509"}
  columns: ['check.id','check.name','check.result','check.rationale']
  279 checks, truncated to 20, sample covers 35500-35504
tool_call: find_document_by_field {"index_pattern":"wazuh-states-*","values":["35509"]}
  columns: []            <- zero rows
answer:    "I can't give you the remediation text for check 35509 - the results I
            gathered don't contain that check's own record."
```

`check_id` does not filter, and the requested check is absent from the returned rows, with no pseudonymization anywhere in the turn.

`CV-042`, privacy off: the second call specifies `_source: ["wazuh.agent.name","network.type","network.dhcp","network.gateway","network.metric","observer…"]` and the answer still reports *"my follow-up query for the protocol fields failed on a bad field name"*.

**So neither symptom is caused by privacy mode.** This removes the one confound that could have deflated this issue's severity, and it means DEFECT-6 does not need to be fixed first.

### Side observation from the same run — privacy ON gave a worse answer (n = 1)

For `CV-042`, the privacy-on turn produced a materially worse result than the privacy-off turn on the same question:

| | privacy OFF | privacy ON |
|---|---|---|
| Answer length | 1,086 chars | **112 chars** |
| Tool calls | 2 | 3 |
| First call's filter field | `wazuh.agent.name` | `agent.name` (wrong path) |
| Field probed | `network.type` (exists) | `network.protocol` (not a field here) |
| Answer | names the gap, the 8 rows, the OS, what is missing | *"No matching results were found for that query."* |

One observation on one question, recorded rather than claimed — it is not enough to file, but it is the kind of thing the Area 7 answer-quality pass should look for deliberately.

## The field is reachable — the human pass found the recovery path

A tester running Chapter 1 step 1 hit this defect, and the assistant offered a way out:

> *"Want me to pull the failed checks for that policy so `35509`'s own rationale and remediation come back directly?"*

They answered **yes**, and it worked. Reproduced on the API twice (`ai-qa-results/human-pass/repro2.json`):

```text
tool_call: get_sca_checks {"agent_id": "001", "policy_id": "cis_ubuntu24-04",
                           "result": "failed", "limit": "200"}
  counts:  {total: 117, returned: 117, truncated: false}
  samples[0]: {"check.id": "35509",
               "check.name": "Ensure unused filesystems kernel modules are not available.",
               "check.result": "Failed",
               "check.rationale": "While loadable filesystem kernel modules are a convenient…",
               "check.remediation": "- IF - the module is available in the running kernel:
                                     - Unload the filesystem kernel module … /etc/modprobe.d/ …"}
```

Check `35509` comes back as the **first sample row, with its full remediation text**.

**Three consequences, all of which narrow this issue rather than widen it:**

1. **`check.remediation` is projected.** The earlier claim in this issue that it is "never projected" was wrong and has been corrected above. The tool can return it.
2. **The real defect is `check_id`.** It is accepted, silently ignored, and the resulting rows are a 20-row sample of a 279-check policy — so a single-check question fails by sampling, not by projection.
3. **The recovery is not reliable.** In the tester's run, answering "yes" produced the wide call and the correct answer. In an API re-run of the same two turns, the follow-up re-issued `get_sca_checks {check_id: "35509"}` **again** and failed the same way a second time. So the offered escape works only when the model happens to widen the call.

### A separate, smaller inconsistency in the same digest

On the call that works, the digest's `columns` reads:

```text
["check.id", "check.name", "check.result", "check.rationale"]
```

while every `samples` row carries **five** keys — the four above **plus `check.remediation`**. The schema hint the model is given does not list a field the rows actually contain. Worth fixing alongside, and worth checking whether it contributes to the model not using the remediation text when it is present.

## Relation to KN-06

KN-06 is *"The SCA checks result table renders rows where every cell is a dash. The `Reason` column empty"*. Plausibly the same root cause in the SCA detail projection — but the symptom here is different (fields absent from the projection entirely rather than rendered blank), and it **also reproduces outside SCA**, on `search_wazuh_data` over the inventory indices. Filed as new with the relationship flagged rather than suppressed under a row whose described symptom does not match. Fold them together if the projection layer turns out to be one piece of code.

## Why this matters

It silently narrows two whole question classes — *"explain / remediate this specific finding"* and *"what does this host's X look like"* — which are among the most common things an analyst asks. Mitigated by the assistant's honesty: it never invents the missing text, so the failure is visible rather than dangerous.

## Screenshot

Image

Image

## Reproduced on a second environment — with a worse symptom

Reproduced in the Dashboard UI on a **second, freshly provisioned** Profile A deployment of the same build. The `check_id` filter failed the same way:

```text
I could not retrieve the remediation text for check 35509. The lookup resolved it
to agent 001 and policy cis_ubuntu24-04, but the result came back as the full
279-check policy (truncated to 20 rows) rather than that single check, and the
sampled rows cover 35500-35504 only - a follow-up document lookup returned 0 rows.
...
I can't assume 35509 belongs to that theme, and **I can't confirm its result is
Failed**.
```

**And the table rendered directly below that sentence contains:**

```text
35509 | Ensure unused filesystems kernel modules are not available. | Failed | <full rationale>
```

### The asymmetry this exposes

The digest handed to the model carries **5 sample rows** (35500–35504); the table rendered to the user carries **20**, including 35509 with `result: Failed`. So the assistant truthfully reports that it cannot confirm something **the user can read on screen, in the answer's own table**.

That makes the defect worse than "a field is missing": the model is blind to data the UI is already showing. It also explains why the assistant's honesty reads as confusing rather than helpful here — it is being accurate about its own inputs while the screen contradicts it.

Note also that the `REASON` column **is** populated in this run (with `check.rationale`), so **KN-06** — *"SCA checks table renders all-dash rows, Reason column empty"* — **did not reproduce**. That weakens the KN-06 relationship hypothesized below: this is more likely its own defect.

### Evidence

[supplement.json](https://github.com/user-attachments/files/31452934/supplement.json)

[probe_defect3.json](https://github.com/user-attachments/files/31452947/probe_defect3.json)

### LINKED ISSUE #1527: The same answer is emitted twice, in full, reworded
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | all four Wazuh plugins `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents) |
| Provider used | **Anthropic** (`claude-opus-5`) |
| Where found | Area 3, battery item **CV-009** |
| Proposed severity | **MINOR** — cosmetic and confusing, no wrong data |
| Known-issue match | **Adjacent to KN-11, not a match.** |

## Summary

A single turn delivers **two complete, standalone answers to the same question**, reworded. The user reads the same content twice.

## Question asked (verbatim)

```text
List all 345 decoders
```

## Verbatim response — the whole answer, both stages

```text
There are actually **506 decoders** in this deployment, not 345 — the table below
shows **500** of them (the maximum a single listing can return), so **6 rows are
truncated**.
- All the returned decoders sit in the **`standard`** space and are **enabled**,
  covering log sources like `decoder/iis/0`, `decoder/kafka-log/0`, and
  `decoder/containerd-daemon/0`.
- The count spans **every space** (no space or enabled filter applied), so it
  includes any disabled entries as well.
If you want the remaining rows or a narrower slice (e.g. only custom-space
decoders), I can filter by name or space.
This deployment has **506 decoders**, not 345 — the table below lists **500** of
them, the maximum a single listing returns, so **6 rows are truncated**.
- Every returned decoder is **enabled** and belongs to the **`standard`** space.
- The set covers a broad range of log sources, including `decoder/iis/0`,
  `decoder/kafka-log/0`, `decoder/containerd-daemon/0`, `decoder/wazuh-dashboard/0`,
  and `decoder/microsoft-exchange-online-message-trace/0`.
- Coverage note: this is the Security Analytics decoder catalog across all spaces;
  the Manager API's own decoder set was not queried, and I could not produce a
  Discover link for the 6 remaining rows on this turn.
Want me to narrow the listing (for example by name pattern or space) so the full
matching set fits under the limit?
```

Both halves open with the same corrected count, both describe the same truncation, both end with the same follow-up offer.

## Expected vs got

**Expected** — one answer per turn.

**Got** — two complete answers, 1,369 characters total for a question whose answer fits in about half that.

## Two things worth noting for triage

**The second stage is the better one.** It carries a coverage statement the first lacks: *"this is the Security Analytics decoder catalog across all spaces; the Manager API's own decoder set was not queried, and I could not produce a Discover link for the 6 remaining rows on this turn."* If one stage has to win, it is the second — that sentence is exactly the scope disclosure the scale-aware contract asks for.

**The answer itself is good.** The question contains a false premise ("345 decoders") and the assistant corrected it with a live count — there are 506 — rather than playing along. That part is a pass; only the duplication is the defect.

## Relation to KN-11

KN-11 is *"Text from two answer stages is joined with no space or blank line"*. Same family — multi-stage answer assembly — but a different symptom: here the stages are properly separated and **both are full answers**. KN-11's concatenation symptom appears separately in the same run (`CV-036`, and Area 3a unit 13), so this build exhibits **two distinct defects in multi-stage answer assembly**. Filed as new with the relationship flagged.

## Screenshot

Image

## Reproduced on a second environment, and a second witness

Reproduced in the Dashboard UI on a **second, freshly provisioned** Profile A deployment of the same build:

```text
stage 1: "The table below lists **345 decoders**, but that isn't the full set -
          there are actually **506 decoders** in the pipeline catalog across all
          spaces, so the list is truncated at the 345 you asked for." + 2 bullets

stage 2: "**To recap:** the table below shows **345 decoders**, all in the
          `standard` space and all enabled, but the catalog actually holds
          **506** - so this is a truncated view, and my per-call ceiling of
          **500 rows** means no single listing can cover all of them." + 2 bullets
```

Stage 2 opens with "To recap:", which reads marginally more natural than the first occurrence, but it is still the same complete answer delivered twice.

**Second witness, different question:** Area 4's turn 15 (*"Compare that to last week."*) does the same thing — see `area-4/CHECKLIST.md` Observation 5. So the duplication is not specific to the decoder-listing question.

## Reliable reproduction — Area 4 turn 15, 3 of 3 runs

The 20-turn Area 4 script was run three times on the same build. Turn 15 (*"Compare that to last week."*, asked right after a vulnerability question) produced **two complete, standalone answers in every run**.

**This is a deterministic reproduction recipe**, which this issue previously lacked: run the Area 4 script and read turn 15.

Run 3, verbatim, both stages:

```text
I can't do that comparison: vulnerability data in Wazuh is current-state only — the
index holds what's outstanding right now, with no per-day snapshots and no
patched/resolved history, so there's no "last week" baseline to diff against. The
92 vulnerabilities on agent 005 I just reported are today's state only.
- What I *can* compare over time is findings (rule matches), which are timestamped…
- If you track it externally, exporting today's 4 Critical / 50 High counts gives
  you a baseline for future comparisons.
Discover link below for the raw current-state vulnerability documents on this agent.

I can't do that comparison: vulnerability data in Wazuh is current-state only — the
index holds what's outstanding right now, with no daily snapshots and no
patched/resolved history, so there's no "last week" baseline to diff against. The
92 vulnerabilities on agent 005 (4 Critical, 50 High, 32 Medium, 2 Low, 4 unscored)
reflect today's state.
- What *is* time-based is findings data — rule matches are timestamped, so
  week-over-week comparisons work there, just not for vulnerability state.
- Recording today's counts gives you a manual baseline for future comparisons.
- I wasn't able to produce the Discover handoff link for the raw current-state
  documents on this turn.
Want me to break the current 92 down by package instead, so you can see which
upgrades clear the most at once?
```

The pattern from CV-009 holds here too: **the second stage is the better one**. It carries the severity breakdown and the honest note that the Discover link could not be produced — both absent from the first. Whatever assembles these two stages is discarding the wrong one when it picks, and emitting both when it does not.

Answer lengths for turn 15 across the three runs: 1,054 / 1,534 / 1,493 characters, against a campaign mean of 782 — the duplication is visible in the length alone.

### Evidence

[CV-009.md](https://github.com/user-attachments/files/31453029/CV-009.md)

[runs-2-3.zip](https://github.com/user-attachments/files/31453051/runs-2-3.zip)

### LINKED ISSUE #1528: Privacy mode ON: the cluster node name is sent to the AI provider in clear, while its field name is pseudonymized
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | all four Wazuh plugins `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents) |
| Provider used | `openai_compatible` scratch endpoint (local capture listener) |
| Privacy mode | **ON** |
| Where found | Area 6a leftover spot checks |
| Proposed severity | **MAJOR** |
| Known-issue match | none — KN-01…KN-11 are all answer-quality |

## Summary

With privacy mode ON, the digest sent to the AI provider carries the deployment's real cluster node name in clear, under a **pseudonymized field name**. The two halves are exactly inverted: the identifier that should have been replaced was kept, and the field name that should have been kept was replaced.

## Wire evidence

Same `search_wazuh_data` query over `wazuh-metrics-*`, run twice, privacy off then on. Both captures are in `capture_directed.log` (requests 23 and 20).

**privacy OFF** — baseline, what the data actually holds:

```json
{"wazuh.agent.name":"ai-qa-agent-ubuntu24",
 "wazuh.agent.host.ip":["172.31.32.160"],
 "wazuh.cluster.node":"ai-qa-aio-node"}
```

**privacy ON** — what reached the provider:

```json
{"wazuh.agent.name":"HOST_2",
 "wazuh.agent.host.ip":["IP_1"],
 "HOST_1":"ai-qa-aio-node"}
```

`ai-qa-aio-node` is the AIO manager's own node name — an operator-chosen identifier for a real machine.

**Reproduced independently** on a different index and query (request 38, five sample rows over `wazuh-events-v5-*`):

```json
{"wazuh.agent.name":"HOST_5","HOST_4":"ai-qa-aio-node"}
```

## Root cause

Two independent defects meeting on one field.

**1. `wazuh.cluster.*` is `allow`.** The shipped policy carries a single wildcard row, `{field: "wazuh.cluster.*", action: "allow"}`. `privacy.js`'s own contract for that action is explicit:

> `allow`: the provider receives the real value, **completely unscanned**.

and it states what the action is meant for:

> Reserved for CURATED vocabulary (MITRE technique names, compliance ids, `check.id`, rule tags/category/title) whose values are not analyst/attacker/third-party-supplied free text.

`wazuh.cluster.name` fits that description — it is the cluster's configured label, commonly the default `wazuh`. **`wazuh.cluster.node` does not.** It is a hostname chosen by the operator, unique to the deployment, and it identifies a specific machine on every single document that carries it. The wildcard sweeps a curated label and a host identity into the same decision.

This is the same class as **DEFECT-1** (`wazuh.rule.title`, also `allow`, carrying `AI-QA-AGENT-WIN`) but a cleaner instance: rule titles are *mostly* curated vocabulary and identity leaks through them only when a custom rule interpolates a capture group. A cluster node name is identity **100% of the time**, so there is no usefulness argument for keeping it readable — the model never needs the real node name to answer anything.

**2. The compensating scan cannot see it.** `privacy.js` documents a second layer for exactly this residual risk — `chat.ts`'s `scrubMessagesForProvider` runs `prescanAndMintToolContent` over every tool-result string value. That scan mints by **value shape**: `IPV4_TOKEN_RE`, `IPV6_TOKEN_RE`, `FQDN_TOKEN_RE`. `FQDN_TOKEN_RE` requires at least one dot, a limitation the source records deliberately:

> a bare single-word hostname ("webserver", no dot) is NEVER matched … A bare-word hostname is still caught later, once it appears in a typed digest field, by the existing field-policy scrub in `applyFieldPolicy` (e.g. `wazuh.agent.name`).

That fallback is precisely what fails here: the field-policy scrub is the thing that said `allow`. So the bare-hostname gap and the `allow` decision cover for each other, and `ai-qa-aio-node` passes both.

**This is the same two-step that produces DEFECT-1**, and it is worth fixing at the shared step rather than field by field.

## Expected vs got

**Expected** — with privacy ON, no identifier of a real machine in the monitored estate reaches the provider. `wazuh.cluster.node` arrives as `HOST_n`, and the sample key stays `wazuh.cluster.node` so the model can still tell what the column means.

**Got** — the value arrives in clear and the key arrives as `HOST_1`.

## Suggested fix

Split the wildcard:

```js
{ field: 'wazuh.cluster.name', action: 'allow' },              // curated label
{ field: 'wazuh.cluster.node', action: 'anonymize', kind: 'HOST' },
```

Keeping a wildcard for the rest is fine, but a wildcard should not be able to grant `allow` to a field nobody reviewed individually — that is how both this and DEFECT-1 happened. Consider making `allow` reachable only by an **exact** field entry, with wildcards limited to `allow-scan` or stricter. `allow-scan` would have caught this on its own: it is `allow` plus both scans, and the known-entity dictionary scan (`scrubKnownEntities`) would match `ai-qa-aio-node` once any sibling turn had minted it.

## Why this matters

Privacy mode's whole contract is that the provider does not learn the identity of the customer's machines. This is a manager node name, present on effectively every document the metrics and events tools return, so it is sent on **most** privacy-on turns rather than in a rare corner case — the two probes here hit it without looking for it, on two unrelated queries.

### Evidence

[capture_directed.log](https://github.com/user-attachments/files/31453112/capture_directed.log)

[REPORT.md](https://github.com/user-attachments/files/31453118/REPORT.md)

[pre_snapshot.json](https://github.com/user-attachments/files/31453120/pre_snapshot.json)

[probe_directed.json](https://github.com/user-attachments/files/31453123/probe_directed.json)

[probe_close.json](https://github.com/user-attachments/files/31453124/probe_close.json)

### LINKED ISSUE #1529: Privacy mode ON: field names in the digest are replaced by `HOST_n` pseudonyms, corrupting the schema hint the model reasons over
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | all four Wazuh plugins `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents) |
| Provider used | `openai_compatible` scratch endpoint (local capture listener) |
| Privacy mode | **ON** |
| Where found | Area 6a leftover spot checks |
| Proposed severity | **MAJOR** — not a privacy leak, but it degrades every privacy-on answer that touches an unrecognized field |
| Known-issue match | none |

## Summary

With privacy mode ON, **field names** in the digest — both the `columns` schema hint and the keys of `samples` rows — are treated as hostnames by the value-shape scan and replaced with `HOST_n`. The model is handed a table whose columns are named `HOST_1`, `HOST_2`, `HOST_3`.

This contradicts an invariant the code states about itself. `applyFieldPolicy`:

> `sampleKey` is what the digest stays KEYED by (**never rewritten** — the model's view of the digest shape must not change); `field` is only what the policy is resolved against.

`applyFieldPolicy` honours that. A later layer does not.

## Wire evidence

`search_wazuh_data` over `wazuh-events-v5-*`, same query twice.

**privacy OFF** (request 41):

```json
"samples": [{"related.hosts":["ai-qa-agent-amazon2023"],
             "related.ip":["100.53.8.44"],
             "related.user":["ec2-user"],
             "wazuh.agent.name":"ai-qa-agent-amazon2023",
             "wazuh.cluster.node":"ai-qa-aio-node"}]
```

**privacy ON** (request 38):

```json
"samples": [{"wazuh.agent.name":"HOST_5","HOST_4":"ai-qa-aio-node"}],
"columns": ["HOST_1","HOST_2","HOST_3","wazuh.agent.name","HOST_4"]
```

Four field names were minted as hostnames:

| Real name | Sent as |
|---|---|
| `related.hosts` | `HOST_1` |
| `related.ip` | `HOST_2` |
| `related.user` | `HOST_3` |
| `wazuh.cluster.node` | `HOST_4` |

Also observed: the index pattern `wazuh-alerts-5.x` minted as `HOST_1` in another turn, and the filename `containerd.log` minted as `HOST_1` in a `get_threat_intel_components` result.

## Root cause — and the exact rule for which names survive

`prescanAndMint`'s FQDN pass excludes a token that "reads as the user NAMING A FIELD" via `isFieldPathToken`:

```js
const FIELD_PATH_WORDS = new Set(
  [...Object.values(WAZUH_FIELD), ...FIELD_POLICY_DEFAULTS.map(e => e.field)]
    .map(f => f.split('/').pop())
    .flatMap(f => f.replace(/\.\*$/, '').split('.'))
    .map(s => s.toLowerCase()).filter(s => s.length > 0));

function isFieldPathToken(token) {
  return token.split('.').every(seg => FIELD_PATH_WORDS.has(seg.toLowerCase()));
}
```

The guard is a set of **path segments**, and it requires **every** segment of the token to be known. That explains the otherwise-baffling split observed on the wire, where two sibling fields behave differently:

| Token | Segments | In vocabulary? | Result |
|---|---|---|---|
| `wazuh.cluster.name` | `wazuh`, `cluster`, `name` | all three — `cluster` from the `wazuh.cluster.*` row, `name` from `wazuh.agent.name` | survives |
| `wazuh.cluster.node` | `wazuh`, `cluster`, **`node`** | `node` appears in no `WAZUH_FIELD` value and no default policy row | **minted `HOST_n`** |
| `related.hosts` | **`related`**, `hosts` | neither (`host.hostname` contributes `host`, not `hosts`) | **minted `HOST_n`** |

So the surviving set is not "field names" — it is "field names assembled entirely from words the plugin already ships". Any ECS field outside the curated vocabulary, and any Wazuh field whose leaf word is new, is mangled.

**Two consequences worth calling out separately:**

1. **The guard reads `FIELD_POLICY_DEFAULTS`, not the live policy.** It is a module-scope constant built from the compile-time defaults. An admin who adds a policy row for a new field at runtime does **not** get that field's name recognized — the row will govern the value while the name is still mangled. The source comment says *"Self-updating: a field added to either source is automatically recognized here — no separate list to hand-maintain."* That holds for a field added **in code**; it does not hold for one added through the settings UI.

2. **The wildcard contributes only its literal prefix.** `wazuh.cluster.*` yields the segments `wazuh` and `cluster`, so it cannot vouch for any leaf under it — which is exactly how a field covered by an explicit policy row still gets its name mangled.

## Expected vs got

**Expected** — with privacy ON, values are pseudonymized and field names are not. `columns` and the `samples` keys arrive intact so the model can tell what each column holds.

**Got** — three of five column names arrive as opaque pseudonyms, and a sample row reads `{"HOST_4": "ai-qa-aio-node"}`.

## Why this matters

The digest's `columns` list exists to tell the model what it is looking at. When it arrives as `["HOST_1","HOST_2","HOST_3","wazuh.agent.name","HOST_4"]` the model has no way to know that `HOST_1` means `related.hosts`. Its options are all bad: ignore the columns, guess, or describe them to the user by their pseudonyms.

It also burns pseudonym slots on non-identities — the same effect recorded in Area 4 Observation 3, where `wazuh.rule.tags` values (`attack.execution`, `attack.persistence`) consumed 14 of 19 `HOST_n` slots. Both come from the same shape scan treating dotted vocabulary as hostnames.

Note the asymmetry with DEFECT-5: in the one sample row `{"HOST_4": "ai-qa-aio-node"}`, privacy mode substituted the part that carried no secret and preserved the part that did.

## Relationship to DEFECT-3 — tested, and it is NOT the cause

DEFECT-3's two witnesses both feature the assistant requesting a field explicitly and the call being rejected as an invalid field name, which made this a plausible shared root cause. **It was tested and ruled out**, so DEFECT-3 should be triaged on its own merits.

Both DEFECT-3 questions were re-run against the Anthropic provider with privacy **OFF** and **ON** (`ai-qa-results/area-6a/leftovers/probe_defect3.json`).

**With privacy OFF, DEFECT-3 reproduces in full.** `CV-053`:

```text
tool_call: get_sca_checks {"check_id": "35509"}
  columns: ['check.id','check.name','check.result','check.rationale']
  279 checks returned, truncated to 20, sample covers 35500-35504
tool_call: find_document_by_field {"index_pattern":"wazuh-states-*","values":["35509"]}
  columns: []           <- zero rows
answer:    "I can't give you the remediation text for check 35509 ..."
```

`check_id` still does not filter and `check.remediation` is still absent from the projection, with no pseudonymization anywhere in the turn.

`CV-042` with privacy OFF is equally clear — the model's **second** call does specify the projection explicitly:

```json
"_source": ["wazuh.agent.name","network.type","network.dhcp","network.gateway","network.metric","observer…"]
```

and the answer still reports *"my follow-up query for the protocol fields failed on a bad field name"*. Privacy was off, so no field name in that turn was ever replaced by a pseudonym.

**Conclusion:** the invalid-field-name rejection in DEFECT-3 has a different cause. The two issues are independent and neither blocks the other.

## Second symptom of the same mechanism — `allow` values are scanned too, and the selectivity is inverted

The same FQDN shape scan also mangles **values**, and it does so on fields whose policy action explicitly promises the opposite.

`wazuh.rule.tags` is `allow`. `privacy.js`'s contract for `allow` is *"the provider receives the real value, **completely unscanned**"*. Yet across all three Area 4 long-conversation runs, every privacy-on turn minted the same pseudonyms for that field's values:

```text
attack.execution → HOST_1        attack.persistence → HOST_2
attack.t1543     → HOST_3        attack.defense-evasion → HOST_4
attack.impact    → HOST_5        attack.t1489 → HOST_6
```

Deterministic, 3 of 3 runs. In run 1's accounting, **14 of 19 `HOST_n` slots** were spent on MITRE vendor taxonomy.

Two conclusions follow, and both matter more than the wasted slots:

1. **The `allow` contract is not honoured.** Something downstream scans `allow` values anyway. That is arguably a good thing for safety — it is the very mitigation `privacy.js` cites for `wazuh.rule.title`'s residual risk — but the documented contract and the implementation disagree, and the tracking issue for **DEFECT-1** rests on which of the two is true.

2. **The scan's selectivity is exactly inverted.** In the same campaign, on the same build:

   | Value | Shape | Sensitive? | Outcome |
   |---|---|---|---|
   | `attack.execution` | dotted | no — MITRE taxonomy | **pseudonymized** |
   | `wazuh.cluster.node` (a field *name*) | dotted | no — schema | **pseudonymized** |
   | `AI-QA-AGENT-WIN` | bare | **yes — real hostname** | sent in clear (DEFECT-1) |
   | `ai-qa-aio-node` | bare | **yes — real node name** | sent in clear (DEFECT-5) |

Everything dotted is treated as a hostname; nothing bare is. Real host identifiers in this estate — and in most Wazuh estates — are bare single labels, so the scan catches the harmless class and misses the sensitive one.

This is why the structural fix below is preferable to widening the vocabulary list: the underlying problem is that "contains a dot" is being used as the test for "is a hostname", in a payload that is mostly dotted field paths and dotted taxonomy.

### Evidence

[capture_directed.log](https://github.com/user-attachments/files/31453264/capture_directed.log)

[REPORT.md](https://github.com/user-attachments/files/31453267/REPORT.md)

[REPORT.md](https://github.com/user-attachments/files/31453275/REPORT.md)

## Suggested fix

Do not run the value-shape scan over structural positions at all. `columns` entries and `samples` keys are schema, not data — `applyFieldPolicy` already treats them that way, and the whole-JSON text pass downstream does not know the difference. Scrubbing tool content as a flat string is what erases that distinction; walking the digest structurally and scanning only leaf **values** would fix this and DEFECT-5's key-rewrite half in one change.

Failing that, seed `FIELD_PATH_WORDS` from the live `settings.fieldPolicy` as well as the compile-time defaults, and add the ECS `related.*` family. That is a narrower fix and leaves the next unrecognized field name to be found the same way.

### LINKED ISSUE #1530: A provider connection dropped mid-answer is finalized as a complete answer
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | all four Wazuh plugins `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents) |
| Provider used | `openai_compatible` fault-injection endpoint |
| Where found | **CV-083**, run with fault injection |
| Proposed severity | **MAJOR** — a partial answer presented as a complete one, with no signal |
| Known-issue match | none |

## Summary

When the connection to the AI provider drops **after** the answer has started streaming, the plugin finalizes the turn normally: the partial text is kept, a `done` event is emitted, and **no `error` event is sent**. The user is shown an answer that stops mid-thought, with nothing to indicate it was cut off.

**Reproduced 3 of 3 attempts.**

## Reproduction

Register an `openai_compatible` provider pointing at a local endpoint that streams two content deltas and then closes the socket with no `[DONE]` frame and no `finish_reason`. `fault_listener.py` in the attachment does this — mode `drop-mid-stream`. Ask any normal question.

**Event stream received:**

```text
status: Routing…
status: Querying Wazuh…
delta:  'Here are the top rules for today: the busiest is rule 5501 (PAM login). '
done
```

**Baseline, identical question and provider, no fault injected:**

```text
delta:  'Here are the top rules for today: the busiest is rule 5501 (PAM login). That is the full picture.'
done
```

97 characters against 72. The turn terminates the same way in both cases.

## Expected vs got

**Expected** — CV-083's outcome (6): an infrastructure-failure notice in distinct copy. The plugin already has exactly the right copy for its sibling failure modes:

| Fault | What the user is told |
|---|---|
| 429 on the data call | *"The AI provider rejected this request due to rate limits — try again in a moment."* |
| Provider stops responding | *"The AI provider stopped responding."* |
| HTTP 500 | *"Provider responded with HTTP 500: Internal server error"* |
| **Connection dropped mid-answer** | **nothing** |

At minimum the partial text should be marked as incomplete and an `error` (or a terminal `status`) emitted. The retry machinery that handles the other three modes, including visible *"retrying in 30s…"* progress, does not engage here at all.

**Got** — a truncated answer, indistinguishable from a complete one.

## Why this is worse than a silent empty

Forbidden outcome (a) in CV-083's own list is *"silent empty response with no explanation"*. This is not that — it is one step worse. A silent empty at least signals that something went wrong. A silently truncated answer looks like the assistant simply finished, so a reader has no reason to doubt it or retry.

Concretely: the sentence that was lost in the reproduction was the summary line. An answer that stops after naming one rule reads as "that is the answer", not as "the answer was cut off".

## The recovery path exists — it just cannot fire here

Dropping the connection **before** the first delta does *not* produce this. The plugin falls back to a no-tools retry (`synthesizeNoTextFallback` in `server/routes/chat.js`) and presents what that returns. So a recovery path is already implemented and works.

It cannot help once text has been emitted, because the partial text is already committed to the turn — and that is precisely the window this defect occupies.

## Suggested fix

Treat "stream ended without a `finish_reason` / `[DONE]`" as a failure rather than as a normal end of stream. Two options, in preference order:

1. Retry the completion the way a 429 or 500 is retried, and if the retry also fails, emit the same class of infrastructure-failure notice — the copy already exists.
2. If the partial text must be kept, keep it *and* emit an error, so the UI can mark the answer as incomplete rather than final.

## Context — this closes CV-083

`battery.json` flags CV-083 `ambiguous: true` because the shipped runner cannot force a provider failure. With the injector attached, five fault modes were run:

- 429 on the data call — **PASS**
- HTTP 500 on the data call — **PASS** (final copy is raw plumbing; minor)
- provider stalls — **PASS**
- 429 on the router call — recovers correctly after 60 s, but **silently**: the *"retrying in 30s…"* status the data stage emits is not wired to the router stage, so a rate-limited router call is indistinguishable from a hang. MINOR, recorded in `cv083/REPORT.md`.
- **connection dropped mid-stream — FAIL, this issue.**

The injector is offered for folding into `qa-package.zip` so CV-083 stops being an un-runnable item; see `cv083/REPORT.md` for the two operational notes it needs (threaded server, detached start).

### Evidence

[cv083.zip](https://github.com/user-attachments/files/31453381/cv083.zip)

### LINKED ISSUE #1531: `get_rules` cannot look up a rule by its UUID, and reports the rule does not exist
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | all four Wazuh plugins `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents) |
| Provider used | **Anthropic** (`claude-opus-5`) |
| Where found | **Human pass, Area 5 / Chapter 1 step 3**, then reproduced on the API |
| Proposed severity | **MAJOR** |
| Known-issue match | none |

## Summary

Asked *"What does rule `<uuid>` detect?"*, the assistant sometimes answers correctly and sometimes states the rule **does not exist** — same rule, same environment, different conversation. Which happens is decided entirely by the tool the model reaches for:

| Path | Result |
|---|---|
| `search_wazuh_data` on `.opensearch-sap-pre-packaged-rules-config` | **1 row** — answers correctly |
| `get_rules {name: "<uuid>", enabled: "any"}` | **0 rows** — *"I couldn't find any rule matching …"* |

`get_rules` has **no filter that can match a rule id**. Its `name` filter targets the rule's human title, so a UUID passed to it can never match — by construction, not intermittently.

## How the tester found it

Chapter 1 step 3. In one conversation the assistant gave a full, correct answer:

```text
Rule ad97a19d-24a5-43c4-a749-1f8f0a9172bc is "System service started - {{service.name}}"
— an informational rule that fires when a system service or daemon starts… It has 117
findings in the last 90 days, all mapped to T1543 (Create or Modify System Process)…
```

Opening a **new conversation** and asking the same question returned:

```text
No matching results were found for that query. (Searched: rules, filtered to name
ad97a19d-24a5-43c4-a749-1f8f0a9172bc, enabled any.)
```

### Evidence

[repro.json](https://github.com/user-attachments/files/31453445/repro.json)

[repro2.json](https://github.com/user-attachments/files/31453453/repro2.json)

| Tool path taken | Samples | Outcome |
|---|---|---|
| `search_wazuh_data` on the rules index (alone, or after a broad unfiltered `get_rules`) | 11 | correct answer |
| `get_rules {name: "<uuid>"}` only | **1** | **"I couldn't find any rule matching …"** |

One in twelve on the API, plus the tester's hit in the UI. The rate is low; the failure is fully deterministic once that path is taken.

## Ground truth — the rule exists, and the UUID is in no name field

`GET .opensearch-sap-pre-packaged-rules-config` for that UUID returns exactly one document:

```text
_id:                     ad97a19d-24a5-43c4-a749-1f8f0a9172bc
rule.document.id:        ad97a19d-24a5-43c4-a749-1f8f0a9172bc
rule.metadata.title:     "System service started - {{service.name}}"
rule.mitre.technique.id: ["T1543"]
```

The UUID lives in `_id` and `rule.document.id`. The title is a human string. `get_rules`'s `name` filter matches the title, so it cannot match the id — and there is no `id` parameter to use instead.

## The tool's own description sends the model down this path

From `get-rules.js`:

> Lists Security Analytics correlation rules … **These are Sigma-shaped rules identified by a UUID and a title** … use this tool for "what does the SSTI/SSH/… rule detect" … questions **asked by name or topic**.

The description states these rules are identified by a UUID and directs the model here for "what does rule X detect". It never says the tool cannot filter by that UUID. A model following the description lands on `get_rules {name: <uuid>}` — which is exactly what happened.

## Expected vs got

**Expected** — `get_rules` accepts a rule id (or UUID-shaped lookups route to a tool that can resolve one), and *"what does rule `<uuid>` detect"* answers consistently.

**Got** — a confident, well-written statement that the rule does not exist:

```text
I couldn't find any rule matching `ad97a19d-24a5-43c4-a749-1f8f0a9172bc` — the
lookup against the Security Analytics rule catalog returned 0 rows, so I can't tell
you what it detects.
- The search covered the Security Analytics Sigma/UUID-shaped rule catalog (enabled
  and disabled rules alike); the Wazuh Manager's own ruleset API was not queried…
- A zero-row result means that identifier didn't match in the catalog searched…
```

Every sentence is true *about the tool call*, and the conclusion the reader draws is false. It even says it searched the "UUID-shaped rule catalog" — which is precisely where the rule is.

## Why this is more serious than its reproduction rate suggests

**It is a wrong answer wearing the clothes of a careful one.** The other answer-quality defects in this campaign either visibly collapse (DEFECT-2) or say plainly that something could not be retrieved (DEFECT-3). Here the assistant delivers a clean, sourced negative. An analyst has no cue to doubt it, and the natural next step — concluding the rule id is stale or the catalog is out of sync — is wrong.

It also lands on **Q1 of the ten top-demand analyst questions** (*"What does rule ID NNNNN mean / what does it detect?"*), the most-travelled path in the module.

## Suggested fix

1. Add an `id` / `rule_id` parameter to `get_rules` filtering on `rule.document.id` / `_id`.
2. Until then, make a UUID-shaped `name` argument either fall back to the id field or return a message the model can act on, rather than a silent zero.
3. Amend the tool description so "identified by a UUID" is not asserted by a tool that cannot search by one.

### Screenshots

Image

Image

### LINKED ISSUE #1533: The chat input is disabled while an answer streams, and focus is not returned when it re-enables
| Field | Value |
|---|---|
| Build SHA | `7e935b0f9` |
| Plugin versions | all four Wazuh plugins `5.0.0-06` |
| Environment profile | **A** (AIO + 5 agents) |
| Where found | **Human pass — free-exploration UX observation**, then measured at the DOM |
| Proposed severity | **MINOR** — friction on every single turn |
| Known-issue match | none |

## Summary

Two related problems on every turn of every conversation:

1. **While an answer streams, the message box is `disabled`** — you cannot compose the next question while you read the current answer.
2. **When streaming ends the box is re-enabled but focus is not restored** — the caret is gone and you must click the box again before typing.

## Measured at the DOM

Polled `document.querySelector('textarea')` and `document.activeElement` every 1.5 s across a full turn (`input_state.json`):

| Phase | `textarea.disabled` | `document.activeElement` | Stop button |
|---|---|---|---|
| Before sending | `false` | `TEXTAREA[Chat message]` | absent |
| t = 1.5 s … 7.5 s (streaming) | **`true`** | **`BODY`** | present |
| t = 9.0 s (stream ended) | `false` | **`BODY`** | absent |
| +2.5 s after completion | `false` | **`BODY`** | absent |

`readOnly` stays `false` throughout; the block is the `disabled` attribute.

**Reproduced in both surfaces** — the full page (`/app/wazuhAiAssistant`) and the side panel opened from `wzAiAssistantHeaderButton` — with identical readings.

## Root cause

Disabling a focused element blurs it, and nothing re-focuses the textarea when it is re-enabled. That single fact explains both halves of the report: the user cannot type during the stream *because* of the `disabled` attribute, and cannot type immediately afterwards *because* the blur it caused is never undone.

## Expected vs got

**Expected** — the message box stays editable throughout, so the next question can be composed while the current answer streams; and when a turn ends the caret is back in the box, ready to type.

**Got** — the box is disabled for the duration of the stream, and after it ends the user must click into it before typing.

## Suggested fix

- Keep the **send action** disabled while a stream is in flight (or have Enter queue the message), but leave the **textarea** enabled so the user can type and read at the same time. This is the behaviour of every mainstream assistant chat, which is the comparison the tester drew unprompted.
- On stream completion, call `.focus()` on the textarea. If the input must stay disabled during streaming, this focus restore is the minimum fix and is a one-liner — it removes the per-turn click on its own.

## Why it is worth fixing despite being MINOR

It costs an extra click and a lost train of thought on **every turn of every conversation** — the 20-turn Area 4 script pays it twenty times. It is also the one piece of friction a new user notices in the first minute, and it invites exactly the comparison the tester made: this chat behaves worse than the chats they already use.

### Screenshots

Image

Image

### LINKED ISSUE #9044: Fix assistant defects reported in team review
## Description

The team review of the AI assistant (`wazuh/wazuh-dashboard#1523`) produced ten child findings. This issue tracks the subset that is fixable without a design decision, in the `wazuh-ai-assistant` plugin (`server/` + `public/`).

### Addressed here

- `wazuh/wazuh-dashboard#1526` (witness A) — `get_sca_checks` accepted `check_id` and never put it in the executed query, so a single-check question returned a `limit`-truncated policy-wide sample that usually did not contain the check asked about.
- `wazuh/wazuh-dashboard#1528` — with privacy mode ON the cluster node name reached the provider in clear while its field name was pseudonymized; the `wazuh.cluster.*` wildcard swept a machine hostname into the same `allow` as the cluster's configured label.
- `wazuh/wazuh-dashboard#1530` — a provider connection dropped mid-answer was finalized as a complete answer: partial text committed, `done` emitted, no `error`, nothing telling the reader the answer stops mid-thought.
- `wazuh/wazuh-dashboard#1531` — `get_rules` had no parameter that could match a rule UUID, so a UUID passed as `name` matched the human title, returned 0 rows, and the assistant reported the rule does not exist. The UUID was also missing from the projected columns.
- `wazuh/wazuh-dashboard#1533` (second half) — the composer was not re-focused when a streaming turn ended, so the caret was lost on every turn.

### Deferred

- `wazuh/wazuh-dashboard#1524` — needs a design decision on how free-text rule titles are treated under privacy mode; not a mechanical fix.
- `wazuh/wazuh-dashboard#1525`, `#1527`, `#1529`, `#1532` — answer-shaping / digest-schema behaviour, out of scope for this pass.
- `wazuh/wazuh-dashboard#1526` witness B (follow-up field names rejected as invalid) — separate field-catalog work.
- `wazuh/wazuh-dashboard#1533` first half (composer disabled while an answer streams) — allowing composition during a stream is a product decision about queued turns, not a focus bug.

## Tasks

- [x] Narrow the `get_sca_checks` query with `check_id`
- [x] Pseudonymize `wazuh.cluster.node` under privacy mode, keep `wazuh.cluster.name` allowed
- [x] Surface a mid-answer stream drop as an error beside the partial text
- [x] Let `get_rules` look up a rule by UUID and project the UUID
- [x] Restore composer focus when a streaming turn ends

MiguelCasaresRobles:
> Stacked on #9043 - merge #9043 first; this PR auto-retargets to 5.0.0 when it merges.

## Description

Closes #9044
Closes wazuh/wazuh-dashboard#1528
Closes wazuh/wazuh-dashboard#1530
Closes wazuh/wazuh-dashboard#1531
Closes wazuh/wazuh-dashboard#1529
Closes wazuh/wazuh-dashboard#1524
Closes wazuh/wazuh-dashboard#1527

Fixes the mechanically-fixable subset of the team review findings (`wazuh/wazuh-dashboard#1523`), all inside `plugins/wazuh-ai-assistant` (`server/` plus one `public/` component). One commit per finding, so any single fix can be dropped without touching the rest.

| Finding | Fix |
| --- | --- |
| `wazuh/wazuh-dashboard#1531` | `get_rules` gains an `id` parameter that matches the rule UUID (`ids` OR a `term` on the business-level id field, so it matches whichever one the value really is), projects the UUID in `_source`/`rowFields`/digest sample columns so an answer can cite it, and skips the `enabled` default for an id lookup so a disabled rule is not reported as non-existent. An explicit `enabled` alongside `id` is still honored. |
| `wazuh/wazuh-dashboard#1526` (witness A) | `get_sca_checks` now puts `check_id` into `query.bool.filter` as an exact `check.id` term, so a single-check question returns that check instead of a `limit`-truncated policy-wide sample. In `filter`, not `post_filter`, so the aggregations are computed inside the same narrowed scope. |
| `wazuh/wazuh-dashboard#1528` | `wazuh.cluster.node` is now an explicit `anonymize`/`HOST` field-policy entry, ahead of the `wazuh.cluster.*` wildcard that had swept it into `allow` together with `wazuh.cluster.name` (the configured cluster label, which stays `allow`). The shape-based prescan could not compensate: `FQDN_TOKEN_RE` needs a dot, so a bare single-word node name was never minted. |
| `wazuh/wazuh-dashboard#1530` | The OpenAI-compatible adapter's loop-end tail now distinguishes a stream that ENDED from one that was CUT: if answer text was streamed and no choice ever carried a `finish_reason`, it emits an `error` beside the partial text instead of a bare `done`. Deliberately narrow, see Evidence. |
| `wazuh/wazuh-dashboard#1529` | The outbound FQDN prescan's field-path guard now recognizes a field NAME by matching the whole path against the generated Wazuh Common Schema catalog (`common/field-catalog.ts`, every indexed WCS path), not only by the segment words the ~10 curated policy fields happen to ship. `related.hosts`, `related.ip`, `related.user` and `wazuh.cluster.node` were being rewritten to `HOST_n` inside a digest's `columns` hint, handing the model a table whose column names it could not read, while the sibling `wazuh.cluster.name` survived. Matched as whole paths, deliberately, so no hostname VALUE is spared - see Evidence. |
| `wazuh/wazuh-dashboard#1524` | The digest prose pre-mint pass now also mints the identities an AGGREGATION carries: a bucket row's `key` (a value OF the aggregated field, which resolving by the literal name "key" never matched) and the `breakdown` buckets, which `applyFieldPolicy` only scrubbed after the samples loop had already sent the prose field through. The reported leak is a top-agents bucket whose row carries both the agent name as a key and that same hostname embedded verbatim in `wazuh.rule.title`. Still known-roster scrubbing: only values a premintable identity field in the SAME payload was already going to pseudonymize enter the dictionary. |
| `wazuh/wazuh-dashboard#1527` | A turn whose round already streamed a complete answer to the screen and was then extended by the deferred-offer interception no longer gets the "Now answer the user's question directly" order on its final round; it gets a continuation instruction that forbids a second version or an "as noted above" recap. The anti-fabrication grounding clauses are shared verbatim between the two instructions, so #8893's guarantee is unchanged. Latched only at the interception site - a round that wrote partial text and then called a tool has not answered the user and still owes them the whole answer. |
| `wazuh/wazuh-dashboard#1533` (second half) | The composer's re-focus effect now depends on the same `disabled \|\| isGenerating` value the rendered `disabled` attribute uses, so it re-runs when a stream ends and the field becomes editable again. It previously depended on `disabled` alone and so never re-ran, and disabling a focused element blurs it. |

Deferred, with reasons, in #9044: `wazuh/wazuh-dashboard#1525`, `#1532`, `#1526` witness B, and `#1533`'s first half (composing while an answer streams is a queued-turn product decision, not a focus bug).

## How to Test

Repro steps come from the child issues.

1. **#1531** — ask the assistant "what does rule `ad97a19d-24a5-43c4-a749-1f8f0a9172bc` detect" (any UUID from `wazuh-content-rules*`). Before: the assistant reports the rule does not exist. After: the rule's title/description is returned, and the UUID is citable.
2. **#1526** — ask "how do I remediate failed SCA check 28500" (any check id present in the index). Before: rows for the whole policy, none of them the check asked about. After: that check's row.
3. **#1528** — with privacy mode ON, ask anything that pulls a metrics/cluster document, and capture the outbound provider request. Before: the real node name in clear under a `HOST_n` field name. After: the value is a `HOST_n` pseudonym; `wazuh.cluster.name` is unchanged.
4. **#1530** — point the assistant at an `openai_compatible` fault-injection endpoint that streams a few content deltas and then closes the socket with no `[DONE]`, no usage frame and no `finish_reason`. Before: the partial answer is committed and rendered as final. After: the partial text stays and an error appears beside it. Check the negative too: a normal provider answer still ends cleanly with no error.
5. **#1529** — with privacy mode ON, ask a question whose digest `columns` hint lists non-curated fields (any `search_wazuh_data` question over `wazuh-alerts*` that reaches `related.*`), and capture the outbound provider request. Before: `"columns": ["HOST_1","HOST_2","HOST_3","wazuh.agent.name","HOST_4"]`. After: every field name is readable. Check the negative in the same capture: a real dotted hostname in a VALUE position is still a `HOST_n`.
6. **#1524** — with privacy mode ON, ask "which agents triggered the most alerts and what were the rule titles", on an estate whose rule titles embed the agent hostname (the vault/credential rules do). Before: the agent name is a `HOST_n` in the bucket key and verbatim in the adjacent `wazuh.rule.title`. After: both are the same `HOST_n`, and the persisted digest carries no real name.
7. **#1527** — ask a question that makes the assistant answer and then offer a follow-up it can run itself ("give me the three failed hardening checks that matter most for agent 001"). Before: the answer is delivered, then a second complete answer follows, opening "As noted above…". After: one answer, optionally extended by a short paragraph adding only what the extra tool call found.
8. **#1533** — send a question, wait for the answer to finish, and type without clicking. Before: the keystrokes are lost. After: the caret is already in the message box.

## Results and Evidence

Gate run against `plugins/wazuh-ai-assistant` inside the `osd-dev` container: `yarn test:jest --runInBand` **129/129 suites, 2496/2496 tests passed** (statements 89.15%, branches 84.21%), `yarn lint` **clean**. Re-run independently of the implementation pass, on the pushed head. Per-fix verification notes:

- **#1529 value-position review (the risk this fix carries).** The guard is consulted from one place, the outbound FQDN prescan, which sees plain text and cannot tell a field-name position from a value position - so widening it at all is a value-scan risk, and it was measured before shipping. Flattening the catalog's 5388 paths into the guard's ALL-segments word vocabulary would have added ~550 segment words, and those words include ordinary hostname material (`host`, `hostname`, `server`, `home`, `dns`, `data`, `cloud`, `network`) and the real TLD `io`: dotted hostname VALUES such as `server.home` (a router-assigned local name) or `data.io` would then have had every segment "known" and would have stopped being pseudonymized. The catalog is therefore consulted as WHOLE PATHS, which admits exactly the field names that exist and nothing else; every field name in the report is an exact catalog path, so coverage is identical. Both halves are pinned by tests: the reported names survive, and `server.home` / `data.io` / `dns.cloud.host` still mint.
- **#1524 residual (stated, not closed by this PR).** This stays known-roster scrubbing with no new matching heuristic: only values a premintable identity field in the same payload was already going to pseudonymize (same field, same kind) enter the dictionary, and only exact, whole-token occurrences of them are replaced - so a common word that is not a real minted identity can never match, pinned by a test asserting that a hostname the payload never carries is left verbatim. The residual is unchanged from before: an identifier that is both shapeless (no dot, so the prescan cannot see it) AND never minted anywhere in the payload still passes.
- **#1527 latch scope.** The latch is set at exactly one site - the deferred-offer interception's `offeredTool` branch, the only mechanism that grants a further provider call after a round produced final text and made no tool call. It is deliberately not set on the ordinary `sawToolCall` path (that round wrote partial narration, the user has not been answered, and the final round still owes them the whole answer) nor on the widening retry, which is a tool path that streams no final text. The existing #8893 pin in `chat-tool-chaining.test.ts` drives precisely this interception turn; it was retargeted at the continuation instruction, with an added assertion that the shared grounding clauses are still present.
- **#1530 negative-case review (the risk this fix carries).** The guard fires only when `sawContent && !sawFinishReason && !toolCallsFinalized`. `[DONE]` and the terminal usage frame both `return` from inside the loop, so the tail is only reached when neither arrived. A provider that omits `[DONE]` but sends any `finish_reason` (`stop`/`length`/`tool_calls`/`content_filter`) is still treated as complete. A drop before the first content delta still falls through to `done` with no text, which is what keeps `chat.ts`'s `synthesizeNoTextFallback` no-tools retry working. A round that produced a tool call is excluded, since its answer arrives on a later round. The `finish_reason` recording sits after the only `continue` in the loop (malformed-JSON frames), so no well-formed frame can bypass it. Only the OpenAI-compatible adapter is touched, so the other adapters' terminal behaviour is unchanged.
- **#1528 non-privacy mode.** The change is a field-policy entry, which is only consulted on the privacy-mode scrub path; with privacy mode OFF nothing is consulted and no value changes. Both exact entries precede the wildcard, which `resolveFieldEntry` requires (first match wins), and the wildcard is kept so every other member of the object keeps its previous treatment.
- **#1526 no-`check_id` path.** The term is added through `...(checkId ? [term] : [])`, so with `check_id` absent the request is byte-identical to before. `get_sca_checks` builds a single request shape (`policy_id` is required), so there is no second branch to keep in sync.
- **#1531 guardrails.** Both new clauses are exact-match (`ids` / `term`) and `isExactIdLookupQuery` already accepts a `bool.should` of `term`/`ids` clauses, so no guardrail change was needed. The newly projected `document.id` is a vendor-minted ruleset identifier and is reviewed `allow` in the field policy.

UI-visible change: #1533 alters composer focus behaviour, which has no still-frame representation (the caret position after a turn completes) — flagging that the template's screenshot expectation needs a short capture from a reviewer with the environment running if one is required.


Ripdiegozz (dismissed):
🟢 Tests

## PR #9045 — manual testing and test review

### 1. Rule by UUID — #1531

> what does rule 28747618-8aa0-460d-81d1-48b06ea97929 detect

**Expected:** the rule's title, *Malicious connection detected - {{source.address}}
to {{destination.address}}*, its description, and the UUID cited back in the
answer.

Image

```json
{
  "document.metadata.title": "Malicious connection detected - {{source.address}} to {{destination.address}}",
  "document.level": "medium",
  "document.status": "stable",
  "document.enabled": true,
  "document.mitre.technique.id": [
    "T1071",
    "T1095"
  ],
  "space.name": "standard",
  "document.tags": [
    "attack.command-and-control",
    "attack.t1071",
    "attack.t1095"
  ],
  "document.id": "28747618-8aa0-460d-81d1-48b06ea97929",
  "document.logsource.product": "wazuh-ioc",
  "document.metadata.description": "Detects a network connection whose source IP and port match a known malicious indicator in the threat intelligence IoC database."
}
```
### 2. Disabled rule by UUID — #1531, second half

Flip the rule first:

```bash
curl -sk -u admin:admin -X POST \
  "https://localhost:9200/wazuh-threatintel-rules-b/_update/28747618-8aa0-460d-81d1-48b06ea97929?refresh=true" \
  -H "Content-Type: application/json" -d '{"doc":{"document":{"enabled":false}}}'
```

> what does rule 28747618-8aa0-460d-81d1-48b06ea97929 detect

**Expected:** the same answer as step 1. An id lookup ignores the `enabled`
default.

Image

```json
{
  "document.metadata.title": "Malicious connection detected - {{source.address}} to {{destination.address}}",
  "document.level": "medium",
  "document.status": "stable",
  "document.enabled": false,
  "document.mitre.technique.id": [
    "T1071",
    "T1095"
  ],
  "space.name": "standard",
  "document.tags": [
    "attack.command-and-control",
    "attack.t1071",
    "attack.t1095"
  ],
  "document.id": "28747618-8aa0-460d-81d1-48b06ea97929",
  "document.logsource.product": "wazuh-ioc",
  "document.metadata.description": "Detects a network connection whose source IP and port match a known malicious indicator in the threat intelligence IoC database."
}
```

Restore it with the same call and `"enabled":true`.

### 3. Single SCA check — #1526

> how do I remediate failed SCA check 35500

**Expected:** that one check, *Ensure mounting of cramfs filesystems is
disabled.*, with its remediation. Any count in the answer describes that check,
not the policy.

Image

```json
{
  "check.id": "35500",
  "check.name": "Ensure mounting of cramfs filesystems is disabled.",
  "check.result": "Not applicable",
  "check.rationale": "Removing support for unneeded filesystem types reduces the local attack surface of the system. If this filesystem type is not needed, disable it.",
  "check.remediation": "Run the following script to unload and disable the cramfs module: - IF - the cramfs kernel module is available in ANY installed kernel: - Create a file ending in .conf with install cramfs /bin/false in the /etc/modprobe.d/ directory - Create a file ending in .conf with blacklist cramfs in the /etc/modprobe.d/ directory - Run modprobe -r cramfs 2>/dev/null; rmmod cramfs 2>/dev/null to remove cramfs from the kernel - IF - the cramfs kernel module is not available on the system, or pre-compiled into the kernel, no remediation is necessary #!/usr/bin/env bash.",
  "check.rules": [
    "[\"c:modprobe -n -v cramfs -> r:install /bin/false|install /bin/true|Module cramfs not found\"",
    "\"not c:lsmod -> r:cramfs\"",
    "\"d:/etc/modprobe.d/ -> r:.*\\\\.conf -> r:^blacklist\\\\s+cramfs\"]"
  ]
}
```

Image

```json
{
  "wazuh.agent.name": "wazuh.agent.deb.local",
  "checksum.hash.sha1": "1733c9623e1cde5b8c66d4d7101f1600654a6d9e",
  "wazuh.cluster.name": "wazuh",
  "wazuh.agent.host.hostname": "wazuh.agent.deb.local",
  "wazuh.agent.host.os.name": "Ubuntu",
  "wazuh.agent.host.os.type": "linux",
  "wazuh.agent.host.os.version": "24.04.4 LTS (Noble Numbat)",
  "wazuh.agent.host.os.platform": "ubuntu"
}
```

Image

A count that still reflects the whole policy means the filter landed in
`post_filter` instead of `filter`.

### 4. No second answer — #1527


> give me the three failed hardening checks that matter most for agent 001

**Not reproducible from the UI.** Across six attempts the deferred-offer
interception never fired, so the turn shape this fix addresses was never
reached. Verified from the outbound capture: two turns did reach a final round
and both carried `FINAL_ROUND_ANSWER_INSTRUCTION`, never the continuation
variant, so the latch evaluated and correctly stayed off. Covered end to end by
the edited assertion in `chat-tool-chaining.test.ts`.

### 5. Composer focus — #1533

Send any question, wait for the stream to finish, then type without clicking.

**Expected:** the characters land in the field.

Image

### 6. Cluster node pseudonym — #1528

Privacy mode ON. Ask anything that pulls a cluster or metrics document, and
capture the outbound provider request.

**Expected:** `wazuh.cluster.node`'s value is a `HOST_n`, and
`wazuh.cluster.name` is still the readable cluster label.


```json
"samples": [
  {"wazuh.cluster.node":"HOST_6","wazuh.cluster.name":"wazuh","wazuh.agent.name":"HOST_7"},
  {"wazuh.cluster.node":"HOST_6","wazuh.cluster.name":"wazuh","wazuh.agent.name":"HOST_8"}
],
"columns": ["wazuh.cluster.node","wazuh.cluster.name","wazuh.agent.name"]
```

The node value is pseudonymized, the cluster label stays readable, and the
field names survive. `node01` and `wazuh.agent.deb.local` appear zero times
across all six outbound requests of that turn.

The UI shows the real values because they are restored on the way back.

Image

### 7. Field names stay readable — #1529

Privacy mode ON.

> search the index wazuh-events-v5-* from now-24h to now and return only these fields: wazuh.cluster.node, wazuh.cluster.name, wazuh.agent.name

**Expected:** the digest's `columns` hint lists readable field names. In the same
capture, a real dotted hostname in a VALUE position is still a `HOST_n`.

Image

Outbound request, same turn:

```json
"columns": ["wazuh.cluster.node", "wazuh.cluster.name", "wazuh.agent.name"],
"samples": [
  {"wazuh.cluster.node":"HOST_6","wazuh.cluster.name":"wazuh","wazuh.agent.name":"HOST_7"},
  {"wazuh.cluster.node":"HOST_6","wazuh.cluster.name":"wazuh","wazuh.agent.name":"HOST_8"}
]
```

### 8. One pseudonym per host — #1524

Privacy mode ON.

>  which source IPs failed authentication the most and what were the rule titles

**Expected:** the bucket key and the identity embedded in `wazuh.rule.title`
share one pseudonym, two distinct addresses stay distinct, and no real value
reaches the outbound body.

Outbound `breakdown`:

```json
{"key": "HOST_1", "count": 20, "agg": "wazuh_agent_name"},
{"key": "HOST_2", "count": 20, "agg": "wazuh_agent_name"},
{"key": "Failed authentication attempt - root from IP_1", "count": 24, "agg": "wazuh_rule_title"},
{"key": "Multiple SSH authentication failures for root from IP_1", "count": 6, "agg": "wazuh_rule_title"}
```

The samples carry the same `IP_1` inside their `wazuh.rule.title`. `198.51.100.77`
appears zero times across the outbound body.

Image

Image

rodrigofez (dismissed):

yenienserrano (approved):
LGTM :green_circle:

PR #8239, opened 2026-03-25
Create an app/section of regulatory compliance and move all the tabs of the related modules

Ripdiegozz:
### Description

This PR implements a centralized **Regulatory Compliance** application within the Wazuh dashboard, consolidating the PCI DSS, GDPR, HIPAA, NIST 800-53, and TSC modules into a single unified interface. All individual compliance applications have been converted into secondary tabs within this new centralized section, employing a 2-level tab navigation pattern consistent with the **IT Hygiene** module.

### Issues Resolved
- **Issue**: https://github.com/wazuh/wazuh-dashboard-plugins/issues/8228

### Evidence

https://github.com/user-attachments/assets/247fdc6b-0c55-4fd3-be8e-a7aea930b625

### Test
1. Navigate to the new **Regulatory Compliance** application in the main menu.
2. Verify that the first level of tabs displays the available frameworks (PCI DSS, GDPR, HIPAA, NIST 800-53, and TSC).
3. Select any framework and verify that the second level of tabs renders correctly with their respective components.
4. Confirm that switching between frameworks updates the sub-tabs and data sources accordingly.
5. Verify that the legacy individual compliance applications no longer appear in the sidebar or overview app.

### Check List
- [ ] All tests pass
  - [ ] `yarn test:jest`
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [X] Update [CHANGELOG.md](./../CHANGELOG.md)
- [X] Commits are signed per the DCO using --signoff

Issue #8228, opened 2026-03-25
Create an app/section of regulatory compliance and move all the tabs of the related modules


Desvelao:
## Description
Create a new **Regulatory Compliance** app and move all tabs from the related modules into this new area. This restructuring will centralize compliance‑related features and improve navigation.

Follow the layout similar to **IT Hygiene**: app with 2 levels of tabs:
- 1st level: regulatory compliance (PCI DSS, etc...)
- 2nd level: tab of module (Dashboard, Findings)

This issue could be blocked or affected by: https://github.com/wazuh/wazuh-dashboard-plugins/issues/8229

## Tasks
- [x] Define the structure and layout of the new Regulatory Compliance section.
- [x] Identify all modules and tabs related to regulatory compliance.
- [x] Move the identified tabs into the new section, ensuring routing and navigation remain functional.
- [x] Validate the new section in the UI.

PR #9049, opened 2026-08-26
fix(regulatory-compliance): fix Controls tab action buttons

### LINKED ISSUE #9038: Regulatory compliance > Controls call to actions not working
# Description

The buttons that are diplayed when the user hovers over a requirement _[Show in dashboard / Inspect in findings]_ are not working. This happens in the main Control tab view, but also in the requirement details flyout.

## Tasks

- [x] Fix the navigation action of the _[Show in dashboard / Inspect in findings]_ buttons
- [x] Fix the button positions when scrolling the requirements

Image

Image

Desvelao:
## Description

Closes #9038

In the **Regulatory compliance > \<Framework\> > Controls** tab, hovering a
requirement tile reveals two action icons — *Show in dashboard* and *Inspect in
findings* — in both the tile itself and the requirement details flyout. Neither
worked, and the icons drifted from their tile once the requirements list was
scrolled.

## Proposed Changes

- Fixed the navigation action: the click handlers called
  `props.onSelectedTabChanged(...)`, a prop that only exists under the legacy
  tab system. The Controls tab is rendered through the newer
  `ComplianceModule` / URL-search-param sub-tab navigation, which never passes
  that prop, so the call threw and the tab never switched. Both actions now
  navigate via `NavigationService.updateAndNavigateSearchParams({ tabSubView })`,
  matching how sub-tab switching actually works today.
- Fixed the filter these actions add: it was built against
  `AppState.getCurrentPattern()`, a hardcoded legacy index pattern
  (`wazuh-events-v5*`) unrelated to the compliance framework's own data source.
  It now uses the Controls tab's actual `indexPatternId`, so the filter applies
  correctly against the Dashboard/Events tab it navigates to.
- Fixed the button positions when scrolling: the action icons were
  `position: fixed`, which pins them to the viewport instead of the
  (scrollable) requirements grid. They're now rendered in normal document flow
  (`display: inline-flex`) inside the tile's existing `.wz-mitre-width`
  reserved space, so they scroll together with their requirement tile and sit
  next to the quantity badge instead of overlapping it.
- Fixed a residual scroll-detach issue with the icons' own tooltips: each
  action icon opens its own `EuiToolTip`, which EUI renders through a React
  portal positioned once in viewport coordinates — that portal only listens
  for `window` resize, not for scrolling of the requirements grid's own
  internal scroll container, so an open tooltip bubble could visually lag
  behind its icon on scroll. The requirements grid now clears the hovered
  tile's state (closing its icons and any open tooltip) on `scroll`, the
  standard "close on scroll" pattern, rather than fighting the tooltip
  library's positioning.

### Artifacts Affected

- `plugins/main` (public)

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `plugins/main/public/components/overview/compliance-table/components/subrequirements/subrequirements.test.tsx`:
  - `openDashboard`/`openDiscover` navigate via
    `NavigationService.updateAndNavigateSearchParams` with the correct
    `tabSubView`, and the filter they build targets the tab's own
    `indexPatternId`.
  - Scrolling the requirements grid clears the hovered tile's state (closing
    its icons and any open tooltip), and doesn't throw when nothing is
    hovered.

### Note on CHANGELOG.md

No CHANGELOG entry was added for #9038: the navigation bug is a same-version
(5.0.0, unreleased) regression, traced to PR #8239 / issue #8228 (the
regulatory-compliance consolidation), and never shipped in a released version.
Instead, `CHANGELOG.md`'s existing entry for that consolidation work
(previously linking PRs #8239/#8303) now links its underlying issue (#8228)
instead, per this repo's "link the issue, not the PR" convention, and #9038
is tracked as a GitHub sub-issue of #8228 to preserve the causal relationship.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


rodrigofez (dismissed):
🟢 Test

https://github.com/user-attachments/assets/3bd227e3-4502-44e1-af04-141066fb0b6a


Note: maybe out of the scope, but I think it would be better if that when switching from a compliance framework to another one the framework specific filters are not persisted

asteriscos (dismissed):
CR: 🟢 
Test: 🟢 

https://github.com/user-attachments/assets/7dceecc6-573a-4c5b-92da-3107e23a740d

asteriscos (approved):

Issue #8956, opened 2026-08-12
Deprecate `/daemons/stats` endpoint and `node_name` property


asteriscos:
## Description

The Wazuh Server is deprecating the endpoint `GET /agents/{agent_id}/daemons/stats`
and the `node_name` property across the agents endpoints (see
wazuh/wazuh#38351). In Wazuh 5.x, agents communicate over HTTPS through a load
balancer instead of a persistent TCP connection to a specific manager node, so
consecutive requests from the same agent can be handled by different cluster
nodes. Persistent node affinity is no longer a valid concept, which makes both
the `node_name` property and the per-node `daemons/stats` endpoint obsolete:

- `GET /agents/{agent_id}/daemons/stats` relies on per-node in-memory
  connection data that cannot function correctly for load-balanced agents.
- `node_name` is being removed from the agent table schema, `agent_info_data`,
  API filters, and response objects on the server side.

This dashboard needs to stop relying on both so it doesn't break once the
server side removes them: drop the `node_name` column/field/filter from the
agents table, agent info panel, CSV export, and search bar; and remove the
`daemons/stats` entries from the API reference / RBAC metadata that document
it.

### Code references

**`node_name` usages**

- https://github.com/wazuh/wazuh-dashboard-plugins/blob/8d865cf75f24e4bbe74148ef3fac8ac3f7100630/plugins/main/public/components/endpoints-summary/types.ts#L24 — `node_name` field in the `Agent` interface
- https://github.com/wazuh/wazuh-dashboard-plugins/blob/8d865cf75f24e4bbe74148ef3fac8ac3f7100630/plugins/main/public/components/endpoints-summary/table/columns.tsx#L73 — `node_name` column ("Cluster node") in the agents table
- https://github.com/wazuh/wazuh-dashboard-plugins/blob/8d865cf75f24e4bbe74148ef3fac8ac3f7100630/plugins/main/public/components/endpoints-summary/table/agents-table.tsx#L271-L273 — row mapping of `item.node_name`
- https://github.com/wazuh/wazuh-dashboard-plugins/blob/8d865cf75f24e4bbe74148ef3fac8ac3f7100630/plugins/main/public/components/endpoints-summary/table/agents-table.tsx#L312 — search-bar suggestion for `node_name` ("filter by cluster name")
[...]

PR #9054, opened 2026-08-27
Remove per-node agent count from Server Status

### LINKED ISSUE #9052: Per-node agent count in Server management > Status is always 0 after the `node_name` removal
| Wazuh | Rev | Browser |
|-------|-----|---------|
| 5.0.0 | `5.0.0-latest` public ECR `wazuh-cicd` images | Browser independent |

## Descriptiom 

`Server management > Status` reports `Agents 0` on every cluster node, on every 5.0.0 deployment, including single-node ones.

Image

The server removed the agent-to-node affinity data: `node_name` is gone from `GET /agents` (as a returned field, as a `select` field and as a filter) and `GET /overview/agents` no longer returns the `nodes` breakdown. That removal is deliberate and complete, spec included, and it follows from 5.x agents talking HTTPS through a load balancer instead of holding a persistent TCP connection to one manager node. The dashboard was cleaned up for it in #8958, but the per-node agent counter in the Status page was explicitly left out of that pass and is now the last consumer of a field that no longer exists, so its lookup always misses and the panel falls back to `0`.

This is the follow-up issue that #8958 asked for. Quoting its own "Out of scope" section:

> `management/status/status-node-info.js` computes the agent count per node from `stats.agentsCountByNode`, fed by the `nodes` array of `GET /overview/agents`, which the server derives by grouping agents by `node_name`. If the server drops that grouping, Server Management > Status will show `0` agents for the selected node. It does not throw and it was not listed in the issue, so it is left untouched; it is the last remaining dependency on agent-to-node affinity and deserves a follow-up issue.

The server has since dropped that grouping, so the predicted `0` is now the observed behavior.

## Preconditions

1. Any Wazuh 5.0.0 deployment. A single-node install is enough, see the note on `clusterEnabled` below.
2. At least one agent registered, so the cluster-wide totals on the same screen are non-zero and the discrepancy is visible.

## Steps to reproduce

1. Navigate to `Server management > Status` (application `server-status`, URL `/app/server-status`, which redirects to `/manager/?tab=status`).
2. Read the summary cards at the top of the page: `Nodes`, `Agents`, `Total agents`, `Active`, `Agents coverage`.
3. Read the `Agents` row of the `<node name> information` panel below them.
4. On a cluster, switch the node selector between each node and read that row again.

### Expected Result

1. The per-node figure is consistent with the cluster totals shown directly above it, or the row is removed if the figure is no longer meaningful.

## Actual Result

1. The `Agents` row of the `<node name> information` panel reads `0` for every node, while the summary cards on the same screen report the correct totals.

Observed on a two-node cluster with seven registered agents, four of them active:

```
nodes in the selector: ['manager', 'worker01']
  selected node manager    panel 'manager information'  -> Agents=0   (cluster: total=7 active=4)
  selected node worker01   panel 'worker01 information' -> Agents=0   (cluster: total=7 active=4)
```

The same screen reports `Nodes 2`, `Agents 7`, `Total agents 7`, `Active 4`, `Agents coverage 57.14%` and names the last registered agent correctly. Only the per-node figure is wrong, and it is wrong identically on both nodes.

### Where the number comes from

`plugins/main/public/controllers/management/components/management/status/status-overview.js:107` feeds the store straight from the overview response:

```js
agentsCountByNode: agentsOverview?.nodes,
```

`plugins/main/public/controllers/management/components/management/status/status-node-info.js:35-38` looks the selected node up by `node_name`:

```js
const agentsCountByNode = stats?.agentsCountByNode || [];
const agentsNodeCount = clusterEnabled
  ? (agentsCountByNode.find(node => node.node_name === selectedNode) || {})
      .count || 0
  : stats?.agentsCount?.total || 0;
```

Both halves of the cluster branch are unsatisfiable now: `agentsOverview.nodes` is `undefined`, so `agentsCountByNode` is `[]`, so `.find()` returns `undefined`, so the `|| 0` fires.

### The `clusterEnabled` guard does not limit this to clusters

`clusterEnabled` is initialised to `true` in `plugins/main/public/redux/reducers/statusReducers.js:21` and nothing ever writes it. The reducer case that would update it exists at line 67, and `UPDATE_CLUSTER_ENABLED` appears nowhere else in the repository: a GitHub code search for it returns exactly one hit, the reducer that consumes it, and a search for a matching `updateClusterEnabled` action creator returns none.

So the cluster branch is taken unconditionally and the non-cluster fallback is dead code. That means the per-node agent count reads `0` on single-node deployments too, which makes this cheap to reproduce and cheap to assert in a test. It is also consistent with 5.0.0 making the cluster unconditional: the shipped configuration has no `<disabled>` element, where 4.14 ships `<disabled>yes</disabled>`, and `GET /cluster/status` answers `{"running":"yes"}` on a stock install.

### Server side, for reference

Verified on `wazuh` `origin/5.0.0` at `9b0a0326e5`:

- `framework/wazuh/agent.py:1413` builds the `get_full_overview` result without a `nodes` key. The 4.14 equivalent (line 1573) has `'nodes': stats_distinct_node`, populated by `get_distinct_agents(fields=['node_name'])`.
- `api/api/controllers/agent_controller.py` contains no occurrence of `node_name`. `api/api/spec/spec.yaml` keeps six, all of them the cluster node's own name, none of them a parameter or response field of an agents endpoint.
- At `1751607b46` (2026-08-07) the same files had 15 and 30 occurrences respectively, so the server-side removal landed after that commit.

Consequently the API rejects the old contract cleanly rather than ignoring it:

```bash
curl -sk -H "Authorization: Bearer $T" 'https://localhost:55000/overview/agents' \
  | python3 -c 'import sys,json; d=json.load(sys.stdin)["data"]; print("nodes" in d, sorted(d))'
# -> False ['agent_os', 'agent_status', 'agent_version', 'groups', 'last_registered_agent']

curl -sk -H "Authorization: Bearer $T" 'https://localhost:55000/agents?select=id,name,node_name'
# -> 400 Not a valid select field

curl -sk -H "Authorization: Bearer $T" 'https://localhost:55000/agents?node_name=worker01'
# -> 400 Extra query parameter(s) node_name not in spec
```

### Restoring the `nodes` breakdown is not the fix

It would reintroduce a value that no longer has a defined meaning. Per #8956 and wazuh/wazuh#38351, in 5.x consecutive requests from the same agent can be served by different cluster nodes, so there is no node an agent belongs to. Any restored per-node count would be an artefact of whichever node happened to answer last.

The question that remains answerable is the retrospective one, "which node processed this agent's data", and the indexer already answers it for events and findings. Verified on this cluster:

```
findings: agent 007 -> worker01=32    agent 002 -> manager=1193
events:   agent 007 -> worker01=57    agent 002 -> manager=9381
```

So the correct resolution here is removal on the dashboard side, not restoration on the server side.

Desvelao:
## Description

Closes #9052

This is the dashboard-side follow-up requested by #8956: the server has
now dropped agent-to-node affinity data entirely (`node_name` is no
longer returned by `GET /agents` or `GET /overview/agents`), which was
predicted but explicitly left out of scope there. As a result, the
per-node `Agents` figure in `Server management > Status` always read
`0` on every node, on every 5.0.0 deployment (cluster or single-node),
while the summary cards on the same screen reported the correct
totals.

Restoring a per-node breakdown server-side is not the fix: in 5.x an
agent's requests can be served by different cluster nodes, so "the
node this agent belongs to" no longer has a stable meaning. This PR
removes the now-meaningless row instead.

No new `CHANGELOG.md` entry is added — #8956's existing entry already
covers this removal, so this PR carries the `no-changelog` label.

## Proposed Changes

- `status-node-info.js`: remove the per-node `Agents` row and the
  `agentsCountByNode`/`clusterEnabled` lookup that always resolved to
  `0`.
- `status-overview.js` / `actions-buttons-main.js`: stop requesting the
  now-nonexistent `nodes` breakdown from the overview response.
- `redux/reducers/statusReducers.js`: remove the dead `clusterEnabled`
  state and its `UPDATE_CLUSTER_ENABLED` reducer case (initialized to
  `true`, never dispatched anywhere in the codebase).

### Results and Evidence

image

### Tests Introduced

- Added `status-node-info.test.tsx`: renders `WzStatusNodeInfo` with a
  selected node and node info, and asserts the panel no longer renders
  an `Agents` row while the other fields still render correctly.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...


rodrigofez (dismissed):
🟢 CR
🟢 Test

Screenshot 2026-08-27 at 07 40 01

rodrigofez (dismissed):

yenienserrano (approved):
LGTM :green_circle:

asteriscos (approved):
CR: 🟢 
Test: 🟢 

Image

PR #9059, opened 2026-08-27
Remove connection information from the GitHub and Office 365 UI

### LINKED ISSUE #9056: Remove connection information from the GitHub and Office365 UI
# Description

We need to remove the client information from the GitHub and Office365 flyout.

## Tasks

- [x] Remove the client ID and client secret from the Office365 flyout
  - Leave only the tenand ID at the root level
- [x] Remove the token property from the GitHub flyout
  - Leave only the organization property at the root level

#### Office365

Image

#### GitHub

Image

Ripdiegozz:
## Description

The Office 365 and GitHub module configuration panels rendered the credentials the
agent reports for each `api_auth` entry, so anyone who could open the module flyout
could read the connection secrets straight off the screen. Displaying them buys the
operator nothing: to tell one credential block from another, the identification
fields are enough.

This removes the connection information from both panels, keeping only the
non-sensitive fields that identify each entry.

Closes #9056

## Proposed Changes

- **Office 365** — the `api_auth` panel now renders only `Tenant ID`, as the issue
  asks. Removed `Client ID`, `Client secret`, `Path file of client secret`, and
  `API type`. The secret path goes because it discloses where the secret lives on
  disk; `API type` is not client information, but keeping it would contradict
  "leave only the tenant ID at the root level".
- **GitHub** — the `api_auth` panel now renders only `Organization`. Removed `Token`.
- **Entries with no displayable field** — both panels filter out fields whose value
  is `undefined`. With fewer fields retained, an entry can now filter down to
  nothing (an Office 365 entry configured with client credentials but no
  `tenant_id`). Rather than rendering a blank panel or silently dropping the entry,
  the panel keeps its place and shows `No identification fields configured`, so a
  misconfigured credential block stays visible to the operator.
- **React keys** — the panel key is now derived from the map index instead of the
  field values. This also fixes a pre-existing defect: the GitHub key interpolated
  `v.client_id`, a field GitHub `api_auth` never provides, so it resolved to
  `undefined` for every entry.

Both panels feed the same shared `PanelModuleConfiguration` renderer, so the change
covers the agent flyout and the manager/cluster node view without a second code path.

This is a display-layer change. No server-side or indexer-side filtering was added —
the values still travel in the API response, and the raw XML/JSON configuration
editor under Management > Configuration still shows them. That surface is out of
scope here and is worth its own issue.

### Results and Evidence

image

image

Full `plugins/main` unit test suite:

```
Test Suites: 9 skipped, 230 passed, 230 of 239 total
Tests:       84 skipped, 1745 passed, 1829 total
Snapshots:   67 passed, 67 total
```

### Artifacts Affected

None. The change is confined to two React view modules in the `wazuh` plugin and
their colocated tests.

### Configuration Changes

None. No configuration parameter is added, removed, or changed — only which of the
already-reported fields are displayed.

### Documentation Updates

None required.

### Tests Introduced

The colocated tests for both panels were rewritten rather than trimmed. They now
assert the **absence** of the removed values and their labels, alongside the
presence of the retained ones:

- Office 365: single `api_auth` object and array of entries, an entry using
  `client_secret_path` instead of `client_secret`, an entry carrying `api_type`
  (asserted absent), and an entry with no retained field (placeholder case).
- GitHub: single and multiple entries, plus the placeholder case.
- The non-array `api_auth` case was kept and rewritten so the single-object path
  through `toApiAuthEntries` stays covered.
- `module_configuration.test.tsx` runs unchanged as a regression check.

## Review Checklist

- [x] Code changes reviewed


asteriscos (approved):
### Office365 🟢 

Image

### Github 🟢 

Image

PR #9062, opened 2026-08-27
Adapt the dashboard to the unified agent endpoint

### LINKED ISSUE #9053: Adapt the dashboard to the unified agent `<endpoint>` and remoted `global_prefix`
## Description

Two coordinated Wazuh core issues change how the agent–manager connection target is configured in 5.0.0, and both surface in the dashboard:

- **Agent side** — [wazuh/wazuh#38624](https://github.com/wazuh/wazuh/issues/38624) collapses the agent's `<manager>` block into a single `<endpoint>` value carrying the full connection target:

  ```xml
  <manager>
    <endpoint>192.168.0.60:1517/wazuh-manager/</endpoint>
  </manager>
  ```

  `<address>` and `<port>` are removed from `Read_Agent_Manager()` (along with `<interface_index>`/`network_interface`), and omitted components fall back to defaults — port `1517`, path prefix `/wazuh-manager/`.

- **Manager side** — [wazuh/wazuh#38650](https://github.com/wazuh/wazuh/issues/38650) aligns remoted's `<remote><https><global_prefix>` with that grammar: it gives remoted a built-in default of `/wazuh-manager/`, makes `w_remoted_validate_global_prefix()` accept exactly the path grammar #38624 settles, and — the item that reaches us — **reports `remote.https.global_prefix` through `GET /manager/configuration`**, which it does not today.

Both are unreleased-only changes: 5.0.0 has shipped no `<address>`/`<port>` configuration, so they are straight replacements rather than deprecations, and there is no transition period for the dashboard to lean on.

The dashboard does not parse `ossec.conf`, so the impact is limited to the places where we **display** the reported configuration and where we **generate** the deployment command that produces it. Since #9050 the agent configuration is read from the `wazuh-agent-config*` index rather than from the Server API, so what matters on the agent side is the shape the agent reports to the indexer — the 5.0.0 API spec models no agent configuration payload and gives no early warning there.

### Impact

**1. Agent Configuration → Communication breaks silently.** The "Server settings" table renders exactly two columns, `address` and `port`, over `clientConfig.manager`. Once the agent reports `endpoint` instead, both columns resolve to empty/default values and the view presents a manager configuration the user never wrote, without raising an error.

Two defects already present in that same code that this is the natural moment to fix:

- The port column defaults to **`1514`**, which is stale today — the 5.0.0 HTTPS transport uses **1517**, before #38624 even lands.
- `asArray()` supports a list of managers, but 5.0.0 `<manager>` is single-entry and #38624 explicitly keeps it that way (multiple `<endpoint>` entries are a stated non-goal). The table-of-managers shape can collapse into a settings group.

The file has no colocated test, so nothing in CI will catch the regression.

**2. Server Management → Configuration → Remote settings does not show the prefix.** The Remote settings table still renders the pre-HTTPS remoted shape — `connection`, `port`, `protocol` (defaulting to `udp`), `ipv6`, `allowed-ips`, `denied-ips`, `local_ip`, `queue_size` — with no notion of the `<https>` sub-block. Once #38650 reports `remote.https.global_prefix`, that value should be surfaced: a prefix mismatch between manager and agent is precisely the failure mode #38650 describes (the manager serves `POST /stateless` while every agent requests `POST /wazuh-manager/stateless`, so every request `404`s), and the dashboard's configuration view is where an operator would look to diagnose it. Today it cannot be read back from the UI at all.

**3. Deploy new agent — blocked on an unresolved upstream decision.** The wizard emits `WAZUH_MANAGER='<value>'` from a `serverAddress` field validated by `SettingsValidator.serverAddressHostnameFQDNIPv4IPv6`. #38624 requires the install-time generators (`register_configure_agent.sh`, `inst-functions.sh`, `InstallerScripts.vbs`) to emit the new grammar, but does not state what happens to the `WAZUH_MANAGER` environment variable:

- If `WAZUH_MANAGER` keeps accepting a bare host and the scripts synthesize the defaults, the wizard needs no change.
- If `WAZUH_MANAGER` starts carrying the full `host[:port][/path]` grammar, the current validator rejects every valid endpoint: it allows only hostname/FQDN/uncompressed IPv6 — no `:port`, no `/path`, no bracketed IPv6, no `%25<ifname>` zone id, and not even compressed IPv6. That also affects the `enrollment.dns` plugin setting, which shares the validator and prefills the field.

This must be confirmed with the agent team before the wizard work is planned.

Once #38650 exposes `global_prefix` through the API, there is also an opportunity worth evaluating: the wizard already fetches the master node's remote configuration, so it could generate a **complete** endpoint (host, port and the manager's actual prefix) instead of a bare host, which removes the mismatch class of failure from the deployment path entirely.

**4. Adjacent cleanup.** The wizard still emits `WAZUH_PROTOCOL`, driven by `remote.isUdp` from the master configuration. `getRemoteConfiguration()` derives both `isUdp` and `haveSecureConnection` from `item.connection === 'secure'` and `item.protocol[]` — all pre-HTTPS concepts. With HTTPS as the only transport the scheme is part of the endpoint and TCP/UDP selection has no effect.

**5. Mocks and documentation.** The imposter agent configuration mocks still return the 4.x shape (`client.server[].address/port`, `protocol: tcp`, port `1514`), already stale against the `agent.agent.manager[]` reader the UI uses today. The one-liner deployment reference documents `WAZUH_MANAGER`, `WAZUH_MANAGER_PORT` and "multiple managers separated by commas" — all three statements go stale under the new grammar.

### Not affected

The `registration-service` (authd) view and the Server API client in `wazuh-core` (dashboard → manager on port 55000). Neither reads the agent's `<manager>` block nor remoted's HTTPS settings.

### Sequencing

#38624 leaves the exact grammar open — it flags `host:port/path` versus `host/path:port` ordering as an open point to settle during implementation, and #38650 exists specifically to agree the shared path grammar between both ends before either lands. Any validator we write has to match that grammar exactly, so this work should start once both issues have settled. The Remote settings item additionally depends on #38650's API change actually shipping.

### Code references

- Communication view, `address`/`port` columns and the stale `1514` default:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/controllers/management/components/management/configuration/client/client.js#L82-L83
- Manager list handling in the same view:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/controllers/management/components/management/configuration/client/client.js#L97
- Reported configuration read from the `wazuh-agent-config*` index:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/controllers/management/components/management/configuration/utils/agent-config-service.ts
- Remote settings table columns, with no `<https>`/`global_prefix` notion:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/controllers/management/components/management/configuration/global-configuration/global-configuration-remote.js#L56-L85
- `getRemoteConfiguration()` reading `/cluster/{node}/configuration/request/remote`:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/components/endpoints-summary/register-agent/services/register-agent-services.tsx#L34-L73
- `serverAddress` → `WAZUH_MANAGER` mapping:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/components/endpoints-summary/register-agent/core/config/os-commands-definitions.ts#L213-L214
- `WAZUH_PROTOCOL` optional parameter:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/components/endpoints-summary/register-agent/core/config/os-commands-definitions.ts#L239
- Wizard `serverAddress` field and its validator:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/main/public/components/endpoints-summary/register-agent/containers/register-agent/register-agent.tsx#L85-L91
- `SettingsValidator.serverAddressHostnameFQDNIPv4IPv6`:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/wazuh-core/common/services/settings-validator.ts#L354
- `enrollment.dns` setting sharing the same validator:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/plugins/wazuh-core/common/constants.ts#L373
- Stale imposter agent configuration mocks:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/docker/imposter/agents/configuration/agent-client.json
- One-liner deployment reference documenting `WAZUH_MANAGER_PORT`:
  https://github.com/wazuh/wazuh-dashboard-plugins/blob/7b95b676a058e74c1d2c907bf60d241ebaebe016/docs/ref/agent-deploy-one-liner.md#L117-L118

## Tasks

- [ ] Confirm with the agent team whether `WAZUH_MANAGER` keeps accepting a bare host or starts carrying the full `host[:port][/path]` grammar, and obtain the final grammar agreed between wazuh/wazuh#38624 and wazuh/wazuh#38650.
- [ ] Rewrite the Communication view to render the single `endpoint` value instead of the `address`/`port` table, dropping the multi-manager shape.
- [ ] Show the effective defaults (port `1517`, prefix `/wazuh-manager/`) when the reported endpoint omits them, replacing the stale `1514` default.
- [ ] Add a colocated test for the Communication view covering the endpoint value, an endpoint with omitted components, and the no-configuration empty state.
- [ ] Surface `remote.https.global_prefix` in the Remote settings view once wazuh/wazuh#38650 reports it, including its effective default.
- [ ] Review the rest of the Remote settings columns against the 5.0.0 `<remote>` shape, which now carries an `<https>` sub-block the view does not model.
- [ ] If `WAZUH_MANAGER` adopts the new grammar, update `SettingsValidator.serverAddressHostnameFQDNIPv4IPv6` (or add a dedicated endpoint validator) to accept the optional port, optional path prefix, bracketed IPv6 with an optional `%25<ifname>` zone id, and compressed IPv6 — with unit tests for each form and for malformed strings.
- [ ] If the validator changes, review the `enrollment.dns` setting and the wizard field label/help text for the wider input it now accepts.
- [ ] Evaluate prefilling the wizard with a complete endpoint built from the master node's reported `global_prefix`, instead of a bare host.
- [ ] Remove the `WAZUH_PROTOCOL` optional parameter and the `remote.isUdp` lookup that drives it, now that HTTPS is the only transport.
- [ ] Update the imposter agent configuration mocks to the shape the agent reports in 5.0.0.
- [ ] Update `docs/ref/agent-deploy-one-liner.md` for the new environment variable grammar.
- [ ] Add the CHANGELOG entry linking this issue.

rodrigofez:
## Description

In 5.0.0 the agent no longer configures its server with `<address>` and `<port>`. It uses a single `<endpoint>` holding the whole connection target, `host[:port][/path]`, where the port defaults to `1517` and the path prefix to `/wazuh-manager/`. The indexer already reports it that way: `manager.address` and `manager.port` were replaced by `manager.endpoint` in wazuh/wazuh-indexer-plugins#1491.

The dashboard still read the old shape. The Communication view rendered an `address`/`port` table that resolves to empty once the agent reports `endpoint`, showing a manager configuration nobody wrote and raising no error. Its port column also defaulted to `1514`, which was already wrong for 5.0.0.

Closes #9053

## Proposed Changes

**Agent configuration > Communication**

- Renders the single `endpoint` the agent reported, replacing the `address`/`port` table.
- Drops the list-of-managers handling. 5.0.0 allows one `<manager>` block only.
- Removes the stale `1514` port default.

**Deploy new agent**

- The server address step now has three fields: address, port and path prefix. They are kept apart so the address keeps its hostname/FQDN/IP validation.
- Port and path prefix are optional. Left empty they are omitted from the command and the agent applies its own defaults, which the field hints state.
- The three are joined into one value and generated as `WAZUH_MANAGER_ENDPOINT`.
- **Remember address, port, and path prefix** now saves all three.
- Removes `WAZUH_PROTOCOL` and the UDP lookup behind it. HTTPS is the only 5.0.0 transport, so the parameter had no effect.

### Results and Evidence
Screenshot 2026-08-27 at 13 29 50
Screenshot 2026-08-27 at 13 36 11
Screenshot 2026-08-27 at 13 36 59

### Artifacts Affected

None.

### Configuration Changes

Two new settings store what the Remember toggle saves:

| Setting | Purpose | Default |
| --- | --- | --- |
| `enrollment.port` | Port used to prefill the wizard | empty, meaning `1517` |
| `enrollment.path` | Path prefix used to prefill the wizard | empty, meaning `/wazuh-manager/` |

`enrollment.dns` is unchanged and keeps its current validation.

### Documentation Updates

- `docs/ref/agent-deploy-one-liner.md`: replaced `WAZUH_MANAGER`, `WAZUH_MANAGER_PORT` and `WAZUH_PROTOCOL` with `WAZUH_MANAGER_ENDPOINT`, and removed the claim that several managers can be given separated by commas. The agent connects to one manager.
- `docs/ref/migration-4x-5x.md`: updated the example install command.

### Tests Introduced

- `client.test.tsx`: the reported endpoint, endpoints that omit the port or the path prefix, and both empty states.
- `agent-endpoint.test.ts`: composing the endpoint from the three fields, including empty components and a missing leading slash.
- `settings-validator.test.ts`: the port and path prefix validators, covering accepted forms, the charset, the length limit and rejected segments.
- Updated the register-agent tests for the new variable and the removal of `WAZUH_PROTOCOL`.

## Review Checklist

- [x] Code changes reviewed


docker/imposter/agents/configuration/agent-client.json
Desvelao, line 1: **thought**: I am not sure if these files are used.

docker/imposter/agents/configuration/default.json
Desvelao, line 1: **thought**: I am not sure if these files are used.

docker/imposter/agents/configuration/agent-client.json
rodrigofez: you are right, no longer used, the change was unnecessary, deleted this file

docker/imposter/agents/configuration/default.json
rodrigofez: true, also unnecessary, however I didn't delete this file, since it is referenced in another file, and it would be better this cleanup to be part of other task

asteriscos (dismissed):
Test 🟢 

Image

asteriscos (approved):

PR #9064, opened 2026-08-27
Sanitize RBAC-denied AI Assistant responses

### LINKED ISSUE #9057: AI Assistant: RBAC-denied provider/settings write returns HTTP 500 with a raw indexer security_exception (information disclosure)
## Summary

When a user whose indexer backend role lacks the AI Assistant settings permission attempts a provider write, the request is correctly denied, but the failure is surfaced to the client as an HTTP `500 Internal Server Error` carrying the raw OpenSearch `security_exception`. The exception text includes the internal cluster action name and the caller's username and backend roles. An authorization failure should be a clean `403` with a sanitized message, and the internal action name and identity details should not be returned to the browser.

## Environment

- Wazuh 5.0.0, single-node AIO (dashboard + indexer + manager), Ubuntu 20.04.
- Plugin `wazuh-ai-assistant` (`wazuhAiAssistant`), API root `/api/wazuh_ai_assistant`.
- Two dashboard users: `admin` (full access) and a second user with `kibana_user` + `readall` but without the AI Assistant settings permission.

## Steps to reproduce

1. Authenticate to the dashboard as the low-privilege user and obtain a session.
2. Call `POST /api/wazuh_ai_assistant/providers` with a valid provider body and the `osd-xsrf: true` header.

## Expected

`403 Forbidden` with a sanitized message (for example "You do not have permission to manage AI Assistant providers"). No internal cluster action name, username, or backend roles in the response body.

## Actual

`500 Internal Server Error` with the raw indexer exception, for example:

```json
{"statusCode":500,"error":"Internal Server Error","message":"security_exception: [security_exception] Reason: no permissions for [cluster:admin/ai_assistant/settings/read] and User [name=<user>, backend_roles=[...], requestedTenant=null]"}
```

The RBAC boundary itself works: the write does not take effect. The problem is only the response mapping.

## Impact

- Information disclosure: the response leaks the internal cluster action name (`cluster:admin/ai_assistant/settings/read`) and the caller's identity and backend roles to any authenticated but unauthorized user.
- Incorrect status semantics: an authorization failure is reported as a server error (`500`), which also complicates client handling and monitoring.

## Suggested fix

Catch the indexer `security_exception` in the provider/settings write handlers and translate it to a `403` with a generic, non-revealing message. Keep the full exception in the server logs only. Consider applying the same mapping to the other settings routes that check `plugin:wazuh/ai_assistant/settings/{read,write}`.

## Notes

Observed during QA functionality and security testing of the AI Assistant module. Tracked under the QA pentesting issue wazuh/wazuh-qa-automation#8701.

Related: #8383 reviews the read-only role showing raw permission errors across Management, similar UX concern but not the AI Assistant settings API path described here.

Ripdiegozz:
## Description

An RBAC-denied AI Assistant provider/settings write returned `500` with the raw
indexer `security_exception`, leaking the internal cluster action name, the
caller's username and their backend roles. The string was rendered in the
settings flyout callout, so no developer tools were needed to read it.

The authorization boundary worked — only the response mapping was wrong.

Closes #9057

## Proposed Changes

- An indexer `403` now returns `403 { message: 'You do not have permission to perform this action.' }`.
- Detection reads `error.statusCode ?? error.meta.statusCode`, strict `=== 403`. Not gated on `type === 'security_exception'`, because a DLS/FLS 403 carries a different type and would keep leaking.
- `withInternalErrorHandling` takes a required `logger` and records the full exception. It had none, so 11 routes returned errors with no server-side log at all.
- Wrapped `POST /providers/{id}/default` and `PUT /settings`, which had no error handling. `PUT /settings` rethrows permission denials so its actionable ISM `503` survives.
- The remaining `500`/`503` bodies pass through `redactSensitiveDetail`, which strips `User [...]` identity blocks and `cluster:`/`indices:` action names. Operational messages are untouched. The log always gets the full text.

No authorization check was changed.

### Results and Evidence

image

image

image

Reproduced against a live 5.0.0 stack with a user lacking
`plugin:wazuh/ai_assistant/settings/write`:

```
before: HTTP 500  security_exception ... [cluster:admin/ai_assistant/settings/read] and User [name=qa9057, backend_roles=[readall, kibanauser], ...]
after:  HTTP 403  You do not have permission to perform this action.
```

`plugins/wazuh-ai-assistant`: 130 suites, 2523 tests passing.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

New `server/routes/route-helpers.test.ts`, 14 cases. The security cases assert
**absence** — a body carrying the new message proves nothing on its own, since
the leaked text could still be alongside it. Covered: both `ResponseError`
shapes, a non-`security_exception` 403, strict `=== 403` with no coercion, the
unchanged 500 for other errors, the logger receiving full detail on both
branches, the nested `backend_roles=[...]` brackets, and an operational message
surviving redaction unchanged.

### How to Test

1. Create an indexer user with dashboard access and without
   `plugin:wazuh/ai_assistant/settings/write`.
2. Log in as that user, open `#/settings?addProvider=true` in the AI Assistant.
3. Fill the provider form and save. Type must be `openai_compatible` or
   `anthropic` — any other value fails schema validation with a `400` before the
   authorization path is reached.
4. Expected: the callout shows only "You do not have permission to perform this
   action.", and the Network response is `403`.
5. In the terminal running the dashboard, a `wazuhAiAssistant: route error`
   line must carry the full exception, including the action name and roles.

## Review Checklist

- [x] Code changes reviewed


CHANGELOG.md
```diff
@@ -41,6 +41,7 @@
...
+| [#9057](https://github.com/wazuh/wazuh-dashboard-plugins/issues/9057)                                                                                                                                                                                                           | Added sanitized authorization responses and server-side error logging to the AI Assistant routes                                                                                                                                                                                                                                                  |
```
Desvelao: **nitpick**: move the PR entry to the related for the AI assistant plugin, add the working issue as sub-issue of the main AI assistant plugin issue.

rodrigofez (dismissed):
🟢 Test

image

yenienserrano (approved):
LGTM :green_circle:

PR #9076, opened 2026-08-31
Enforce never-send fields with privacy mode off

### LINKED ISSUE #9065: AI Assistant: field-policy actions (including "Never send") are ignored when privacy mode is off, so those fields still reach the AI provider
## Summary

The AI Assistant privacy field policy lets an admin assign a per-field action (`allow`, `allow-scan`, `anonymize`, `never`) that governs what a tool-call digest sends to the external AI provider. That policy is only applied when privacy mode is active for the turn. Because privacy mode ships off by default and can be toggled off per request, a field explicitly configured as **Never send** (and one configured as **Anonymize**) is still transmitted to the provider in clear whenever privacy mode is off, which is the default state. The "Never send" action reads as an absolute egress control ("the field must not be sent at all"), so this silently violates the admin's configuration in the default configuration.

## Root cause

In `server/tools/executor.ts`, `finalizeDigest(digest, privacy, ...)` short-circuits and returns the digest unmodified when the turn has no privacy context:

```js
function finalizeDigest(digest, privacy, toolName, aggFields, isEscapeHatch, sampleFieldMaxLength) {
  if (!privacy) {
    return digest; // privacy off: applyFieldPolicy never runs
  }
  return capDigest(applyFieldPolicy(digest, privacy.fieldPolicy, privacy.pseudonymizer, aggFields, toolName, isEscapeHatch), sampleFieldMaxLength);
}
```

`applyFieldPolicy` (`server/tools/privacy.ts`), the only place the `never`/`anonymize`/`allow-scan` actions are enforced, is reached exclusively from this non-short-circuited branch. `privacy` is `undefined` whenever `resolvePrivacyEnabled` returns false for the turn (`server/routes/chat.ts`). Privacy mode is off by default: `GET /api/wazuh_ai_assistant/settings` returns `privacyDefaultOn: false` and `userCanOverride: true`, and the chat request may set `privacy.enabled: false`. So with privacy off, no field-policy action is applied at all, and every field the tool projects egresses in clear.

## Environment

- Wazuh 5.0.0, single-node AIO (dashboard + indexer + manager), Ubuntu 20.04.
- Plugin `wazuh-ai-assistant` (`wazuhAiAssistant`), API root `/api/wazuh_ai_assistant`.
- One authenticated `admin` dashboard user.

## Steps to reproduce

1. In AI Assistant Settings, set a visible field (for example `wazuh.agent.name`) to **Never send** and save (or `PUT /api/wazuh_ai_assistant/settings` with that field's action set to `never`).
2. Leave privacy mode off (the default), or send a chat request with `privacy.enabled: false`.
3. Ask a question that triggers an indexer tool call returning that field. Capture the request the plugin sends to the provider (for example by configuring an `openai_compatible` provider pointed at a mock endpoint you control and logging its request bodies).

## Expected

A field set to **Never send** is never included in the payload sent to the provider, regardless of the privacy-mode toggle. A hard "do not send" egress control should not depend on a separate, default-off setting.

## Actual

With `wazuh.agent.name` set to `never`, the tool-result digest sent to the provider differs only by the privacy toggle:

- Privacy off (default): `{"tool":"search_wazuh_data","counts":{...},"samples":[{"wazuh.agent.name":"node1-agent"}, ...],"columns":["wazuh.agent.name"], ...}` -- the value reaches the provider in clear.
- Privacy on: `{"tool":"search_wazuh_data","counts":{...},"samples":[{},{},{}],"columns":[], ...}` -- the field is dropped, as documented.

The RBAC and data boundaries are not involved here; this is purely that the configured field policy is not enforced on the provider-egress path when privacy mode is off.

## Impact

- A configured data-egress control silently fails in the default configuration. An admin who marks a sensitive field as "Never send" to keep it away from a third-party AI provider still has that field's real values sent on every privacy-off turn.
- The behavior is easy to miss because the policy works correctly the moment privacy mode is turned on, so a quick check with privacy on looks fine.

## Suggested fix

Enforce the `never` action unconditionally, independent of the privacy-mode toggle: a "must not be sent" field should be dropped from the digest on the provider-egress path whether or not privacy mode is active. `anonymize`/`allow-scan` are reasonably scoped to privacy mode (they are pseudonymization refinements), but `never` is a categorical deny and should always apply. Alternatively, if field policy is intended to be a sub-feature of privacy mode, make the Settings UI state that explicitly (for example disable or annotate the per-field actions while privacy mode is off) so "Never send" is not read as an absolute guarantee.

## Notes

Observed during QA security testing of the AI Assistant module. Tracked under the QA pentesting issue wazuh/wazuh-qa-automation#8701. Distinct from #8821, which was scoped to privacy mode enabled (its precondition 3) and concerned the locally rendered results table and the executed `_source` query, not the digest sent to the provider and not the privacy-off case; #8821 explicitly notes the digest correctly drops `never` fields under privacy mode.

Ripdiegozz:
mise ~/.config/mise/config.toml tools: gh@2.98.0
## Description

A field an admin sets to **Never send** still reached the AI provider whenever privacy mode was off, which is the default. The field policy lives inside `PrivacyContext`, and that object is only built when privacy mode is enabled, so `finalizeDigest` returned early and `applyFieldPolicy`, the only place `never` is enforced, never ran.

`docs/ref/modules/ai-assistant/security.md` lists never-send as its own mitigation, "stripped from digests unconditionally, even with privacy mode off". The contract block in `privacy.ts` and the Settings helper text say the same, so three documents disagreed with the code.

Closes #9065

## Proposed Changes

- Add `dropNeverFields` to `server/tools/privacy.ts`. It removes fields and rewrites nothing, and it covers the three surfaces `never` already covered under privacy mode: values in `samples`, buckets in `breakdown` (with `breakdownNote`), and the field name in `columns`. Aggregation semantics follow `scrubAggKey`.
- Run it from `finalizeDigest` on the privacy-off branch.
- Thread `fieldPolicy` as its own argument through `orchestrate` and `executeToolCall` instead of letting it ride inside `PrivacyContext`, so the toggle can no longer hide it.

`anonymize` and `allow-scan` stay scoped to privacy mode. Both need the pseudonymizer, and the toggle is the right place for them.

`FIELD_POLICY_DEFAULTS` declares no `never` entry, so default installations keep byte-identical digests. Reusing `applyFieldPolicy` over a policy filtered to its `never` entries was rejected: every remaining field would then resolve to allow-by-omission, whose branch in `scrubFieldValue` runs `prescanAndMint` and would rewrite IPs and FQDNs that privacy-off output has always sent in clear.

### Results and Evidence

Verified against the live stack by logging the outbound provider payload, with a control run that disabled the fix.

`get_field_values` over `wazuh.agent.name` set to `never`, privacy off:

Before

```json
"breakdown":[{"key":"node1-agent","count":1},{"key":"node2-agent","count":1},{"key":"node3-agent","count":1}],
"samples":[{"key":"node1-agent","doc_count":1},{"key":"node2-agent","doc_count":1},{"key":"node3-agent","doc_count":1}]
```

After

```json
"samples":[{"doc_count":1},{"doc_count":1},{"doc_count":1}]
```

With the default policy (no `never` entry), the outbound digest is identical with and without the change, both privacy on and privacy off. `anonymize` still yields `HOST_1`/`HOST_2`/`HOST_3` under privacy mode. With two fields set to `never`, an unrelated field (`wazuh.rule.id`) still passes through in full. The analyst keeps seeing real values in the results table, which never left the cluster.

### Artifacts Affected

`wazuh-ai-assistant` plugin only. No packaging or configuration changes.

### Configuration Changes

None. No new setting, and no default value changes.

### Documentation Updates

None needed. The change makes the code match what `docs/ref/modules/ai-assistant/security.md` already states.

### Tests Introduced

11 unit tests. Eight cover `dropNeverFields`: sample drops, tool-scoped resolution, aggregation `key` samples, scalar/`multi_terms`/`composite` bucket semantics, `breakdownNote` removal, and fieldless `date_histogram` pass-through. Three cover `executeToolCall` with privacy off, including a guard that a digest under `FIELD_POLICY_DEFAULTS` is byte-identical to the no-policy path.

Full plugin suite: 129 suites, 2519 tests.

## How to Test

1. Bring up `docker/osd-dev` and open the AI Assistant with a provider configured.
2. In **Settings > Privacy**, leave privacy mode off and set `wazuh.agent.name` to **Never send**. Save.
3. Point the provider at an endpoint whose request bodies you can read, or add a temporary log of the outbound messages before `adapter.chatStream` in `server/routes/chat.ts`.
4. Ask the assistant to list the distinct values of `wazuh.agent.name`, so it calls `get_field_values` over that field. The osd-dev indexer ships with no agent findings, so index a few documents into `wazuh-findings-v5-security` with `wazuh.agent.name` and a recent `@timestamp` first. Ask for that field by name rather than something like "which agent do these belong to": the second phrasing sends the model to `get_agents`, whose policy entry is `get_agents/name`, and it will show the real name.
5. The digest reaching the provider carries no agent name, and the results table in the browser still shows the real names.
6. Turn privacy mode on and repeat. The digest is the same.
7. Set the field back to **Anonymize** and repeat with privacy off. The real value goes out as before this change.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Correct labels applied


rodrigofez (approved):
## 🟢 Review

**Verdict: approve.** The fix does what #9065 asks and I reproduced the before/after on a live stack.

**Code vs. issue**

- Root cause matches: `finalizeDigest` short-circuited on the privacy-off branch, so `applyFieldPolicy` (the only enforcement point) never ran. `dropNeverFields` now runs on that branch and `fieldPolicy` is threaded separately from `privacyCtx`, so the privacy toggle can no longer hide it.
- Scope is right: `anonymize` / `allow-scan` stay privacy-mode-only (both need the pseudonymizer) — exactly the split the issue's *Suggested fix* proposes.
- All three surfaces `never` covers under privacy mode are covered here too: `samples`, `breakdown` (and `breakdownNote` when every bucket drops), and the `columns` schema hint. Bucket-key shapes (scalar / `multi_terms` / `composite`) mirror `scrubAggKey`'s drop semantics without rewriting any value.
- No default-config drift: `FIELD_POLICY_DEFAULTS` declares no `never` entry, so a default install's digests stay byte-identical (asserted by a test).
- `capDigest` deliberately not re-run on the privacy-off path — a drop-only pass cannot grow the digest past the cap `buildDigest` already applied. Reasoning is documented in place.

**Checks**

- `yarn test:jest --testPathPattern=tools/(privacy|executor)` → 221 passed (10 new `dropNeverFields` cases + executor end-to-end).
- Prettier (2.8.8, repo config) and ESLint clean on the touched files. CHANGELOG entry links the issue, not the PR.

**Manual verification**: mock OpenAI-compatible provider (`eval/mock_provider.js`) capturing the real provider request bodies; `wazuh.agent.name` set to **Never send**, privacy mode **off**, tool `get_vulnerabilities` over 1303 real vulnerability documents.

1. Admin config: `wazuh.agent.name = Never send`, privacy mode off (default).

image


2. Chat turn with **Privacy: Off** — local results table still shows the agent name (local render, never egressed).

image

3. What the provider actually received, before vs. after the fix (same request, code reverted in place for the "before" capture): value, agent-name breakdown buckets and the `columns` entry all disappear. Privacy-on output is unchanged.

image

PR #9083, opened 2026-09-01
fix(check-updates): send Accept-Encoding: gzip, br on CTI console requests

Desvelao:
## Description

Every request the dashboard makes against the CTI Console must include the
compression acceptance header `Accept-Encoding: gzip, br`. The two outbound
HTTP calls this repo makes directly to the CTI Console (the OAuth device
authorization token endpoint, used for CTI registration) relied on axios'
implicit default `Accept-Encoding` header rather than sending the exact value
expected by the CTI service.

## Proposed Changes

- Added an explicit `Accept-Encoding: gzip, br` header (alongside the existing
  `Content-Type`) to both CTI Console token requests in
  `plugins/wazuh-check-updates/server/services/cti-registration/token.ts`
  (`getCtiToken` and `pollCtiToken`), consolidated into a single
  `CTI_CONSOLE_REQUEST_HEADERS` constant so both call sites stay in sync.
- All other CTI-adjacent server code (update checks, content-manager
  subscription/update, consumers) talks to the indexer's own OpenSearch
  Content Manager plugin over the OpenSearch JS client transport, not
  directly to the CTI Console over HTTP, so it is unaffected by this change
  and out of scope here.

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- Added `plugins/wazuh-check-updates/server/services/cti-registration/token.test.ts`
  (previously missing at the service layer). Mocks `axios` and asserts that
  both `getCtiToken` and `pollCtiToken` call `axios.post` with
  `Accept-Encoding: gzip, br` and the existing `Content-Type` header, and
  that `pollCtiToken`'s `validateStatus` behavior (200/400 accepted, others
  rejected) is preserved.

## Review Checklist

- [x] Code changes reviewed
- [ ] ...

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -78,7 +78,8 @@ because Jest needs OSD's `setup_node_env`. See
 [`docs/dev/run-tests.md`](docs/dev/run-tests.md).
 
 ```bash
-# 1) Bring up the environment (from repo root)
+# 1) Bring up the environment (from repo root) if is not running already.
+docker ps # find the osd-dev-* container
 cd docker/osd-dev && ./dev.sh up
 
 # 2) Attach to the OSD container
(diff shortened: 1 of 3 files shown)


asteriscos (approved):
CR: 🟢 
Test: 🟢 

The request made to the token endpoint is including the expected header

Image

PR #9084, opened 2026-09-01
Make the AI Assistant out-of-credits message customizable

Ripdiegozz:
mise ~/.config/mise/config.toml tools: gh@2.98.0
## Description

The out-of-credits error in the AI Assistant was a fixed string built from the provider's response,
so a deployment could not point users at its own credits or plan-upgrade flow.

This adds `wazuh_ai_assistant.outOfCreditsMessage`, read from `opensearch_dashboards.yml`. One value
per environment, set by whoever provisions the deployment, deliberately not part of the
customer-editable indexer-backed provider settings. Unset by default, so existing deployments see no
change.

## Proposed Changes

**Setting.** Optional string in the plugin's server config schema, read once in `setup()` into a
`plugin-services.ts` singleton. Blank or whitespace-only counts as unset, since `??` alone would let
it through and render an error with no text.

**Detection** (`server/providers/retry.ts`, shared with `wazuh-brain.ts`), in order:

1. **HTTP 402**, authoritative on its own. DeepSeek, OpenRouter, Vercel AI Gateway, Together and
   Anthropic use it. Protocol semantics outlive prose, and 402 is not in `RETRYABLE_STATUSES`, so
   treating it as terminal costs no retry.
2. **Body markers**, for providers reporting on another status: Anthropic 400, OpenAI 429
   (`insufficient_quota`, `credit_balance_exhausted`). The underscored code does not match the
   spaced `credit balance` marker, so both are listed.

The check runs before the retry decision, so an `insufficient_quota` on a 429 is not retried
pointlessly. Rate limits are untouched: Groq's `rate_limit_exceeded` and Gemini's
`RESOURCE_EXHAUSTED` still retry. Gemini is why the markers avoid the bare word `quota` — it returns
the same "You exceeded your current quota" text OpenAI uses for billing, on a limit that resets on
its own.

**Rendering.** The message renders as Markdown so it can carry a link. `errorMarkdownProcessingPlugins`
overrides only EUI's `a` renderer (`target="_blank"`, `rel="noopener noreferrer"`) and applies to
error text only: a finished assistant answer keeps EUI's stock renderers plus the existing
`sanitizeAssistantMarkdown` defense, so model output and operator text stay on separate trust paths.
`status-callout.tsx` accepts a `ReactNode` body; a string body keeps its `<p>` wrapper.

**Drive-by fix.** A failed turn is marked with a failure reason, never `interrupted`, but the last
turn gets the same in-place Retry either way. `handleRetryLastTurn` accepted only an interrupted
answer, so on a failed turn the button rendered and did nothing. Out-of-credits makes it reachable,
hence fixing it here. A failed last turn is now replaced in place, as an interrupted one always was;
older failed turns still use "Ask again" and append.

#### With the setting

image

image

#### Without the setting

image

### Artifacts Affected

`plugins/wazuh-ai-assistant` only. No packages, executables, or shipped default config files.

### Configuration Changes

| Setting | Default | Effect |
| --- | --- | --- |
| `wazuh_ai_assistant.outOfCreditsMessage` | unset | Replaces the out-of-credits chat error. Accepts a Markdown link. Blank counts as unset. Rate limits unaffected. |

```yml
wazuh_ai_assistant.outOfCreditsMessage: 'Your organization is out of credits. [Add credits](https://example.com/billing) to continue.'
```

### Documentation Updates

`docs/ref/modules/ai-assistant/configuration.md` — added to the table of keys read from
`opensearch_dashboards.yml`.

### Tests Introduced

Colocated, covering both cases in the request plus the boundaries:

- **Detected.** Anthropic 400 and 402 `billing_error`; OpenAI 429 `insufficient_quota` /
  `credit_balance_exhausted`; DeepSeek, OpenRouter, Vercel and Together 402s; a 402 with an empty body.
- **Not detected.** Groq `rate_limit_exceeded` and Gemini `RESOURCE_EXHAUSTED` (both still retry);
  Anthropic 413; a 413 whose text reads like an upsell but is a per-minute token ceiling.
- **Precedence.** Default kept when unset, custom message when set, blank treated as unset, and no
  out-of-credits body consumes the retry budget.
- **No leak into other failures.** With an override configured, 401, 403, 404, a context-length 400
  and a 422 each keep the provider's own text and stay unretried.
- **UI.** Link renders as a new-tab anchor in the failure reason and the error banner, raw HTML is
  stripped, and Retry on a failed last turn re-sends the question.

Full plugin suite green (2616 tests, 133 suites), `yarn lint` and Prettier clean.

### With a real provider

The closest thing to production. Any account with a drained balance works, and none of it costs
anything: you need a key that gets rejected, not one that works.

| Provider | How to get a rejection | Response | Type in Settings |
| --- | --- | --- | --- |
| OpenAI | New account, no payment method | `429` `insufficient_quota` / `credit_balance_exhausted` | `OpenAI-compatible` |
| Anthropic | New console account, no credits purchased | `400` credit balance too low | `Anthropic` |
| OpenRouter | New account, and a **paid** model (a `:free` one still answers) | `402` Insufficient credits | `OpenAI-compatible` |

Confirm the account is actually rejecting before wiring it into the dashboard:

```bash
curl -i https://api.openai.com/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H 'Content-Type: application/json' \
  -d '{"model":"gpt-4o-mini","messages":[{"role":"user","content":"hi"}],"max_tokens":8}'
```

Then add it under **AI Assistant → Settings → Providers**, set it as default, and jump to step 2
below. Put the key in the keystore rather than the YAML file.

Groq is not useful here: it rate-limits rather than reporting a drained balance.

### With a mock

Faster and deterministic, and the only way to exercise the negative cases: the errors that must
keep behaving exactly as they did before this PR. Detection reads the status and body, nothing else.

**1.** Save as `mock-out-of-credits.js`, run `node mock-out-of-credits.js 4545 402`, and point an
`OpenAI-compatible` provider at `http://localhost:4545`:

```js
const http = require('http');
const port = Number(process.argv[2]) || 4545;
const style = process.argv[3] || '402';
const styles = {
  // out of credits
  402: [402, { error: { code: 402, message: 'Insufficient credits.' } }],
  quota429: [429, { error: { code: 'insufficient_quota', message: 'You have no credits remaining.' } }],
  balance400: [400, { error: { message: 'Your credit balance is too low.' } }],
  // must keep their own behaviour
  ratelimit: [429, { error: { code: 'rate_limit_exceeded', message: 'Rate limit reached (TPM).' } }],
  badkey: [401, { error: { code: 'invalid_api_key', message: 'Incorrect API key provided.' } }],
  forbidden: [403, { error: { message: 'You do not have access to this model.' } }],
  notfound: [404, { error: { message: 'The model `gpt-nope` does not exist.' } }],
  toolong: [413, { error: { code: 'context_length_exceeded', message: 'Request too large. Reduce your message size.' } }],
  boom: [500, { error: { message: 'Internal server error' } }],
};
const [status, body] = styles[style] || styles['402'];
http.createServer((_q, res) => {
  res.writeHead(status, { 'Content-Type': 'application/json' });
  res.end(JSON.stringify(body));
}).listen(port, () => console.log(`mock ${style} on :${port} -> ${status}`));
```

**2.** Add `wazuh_ai_assistant.outOfCreditsMessage` to `opensearch_dashboards.yml` and **restart the
dashboard**. It is read once at plugin setup, so a reload alone will not pick up a change.

**3.** Ask anything, then check:

| Setup | Expected |
| --- | --- |
| Setting present (real provider or style `402`) | Configured text, link clickable and opening in a new tab |
| Setting commented out, restart | `Provider responded with HTTP <status>: …` with the provider's own text, never the rate-limit copy |
| Setting set to `''`, restart | Same as above. Never an empty error |
| Styles `quota429`, `balance400` | Same configured text |
| Style `ratelimit` | Configured text must **not** appear; the turn retries and ends with the rate-limit text |
| Click **Retry** on the failed turn | The question is re-sent and the turn is replaced in place |
| Message set to `[x](javascript:alert(1))` | Renders as plain `x`, no link, no dialog |
| Styles `badkey`, `forbidden`, `notfound`, `toolong` | Unchanged: `Provider responded with HTTP <status>: …` with the provider's own text, no retry, and the configured message must **not** appear |
| Style `boom` | Unchanged: retried, then the same raw 500 text |

Automated, from inside the Docker dev container (Jest needs OSD's `setup_node_env`):

```bash
cd plugins/wazuh-ai-assistant && yarn test:jest --runInBand
```

## Review Checklist

- [x] Code changes reviewed
- [ ] PR is linked to the relevant issue(s) — internal request, link intentionally withheld


yenienserrano (approved):
LGTM :green_circle:

```yml
wazuh_ai_assistant:
  encryptionKey: 'X53wBSic7YMVYLYXrbPt93q4Sgbq0LiPzPr7xACTTn4='
```

Image

```yml
wazuh_ai_assistant:
  encryptionKey: 'X53wBSic7YMVYLYXrbPt93q4Sgbq0LiPzPr7xACTTn4='
  outOfCreditsMessage: 'Your organization is out of credits. [Add credits](https://example.com/billing) to continue.'
```

Image

Desvelao (approved):

PR #9095, opened 2026-09-02
Read agents disconnection settings from the node configuration

### LINKED ISSUE #9092: Deprecated `wazuh-manager-monitord` module
# Description

Wazuh Manager will deprecate the `wazuh-manager-monitord` module. We need to assess any related changes that need to be done in the Wazuh dashboard.

#### Related Wazuh Manager issue
- https://github.com/wazuh/wazuh/issues/38589

### Related views

Image

Image

## Tasks

- [ ] Check any used Wazuh Manager endpoint using a monitord property
- [ ] Deprecate any reference or component related to monitord module

rodrigofez:
## Description

The manager is removing the `monitord` module, and with it the `monitor` component of the Server API configuration endpoint.

One place in the dashboard depended on it: the **Agents settings** panel in Server management > Settings > Global Configuration > Global. It asked the manager for `GET /cluster/{node}/configuration/monitor/global`, which the API answered by querying the monitord daemon directly. Once that daemon is gone, the request fails and the panel breaks.

The two settings it shows are not going away, so this reads them from the node configuration instead.

Closes #9092

## Proposed Changes

- The panel now reads `agents_disconnection_time` and `agents_disconnection_alert_time` from `GET /cluster/{node}/configuration`, the same request the Logging panel above it already makes. No extra call is added.
- The labels no longer say "Seconds". The old endpoint always returned seconds (`900`); the node configuration returns the value as the user wrote it in `wazuh-manager.conf` (`15m`), so a label promising seconds would be wrong.
- A setting that is not written in the configuration file now shows `-` instead of disappearing from the panel.
- Removed the imposter mock for the deleted endpoint.

Not changed on purpose: `plugins/main/common/api-info/endpoints.json` still mentions `monitor`. That file is generated by `yarn generate:api-data` from the manager API spec, so it updates itself once the manager side ships.

### Results and Evidence

image

### Artifacts Affected

None. Browser-side only.

### Configuration Changes

None. `agents_disconnection_time` and `agents_disconnection_alert_time` stay where they are in `wazuh-manager.conf`. Only the way the dashboard reads them changes.

### Documentation Updates

None.

### Tests Introduced

New test file for the panel, which had none. It covers the configured values and their labels, a setting missing from the configuration file, a failed read, a node with no `global` block, and the server not being ready yet.

The existing `wz-fetch` test that used `monitor` as its example component now uses `request`.

Full `configuration/` suite passes: 18 files, 125 tests.

## Review Checklist

- [x] Code changes reviewed


Ripdiegozz (commented):
🟢 Tests

Image

With a property commented:

Image

Image

**Comment:** If neither setting is declared in wazuh-manager.conf, global comes back as {} and this panel renders a bare Configuration not available box with no heading, right below Logging settings, which renders fine, so it could read as if logging were the broken one.

Image

Ripdiegozz (dismissed):

Desvelao (approved):

PR #9119, opened 2026-09-04
Add SSL verification switch to the agent enrollment wizard

### LINKED ISSUE #9103: Implement SSL verification in the agent enrollment wizard
## Description

The Wazuh agent will support validating the Wazuh manager certificate during enrollment (see [wazuh/wazuh#38684](https://github.com/wazuh/wazuh/issues/38684)). In 5.x the agent's only transport is HTTPS, and enrollment is authenticated one-way (agent → manager), so the manager is authenticated solely by TLS. Today the agent's `<ssl><verification_mode>` defaults to `none`, which the referenced issue proposes to change so that verification is the expected posture.

The **Endpoints Summary > Deploy new agent** wizard is the place where users build their enrollment command, and it currently has no way to express this. Every command it generates is produced with whatever the agent's default verification posture happens to be, and the user gets no signal that TLS verification is even a decision they are making.

We need the wizard to expose this choice and reflect it in the generated command:

- Add a **switch for SSL verification** to the enrollment wizard (Optional settings step).
- When the switch is **off**, the generated command must include the new environment variable `SSL_VERIFICATION=none`.
- When the switch is **on**, the generated command must include the manager CA so the agent can validate the certificate.

The wizard already has the extension points for this: optional parameters are declared in `tOptionalParameters` and rendered into the command by `optionalParamsDefinitions`, so a new parameter plugs into the same generator used for `WAZUH_MANAGER_ENDPOINT`, `WAZUH_AGENT_NAME`, `WAZUH_AGENT_GROUP` and `WAZUH_REGISTRATION_PASSWORD`. The value must be rendered correctly for all supported targets, since each OS family builds the command differently:

- **Linux (DEB/RPM)** — environment variables are prefixed inline to the install command (`sudo VAR='value' dpkg -i ...`).
- **Windows** — variables are appended as `msiexec.exe` properties.
- **macOS** — variables are written to `/tmp/wazuh_envs`, joined with `&&`, before running the installer.

### Open points to confirm before implementing

- **Exact variable names.** The switch-off value is `SSL_VERIFICATION=none`. The counterpart used to pass the CA (variable name, and whether it takes a CA file path on the endpoint or inline PEM content) must be confirmed with the agent team, since the agent-side issue is still open. The existing wizard variables are all `WAZUH_`-prefixed, so confirm whether these follow the same convention.
- **Where the CA comes from.** Whether the wizard can obtain/serve the manager CA (and therefore prefill the value), or whether the user must supply a path to a CA already present on the endpoint. This determines whether the "on" state needs an additional input next to the switch.
- **Default state of the switch.** It should match whatever the agent ships as its default verification mode once [wazuh/wazuh#38684](https://github.com/wazuh/wazuh/issues/38684) is resolved, so the wizard does not silently generate a command that contradicts the agent default.
- **Version gating.** Older agents will not understand these variables. Confirm whether the switch must be hidden or disabled when the selected package version predates the feature.

## Tasks

- [x] Confirm with the agent team the final variable names and accepted values for both states (verification off, and CA when verification is on).
- [x] Add an `sslVerification` field to the enrollment form configuration in `register-agent.tsx`, with the agreed default value.
- [x] Render the switch in the **Optional settings** step (`optionals-inputs.tsx`), with help text explaining the security implication of turning verification off.
- [x] Add the new parameter to `tOptionalParameters` and its rendering rule to `optionalParamsDefinitions` in `os-commands-definitions.ts`.
- [x] Include the new parameter in the ordered lists in `getAllOptionals` and `getAllOptionalsMacos` so it is emitted for Linux, Windows and macOS commands.
- [x] Verify the generated command is valid for every supported OS/architecture combination (DEB amd64/aarch64, RPM amd64/aarch64, Windows MSI, macOS Intel/Apple silicon), including correct quoting and escaping.
- [x] Decide and implement the behaviour when the CA is not available or not provided while the switch is on (validation error vs. omitting the parameter).
- [x] Add colocated unit tests covering both switch states across the three OS families (extend `register-agent-os-commands-services.test.ts`, `use-register-agent-commands.test.ts` and `command-generator.test.ts`).
- [x] Add the `CHANGELOG.md` entry linking to this issue.
- [x] Attach evidence (screenshots of the switch in both states and the resulting commands per OS) to the PR.

## Additional information

Source issue: [wazuh/wazuh#38684](https://github.com/wazuh/wazuh/issues/38684) — _Agent HTTPS transport defaults to verification_mode none, so a default install performs no TLS certificate verification_.

### Code references

Base commit: `99cf6425c0299f5a4a7ae2c47be63ad86e990eb5`

- Form fields declaration (`initialFields`) — [`containers/register-agent/register-agent.tsx`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/99cf6425c0299f5a4a7ae2c47be63ad86e990eb5/plugins/main/public/components/endpoints-summary/register-agent/containers/register-agent/register-agent.tsx#L71)
- Optional settings step UI — [`components/optionals-inputs/optionals-inputs.tsx`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/99cf6425c0299f5a4a7ae2c47be63ad86e990eb5/plugins/main/public/components/endpoints-summary/register-agent/components/optionals-inputs/optionals-inputs.tsx)
- Step wiring and `setOptionalParams` call — [`containers/steps/steps.tsx#L104-L130`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/99cf6425c0299f5a4a7ae2c47be63ad86e990eb5/plugins/main/public/components/endpoints-summary/register-agent/containers/steps/steps.tsx#L104-L130)
- Optional parameter types — [`core/config/os-commands-definitions.ts#L74`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/99cf6425c0299f5a4a7ae2c47be63ad86e990eb5/plugins/main/public/components/endpoints-summary/register-agent/core/config/os-commands-definitions.ts#L74)
- Optional parameter → command rendering — [`core/config/os-commands-definitions.ts#L211`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/99cf6425c0299f5a4a7ae2c47be63ad86e990eb5/plugins/main/public/components/endpoints-summary/register-agent/core/config/os-commands-definitions.ts#L211)
- Per-OS command assembly and parameter ordering — [`services/register-agent-os-commands-services.tsx`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/99cf6425c0299f5a4a7ae2c47be63ad86e990eb5/plugins/main/public/components/endpoints-summary/register-agent/services/register-agent-os-commands-services.tsx)
- Command generator documentation — [`core/register-commands/README.md`](https://github.com/wazuh/wazuh-dashboard-plugins/blob/99cf6425c0299f5a4a7ae2c47be63ad86e990eb5/plugins/main/public/components/endpoints-summary/register-agent/core/register-commands/README.md)

rodrigofez:
## Description

The agent now verifies the manager's certificate by default (wazuh/wazuh#38684, fixed in wazuh/wazuh#38786), but the **Deploy new agent** wizard had no way to express that choice. Users got no signal that TLS verification was a decision they were making, and no way to point the agent at a manager CA.

This adds a switch for it in the wizard's **Optional settings** step.

Closes #9103

## Proposed Changes

The switch is on by default and reveals an optional field for the manager CA file path:

| Switch | CA path | Command gets |
| --- | --- | --- |
| On | empty | nothing — the agent verifies against the endpoint's OS trust store |
| On | filled | `WAZUH_REGISTRATION_CA='<path>'` |
| Off | — | `SSL_VERIFICATION='none'` |

Two things worth a reviewer's attention:

- **The "on" states add nothing to the command.** The agent already verifies by default, and it works out the right mode on its own when given a CA. Setting the mode explicitly would break things: the agent refuses a CA when the mode is pinned to `system` and discards the CA, so a user adding `WAZUH_REGISTRATION_CA` to a copied command by hand would silently lose it.
- **The strictest mode is not offered.** Verified on a real agent: the default already checks both the certificate chain and the hostname, so the separate `full` mode adds little over what the "on" states give you. Left for a follow-up if that changes.

Files changed:

| File | Change |
| --- | --- |
| `containers/register-agent/register-agent.tsx` | The two new form fields. |
| `utils/validations.tsx` | Path validation. Only a single quote is rejected, since it would break the command's quoting; spaces, backslashes and drive letters are fine, as the path is resolved on the endpoint. |
| `core/config/os-commands-definitions.ts` | How each field turns into a command variable. |
| `services/register-agent-os-commands-services.tsx` | Both fields added to the parameter order lists (Linux/Windows and macOS differ). |
| `services/register-agent-services.tsx` | Drops the CA when the switch is off, so turning it off after typing a path doesn't produce a command that supplies a CA and then refuses to use it. The field keeps its value, so turning the switch back on restores it. |
| `components/optionals-inputs/optionals-inputs.tsx` | The switch, the CA field, and a warning shown while verification is off. |
| `services/register-agent-steps-status-services.tsx` | Field label for the invalid-fields message. |
| `core/register-commands/types.ts` | Parameter values can now be booleans. |
| `core/register-commands/README.md` | Notes for whoever adds the next parameter. |
| `docker/osd-dev/dev.yml`, `docker/osd-dev/manager/entrypoint.sh` | Dev environment agent enrollment (see below). |

### Development environment

`docker/osd-dev` agents stopped enrolling once the agent started verifying by
default, so this branch also fixes the local stack. This is independent of the
wizard change — the agents broke on their own — but it belongs here because it
makes the dev environment exercise the same verified path the switch produces,
so a regression shows up locally.

Four separate problems had to be fixed to get there:

| Problem | Fix |
| --- | --- |
| The `ubuntu` image ships no CA bundle, so the agent exits with *"no OS CA bundle was found"* | install `ca-certificates` in the deb agent |
| The manager served a self-signed certificate whose SAN does not cover `wazuh.manager.local` | serve a CA-signed certificate instead |
| remoted resolves its certificate paths relative to the manager home, so an absolute path into the shared volume is *"missing or unreadable"* and the HTTPS server never starts | the entrypoint copies the pair into `etc/certs/` and writes a relative path |
| The existing `wazuh.manager.local` certificate is generated with cfssl's **client** profile, which agents reject with *"unsuitable certificate purpose (26)"* | the generator now also mints an `agent-comms` certificate with `-profile=server` |

The manager keeps its root CA readable only by its own user and `wazuh-agentd`
drops privileges, so the agents copy the CA to a readable path instead of
pointing `WAZUH_REGISTRATION_CA` at the shared volume directly.

Both agents enroll again, verifying the full chain and the hostname:

```
001 wazuh.agent.deb.local any 596b4a8c...
002 wazuh.agent.rpm.local any 34946bfe...
```

Setup cost is one extra `cfssl gencert` (measured at 0.51 s) plus a 333 KB
package. The wait loop that guards against the certificates not being ready
exits on its first check in practice — the agents spend their first minutes on
`apt`/`yum` — and it now gives up after 120 s with an explanatory message rather
than hanging forever.

### Results and Evidence

Checked against the running dev stack (plugin `5.0.0-beta5`) for every OS and architecture the wizard offers.

**Switch on, CA input revealed:**

**Switch off, warning shown:**

All three states, DEB amd64:

```
on, no CA    wget <url> && sudo WAZUH_MANAGER_ENDPOINT='1.1.1.1' dpkg -i ./wazuh-agent_5.0.0-beta5_amd64.deb
on + CA      wget <url> && sudo WAZUH_MANAGER_ENDPOINT='1.1.1.1' WAZUH_REGISTRATION_CA='/etc/ca.pem' dpkg -i ./…
off          wget <url> && sudo WAZUH_MANAGER_ENDPOINT='1.1.1.1' SSL_VERIFICATION='none' dpkg -i ./…
```

The off case above had a CA path typed in beforehand, and it was correctly left out.

A path with spaces in it, on every platform:

```
RPM amd64      sudo … WAZUH_REGISTRATION_CA='/etc/my certs/ca.pem' rpm -ihv wazuh-agent-5.0.0-beta5.x86_64.rpm
RPM aarch64    sudo … WAZUH_REGISTRATION_CA='/etc/my certs/ca.pem' rpm -ihv wazuh-agent-5.0.0-beta5.aarch64.rpm
DEB aarch64    sudo … WAZUH_REGISTRATION_CA='/etc/my certs/ca.pem' dpkg -i ./…
Windows        msiexec.exe /i $env:tmp\… /q WAZUH_MANAGER_ENDPOINT='1.1.1.1' WAZUH_REGISTRATION_CA='/etc/my certs/ca.pem'
macOS Intel    echo "WAZUH_MANAGER_ENDPOINT='1.1.1.1' && WAZUH_REGISTRATION_CA='/etc/my certs/ca.pem'" > /tmp/wazuh_envs
macOS Apple    echo "WAZUH_MANAGER_ENDPOINT='1.1.1.1' && WAZUH_REGISTRATION_CA='/etc/my certs/ca.pem'" > /tmp/wazuh_envs
```

Switch off, Windows and macOS:

```
Windows        msiexec.exe /i $env:tmp\… /q WAZUH_MANAGER_ENDPOINT='1.1.1.1' SSL_VERIFICATION='none'
macOS Intel    echo "WAZUH_MANAGER_ENDPOINT='1.1.1.1' && SSL_VERIFICATION='none'" > /tmp/wazuh_envs
```

A path containing a single quote is rejected: `The character "'" is not valid in a file path.`

### Artifacts Affected

None. Browser-side changes to the `wazuh` plugin only.

### Configuration Changes

No dashboard settings change. The wizard now emits two agent install-time variables that already exist in the agent:

- `SSL_VERIFICATION='none'` when the switch is off.
- `WAZUH_REGISTRATION_CA='<path>'` when the switch is on and a path is given.

With the switch left alone and no CA supplied, the generated command is identical to before this change.

### Documentation Updates

`core/register-commands/README.md` — how a boolean-backed parameter behaves, and the reminder to register a new parameter in both order lists.

### Tests Introduced

135 tests pass across 13 suites in the register-agent area. The full `plugins/main` suite passes: 243 suites, 1893 tests, no failures.

- `core/config/os-commands-definitions.test.ts` (new) — what each field renders per platform: on renders nothing, off renders `SSL_VERIFICATION='none'`; CA paths are trimmed, spaces kept, empty and whitespace-only render nothing.
- `services/register-agent-os-commands-services.test.ts` — the three states across DEB amd64/aarch64, RPM amd64/aarch64, Windows and macOS, plus a spaces-in-path case.
- `services/register-agent-services.test.ts` — both fields reach the command generator, and the CA is dropped when the switch is off and kept when it's on.
- `utils/validations.test.tsx` — path validation accepts POSIX, Windows and relative paths and paths with spaces; rejects a single quote.

## Review Checklist

- [x] Code changes reviewed


Desvelao (dismissed):

yenienserrano (changes requested):
If you enter “Invalid CA path + switch off,” it returns an error message and does not display the command to install.

https://github.com/user-attachments/assets/60075b76-4edd-44b9-85d7-ae90cfbedf10

yenienserrano (approved):
LGTM :green_circle:

https://github.com/user-attachments/assets/05f811a1-0d6f-4d70-9276-9fee0414ca6e

PR #9129, opened 2026-09-07
Neutralize formula values in CSV exports

Ripdiegozz:
## Description

CSV exports wrote cell values verbatim, so a value starting with `=`, `+`, `-` or `@` was
evaluated as a formula when an analyst opened the file in a spreadsheet (OWASP CSV Injection,
CWE-1236).

The values are not all operator-supplied: an agent's `os.*` fields are reported by the agent and
declared as free-form strings with no `pattern` in the Server API specification, so a compromised
host chooses them.

## Proposed Changes

- **New** `plugins/main/common/services/neutralize-csv-formula.ts` — a pure, recursive helper that
  prefixes an apostrophe to any string starting with a trigger character (`=` `+` `-` `@`, tab,
  carriage return, and the full-width variants OWASP calls out; a leading line feed is excluded on
  purpose). It lives in `common/` because one call site is in `server/` and two are in `public/`.
- Applied **unconditionally** at all three CSV writers in `plugins/main`: the `POST /api/csv` route
  (`server/controllers/wazuh-api.ts`), the data-grid export (`data-grid-service.ts`) and the
  agent-stats export (`agents/stats/table.tsx`).
- **`agents/stats/table.tsx` migrated to `json2csvAsync`.** Its hand-rolled `join(',')` writer
  emitted no quoting, so a value containing a comma shifted every later column.
- **`Content-Disposition: attachment` added** to `POST /api/csv`. Its filename derives from the
  caller-supplied path, so it is constrained to `[A-Za-z0-9_.-]` — without that, adding the header
  would have introduced response header injection while fixing a different injection.

**The traversal has to be recursive**, and that is the whole point: `json-2-csv` resolves its
dotted `options.keys` selectors against the still-nested object and flattens *afterwards*, so a
top-level-only pass leaves `os.name` live — the exact field the defect exploits.

### Results and Evidence

An agent reporting `os.name` as `=1337*2` was exported through each of the three writers, on a
5.0.0 stack:

| Writer | Downloaded via | Exported cell | Calc |
| --- | --- | --- | --- |
| `POST /api/csv` | the endpoint, `path: /agents` | `'=1337*2` | inert text |
| data-grid | IT Hygiene → System → OS → **Export Formatted** | `'=1337*2` | inert text |
| agent-stats | Agent → Stats → **Download CSV** | `'=1337*2` | inert text |

**Before** — the same export without the fix. `OS name` renders `2674`: Calc evaluated
`=1337*2`.

vulnerable-1

**After** — `OS name` renders `'=1337*2` as literal text.

fixed-1

The agent-stats export also demonstrates the quoting fix, on a logcollector path with a comma:

```
fixed:    "/var/log/app,prod.log",120,4096   → 3 fields, aligned
pre-fix:   /var/log/app,prod.log,120,4096    → 4 fields, every later column shifted
```

Checks:

```
jest (full plugins/main)   246 suites, 1944 tests, 69 snapshots — pass
prettier                   clean
eslint                     zero delta against the base tree
tsc --noEmit --strict      clean on the new helper
knip                       nothing from this change reported
```

### How to Test

Requires a running `docker/osd-dev` stack with at least one enrolled agent. The blocks below
handle any number of agents.

**1. Plant the payload and download the server-side export.** One paste — it finds every agent
container and the API host id on its own, waits until all of them have synced, and writes
`agents.csv` to the current directory:

```bash
AGENTS=$(docker ps --format '{{.Names}}' | grep 'wazuh.agent')
for AGENT in $AGENTS; do
  docker exec "$AGENT" bash -c '
    [ -f /etc/os-release.orig ] || cp /etc/os-release /etc/os-release.orig
    sed -i "s|^NAME=.*|NAME=\"=1337*2\"|" /etc/os-release
    /var/ossec/bin/wazuh-control restart' >/dev/null 2>&1
  echo "payload planted on $AGENT"
done
EXPECTED=$(echo "$AGENTS" | grep -c .)

DASH=${DASH:-https://localhost:5601}
J=$(mktemp)
curl -sk -c "$J" -X POST "$DASH/auth/login" -H 'osd-xsrf: true' \
  -H 'Content-Type: application/json' -d '{"username":"admin","password":"admin"}' -o /dev/null
HOST=$(curl -sk -b "$J" "$DASH/hosts/apis" | sed -n 's/.*"id":"\([^"]*\)".*/\1/p')
curl -sk -b "$J" -c "$J" -X POST "$DASH/api/login" -H 'osd-xsrf: true' \
  -H 'Content-Type: application/json' -d "{\"idHost\":\"$HOST\"}" -o /dev/null

echo -n "waiting for $EXPECTED agent(s) to sync"
until [ "$(curl -sk -b "$J" -X POST "$DASH/api/request" -H 'osd-xsrf: true' \
  -H 'Content-Type: application/json' \
  -d "{\"id\":\"$HOST\",\"method\":\"GET\",\"path\":\"/agents\",\"body\":{}}" \
  | grep -o '=1337\*2' | wc -l)" -ge "$EXPECTED" ]; do echo -n .; sleep 5; done
echo " ok"

curl -sk -b "$J" -X POST "$DASH/api/csv" -H 'osd-xsrf: true' \
  -H 'Content-Type: application/json' \
  -d "{\"id\":\"$HOST\",\"path\":\"/agents\",\"filters\":[]}" -o agents.csv
echo "neutralized cells: $(grep -o "'=1337\*2" agents.csv | wc -l) of $EXPECTED expected"
```

Expected last line: `neutralized cells: N of N expected`. Without this change it reports `0`.

**2. Download the two browser-side exports** and check their `OS name` / `Location` cells the same
way:

- IT Hygiene → System → OS → **Export Formatted**
- Agent → Stats → **Download CSV**

**3. Open any of the three files in a spreadsheet.** The `OS name` cell must read `'=1337*2` as
literal text.

To see the defect itself, strip the apostrophe and reopen — this is what the previous writer
emitted:

```bash
sed "s/,'/,/g" agents.csv > agents-unfixed.csv
```

`OS name` then renders `2674`, because Calc evaluated `=1337*2`.

Two notes so the result is not misread. In LibreOffice's CSV import dialog **Evaluate formulas**
must be checked, or the payload renders as inert text with *and* without this change; and
**Merge delimiters** must be unchecked, or empty fields collapse and every later column shifts.
Both settings are sticky per user profile. Also, `=` is the only trigger LibreOffice evaluates on
import — the others apply to Excel and are covered by unit tests.

**4. Revert every agent.** Idempotent, and it only touches agents that actually hold a backup:

```bash
for AGENT in $(docker ps --format '{{.Names}}' | grep 'wazuh.agent'); do
  if docker exec "$AGENT" test -f /etc/os-release.orig; then
    docker exec "$AGENT" bash -c '
      mv /etc/os-release.orig /etc/os-release
      /var/ossec/bin/wazuh-control restart' >/dev/null 2>&1
    echo "reverted $AGENT"
  else
    echo "nothing to revert on $AGENT"
  fi
done
```

### Artifacts Affected

`plugins/main` only — one new `common/` service, three modified export paths, four test suites.
No dependency or build change: `json-2-csv` was already a dependency.

### Configuration Changes

None. Neutralisation is unconditional. `wazuh-dashboard-reporting` carries an equivalent
guard, but only on its Excel path (`params.excel`) and only over already-flattened top-level
fields, so it could not be reused here.

Accepted consequence: neutralised values are **not** byte-identical to the source. LibreOffice does
not consume a leading apostrophe on CSV import, so the analyst sees `'=1337*2` in the cell, and a
downstream script consuming these exports sees it too. Per OWASP no sanitisation strategy is safe
for all spreadsheets *and* all consumers, so the tradeoff can only be relocated. Only cells that
actually start with a trigger character are altered; every other value is byte-identical.

### Documentation Updates

None required.

### Tests Introduced

52 tests across four suites.

- `neutralize-csv-formula.test.ts` — every trigger including the full-width variants, `\n` pinned
  as *not* a trigger, nested-object and array depth, non-strings untouched (`-5` the number versus
  `"-5"` the string), key order preserved, input never mutated.
- `server/controllers/wazuh-api.test.ts` — **new file**; this controller had no test at all. Covers
  the nested `os.name` case through the real unmocked `json-2-csv`, plus hostile export paths
  against the `Content-Disposition` filename.
- `data-grid-service.test.ts` and `agents/stats/table.test.tsx` — extended with byte-exact
  assertions on header titles, column order, missing values and comma/quote/newline round-tripping.

## Review Checklist

- [x] Code changes reviewed
- [x] Correct labels applied (`no-changelog`)

### DIFF
--- a/plugins/main/common/services/neutralize-csv-formula.test.ts
+++ b/plugins/main/common/services/neutralize-csv-formula.test.ts
@@ -1,0 +1,148 @@
+import {
+  CSV_FORMULA_TRIGGER_CHARACTERS,
+  neutralizeCsvFormulaValues,
+} from './neutralize-csv-formula';
+
+describe('neutralizeCsvFormulaValues', () => {
+  describe('formula-leading strings are neutralized', () => {
+    it.each`
+      trigger | value         | description
+      ${'='}  | ${'=1+1'}     | ${'equals'}
+      ${'+'}  | ${'+1+1'}     | ${'plus'}
+      ${'-'}  | ${'-1+1'}     | ${'minus'}
+      ${'@'}  | ${'@SUM(1)'}  | ${'at'}
+      ${'\t'} | ${'\t=1+1'}   | ${'tab'}
+      ${'\r'} | ${'\r=1+1'}   | ${'carriage return'}
+      ${'＝'} | ${'＝1+1'}    | ${'full-width equals'}
+      ${'＋'} | ${'＋1+1'}    | ${'full-width plus'}
+      ${'－'} | ${'－1+1'}    | ${'full-width minus'}
+      ${'＠'} | ${'＠SUM(1)'} | ${'full-width at'}
+    `(
+      'neutralizes a value leading with $description ($trigger)',
+      ({ trigger, value }) => {
+        const result = neutralizeCsvFormulaValues(value);
+
+        expect(result.startsWith(trigger)).toBe(false);
+        expect(result).toContain(value);
+      },
+    );
+
+    it('exposes the exhaustive trigger set', () => {
+      expect(CSV_FORMULA_TRIGGER_CHARACTERS).toEqual([
+        '=',
+        '+',
+        '-',
+        '@',
+        '\t',
+        '\r',
+        '＝',
+        '＋',
+        '－',
+        '＠',
+      ]);
+    });
+
+    it('prefixes an apostrophe so the original text stays readable', () => {
+      expect(neutralizeCsvFormulaValues('=1+1')).toBe("'=1+1");
+    });
+  });
+
+  describe('values that must not change', () => {
+    it('does not neutralize a leading newline', () => {
+      expect(neutralizeCsvFormulaValues('\n=1+1')).toBe('\n=1+1');
+    });
+
+    it.each`
+      value             | description
+      ${'plain'}        | ${'a plain string'}
+      ${'Ubuntu 22.04'} | ${'an ordinary os name'}
+      ${'a=1+1'}        | ${'a trigger that is not leading'}
+      ${''}             | ${'an empty string'}
+      ${-5}             | ${'a negative number'}
+      ${0}              | ${'zero'}
+      ${true}           | ${'a boolean'}
+      ${null}           | ${'null'}
+      ${undefined}      | ${'undefined'}
+    `('returns $description unchanged', ({ value }) => {
+      expect(neutralizeCsvFormulaValues(value)).toBe(value);
+    });
+
+    it('returns a bigint unchanged', () => {
+      expect(neutralizeCsvFormulaValues(BigInt(-5))).toBe(BigInt(-5));
+    });
+
+    it('neutralizes the string "-5" but not the number -5', () => {
+      expect(neutralizeCsvFormulaValues({ a: -5, b: '-5' })).toEqual({
+        a: -5,
+        b: "'-5",
+      });
+    });
+  });
+
+  describe('recursion reaches every depth', () => {
+    it('reaches a value nested in an object', () => {
+      expect(
+        neutralizeCsvFormulaValues({ id: '001', os: { name: '=1+1' } }),
+      ).toEqual({ id: '001', os: { name: "'=1+1" } });
+    });
+
+    it('reaches a value inside an array', () => {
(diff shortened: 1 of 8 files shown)


Desvelao (approved):

rodrigofez (approved):
image

PR #9138, opened 2026-09-08
Verify user permissions before starting a registration process

asteriscos:
## Description

The CTI registration process can be started by any user with access to the Wazuh dashboard, but completing it requires a privilege on the Wazuh indexer that is only evaluated at the end of the flow — after the environment has already been created on the CTI side. A user without that privilege is therefore invited into a process that cannot finish, leaving an incomplete registration.

This pull request checks the current user's permissions when the registration modal opens and, when the user cannot register, replaces the invitation with a warning instead of offering the action.

## Proposed Changes

- New read-only endpoint `GET /api/wazuh-check-updates/cti-registration/permission`, which asks Content Manager whether the current user holds the privilege required to register a CTI subscription. The probe sends no body, has no side effects, and always answers `200` — callers branch on `accessAllowed`.
- The registration modal runs the check on open, and only when the environment is not already registered. While the check is in flight the modal shows its existing loading state, so no action is offered before the answer arrives.
- When the user is not allowed to register, the modal renders a warning callout naming the missing privileges and the **Register** button is not shown. Every other modal state — in progress, registered, failed — is unchanged.
- The check fails open: an unreadable answer, a transport failure, or a Wazuh indexer that does not support the probe resolves to "allowed" and logs a warning, so registration keeps working exactly as before on those environments. The Wazuh indexer remains the component that authorizes the operation.

### Artifacts Affected

`wazuh-check-updates` plugin only. No packaging or default configuration files.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

24 unit tests, colocated with their sources:

| Scope | Coverage |
|---|---|
| Server service | Probe shape (no body, query parameter), allowed, denied with missing privileges, non-string filtering, four fail-open paths |
| Server route | `200` for allowed, `200` — not `403` — for denied, `200` when the probe fails |
| Public service | Route called, denial passed through, malformed payload normalized, fail-open on rejection |
| Hook | Skipped when disabled, re-probes on reopen, resets when disabled, ignores a late resolution |
| Modal | Warning callout shown, `Register` absent, missing privileges listed, no action while in flight, probe skipped when already registered |

Full plugin suite: 128 tests, 0 failures.

## Review Checklist

- [ ] Code changes reviewed


rodrigofez (commented):
🟢 Test

Privileged user - admin

image

```json
{"accessAllowed":true,"missingPrivileges":[]}
```

Restricted user
internalusers/restricted
roles: kibana_user, readall

image

```json
{"accessAllowed":false,
 "missingPrivileges":[
   "cluster:admin/content_manager/subscription/create"]}
```

rodrigofez (approved):

PR #9140, opened 2026-09-08
fix(main): mark match_only_text fields non-aggregatable in known-fields generator

### LINKED ISSUE #9135: Adapt dashboards to new field types
# Description

According to https://github.com/wazuh/wazuh-indexer-plugins/issues/1529 some field types will change from the WCS to the ECS. This may have several implications in the Wazuh dashboard. We need to assess and adapt any related changes.

### Source issue

- https://github.com/wazuh/wazuh-indexer-plugins/issues/1529

## Tasks

- [x] Adapt visualization and index-pattern definitions to the new field types.

Desvelao:
## Description

Following up on the dashboard-side analysis in #9135 (itself tracking [wazuh-indexer-plugins#1529](https://github.com/wazuh/wazuh-indexer-plugins/issues/1529), which moves the indexer's `message` field from `keyword` to ECS's `match_only_text`).

The dashboard doesn't commit its index-pattern field definitions — `plugins/main/common/known-fields/*.json` (used to seed `index-pattern` saved objects, e.g. `wazuh-alerts-*`, when no backing index exists yet) are generated by `plugins/main/scripts/generate-known-fields/generate-known-fields.js`, which runs on `postinstall` and fetches live index templates from `wazuh-indexer-plugins`.

That generator's `isAggregatable()` and `shouldReadFromDocValues()` helpers only special-cased ES type `text`, not `match_only_text`. Since `match_only_text` has no doc-values and isn't aggregatable by default (same as `text`), any WCS field of that type was mis-flagged `aggregatable: true` / `readFromDocValues: true`. This was already wrong today for existing `match_only_text` fields (`file.diff`, `wazuh.case.title`, `wazuh.case.description`, `wazuh.case.comments.comment`), and would also mis-flag `message` once the indexer ships the #1529 fix.

**Note on the sort-gating fix**: every Discover-style table's `mapToDataGridColumn()` (`data-grid-service.ts`) computes `isSortable: field.aggregatable` generically, so a non-aggregatable field never offers a sort control (OpenSearch rejects sorting on such fields). That generic fix — and its Jest coverage — already landed on `5.0.0` (via #9149) before this branch was rebased here, so it isn't part of this PR's diff. It's the reason `wazuh.case.title`'s old hand-written `isSortable: false` workaround (with a comment asking for a generic fix) is now dead code, removed below.

Closes #9135

## Changes

- `isAggregatable()` / `shouldReadFromDocValues()` in `generate-known-fields.js` now treat `match_only_text` the same as `text`.
- Added an explicit `match_only_text: 'string'` entry to `mapIndexFieldType()`'s type map (parity/clarity only — the fallback already produced `'string'`).
- Removed the now-redundant manual `isSortable: false` + workaround comment on `wazuh.case.title` in `case-management/cases/table-columns.ts` — `5.0.0` already computes its sortability generically from `field.aggregatable`, so the override was dead code (no behavior change: the field is non-aggregatable either way).
- CHANGELOG entry added under `### Fixed`, crediting both this fix and the already-shipped (but previously undocumented) sort-gating behavior to #9135.

## How to Test

The known-fields generator script sits outside the plugin's Jest/lint/format globs (`scripts/` isn't included in `test/jest/config.js` roots nor the `lint`/`format` package.json globs), so it isn't exercised by `yarn test:jest`. Verified manually:

```js
// with the fix applied
isAggregatable({}, 'match_only_text')       // false (was: true)
shouldReadFromDocValues({}, 'match_only_text') // false (was: true)
isSearchable({}, 'match_only_text')         // true (unchanged)
isAggregatable({}, 'keyword')               // true (unchanged)
```

This fix only affects freshly generated `known-fields/*.json` — it doesn't touch any committed file, since those are regenerated by `postinstall`/build against live indexer templates, not checked in. To validate against real data, run `node plugins/main/scripts/generate-known-fields/generate-known-fields.js` (needs network access to `wazuh-indexer-plugins`) and check that `file.diff` in the regenerated `plugins/main/common/known-fields/events-security.json` now has `"aggregatable": false, "readFromDocValues": false`.

For `table-columns.ts`: confirm removing the manual override doesn't change behavior — `wazuh.case.title` should still show no sort control in **Case Management → Cases**, now purely because it's non-aggregatable, not because of the removed hand-written override. This is already covered by pre-existing (not new) Jest coverage:

```bash
yarn test:jest --testPathPattern=data-grid-service
```

## Results and Evidence

No user-visible UI change: the sort-gating behavior this PR's CHANGELOG entry documents is already live on `5.0.0`, and the generator fix only affects known-fields JSON regenerated at build/`postinstall` time — nothing already committed changes today. `wazuh.case.title`'s column keeps showing no sort control before and after this PR, now for a purely generic reason instead of a per-module override, so no screenshot applies here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Desvelao:
## Manual test plan: reproduce the bug on `5.0.0`, confirm the fix here

Full end-to-end procedure against a real dev environment (index-pattern lifecycle + a live cluster), complementing the script-level checks above.

All commands run from the repo root; this assumes the dev container mounts this checkout and the indexer/OSD are reachable at `https://localhost:9200` / `https://localhost:5601` (`admin:admin`).

**1. Checkout the pre-fix branch:**
```bash
git checkout 5.0.0
```

**2. Remove the current known fields and regenerate them from this (buggy) branch:**
```bash
cd plugins/main
rm -f common/known-fields/*.json
node scripts/generate-known-fields/generate-known-fields.js
cd ../..
```
Sanity-check the bug is baked into the file:
```bash
node -e "console.log(require('./plugins/main/common/known-fields/findings.json').find(f=>f.name==='wazuh.case.title'))"
# -> aggregatable: true, readFromDocValues: true  (wrong)
```

**3. Environment is already up; delete the affected index pattern's saved object** so the server has to recreate it from the known-fields fallback:
- UI: Stack Management → Index Patterns → `wazuh-findings-v5*` → trash icon → Delete, **or**
```bash
curl -sk -u admin:admin -X DELETE \
  'https://localhost:5601/api/saved_objects/index-pattern/wazuh-findings-v5*' \
  -H 'osd-xsrf: true'
```

**4. Restart the server to force `ensureIndexPatternExistence` to recreate it:** in the container terminal running `yarn start --no-base-path`, `Ctrl+C` then run it again. Confirm the recreated saved object reproduces the bug:
```bash
curl -sk -u admin:admin 'https://localhost:5601/api/saved_objects/index-pattern/wazuh-findings-v5*' \
  | python3 -c "import json,sys; d=json.load(sys.stdin); f=json.loads(d['attributes']['fields']); print(next(x for x in f if x['name']=='wazuh.case.title'))"
# -> aggregatable: true  <- the bug: OSD will offer this text field for Visualize/terms-agg
```

**5. Change the branch to the fix, and regenerate known fields from it:**
```bash
git checkout fix/9135-known-fields-match-only-text
cd plugins/main
rm -f common/known-fields/*.json
node scripts/generate-known-fields/generate-known-fields.js --branch 5.0.0
cd ../..
```
(`--branch 5.0.0` because this branch's own `package.json` version is `5.1.0`, which doesn't exist as a `wazuh-indexer-plugins` branch — `5.0.0` matches what this dev cluster actually runs.) Repeat step 3-4 (delete + restart) once more so the index pattern is recreated from the now-corrected known fields, and confirm `aggregatable: false` this time, straight from the seed - before touching any real data.

**6. Run the verifier script:**
```bash
node plugins/main/scripts/generate-known-fields/verify-field-capabilities.js
```
Expect ✅ for `wazuh.case.title` and `group.description` (their mapping exists even with 0 documents - they're declared in the template's explicit `properties`, not only a dynamic template); `file.diff` and the other two case fields will show ⚠️ "not present in live mapping yet" until step 7.

**7. Create a case and refresh the index pattern:** open the Case Management module, create a new case (populates `wazuh.case.title`/`wazuh.case.description`) and add a comment (populates `wazuh.case.comments.comment`). Then Stack Management → Index Patterns → `wazuh-findings-v5*` → **Refresh field list**. Confirm in the UI (field list "Aggregatable"/"Searchable" columns) and by re-running the verifier script that all fields now report the correct, non-aggregatable value - showing that both the seeded default (this fix) and OSD's live refresh (already-correct core behavior) agree.

Ripdiegozz (commented):
🟢 Tests

- **`file.diff` (match_only_text, non-aggregatable): sort disabled**

Image

- **`wazuh.agent.name` (aggregatable): sort enabled**

Image

Unit tests pass (5/5, `data-grid-service.test.ts`).

**known-fields regenerated against a live 5.0.0 indexer**

`file.diff` now comes out non-aggregatable, no doc_values:

```json
{
  "name": "file.diff",
  "type": "string",
  "esTypes": ["match_only_text"],
  "searchable": true,
  "aggregatable": false,
  "readFromDocValues": false
}
```

yenienserrano (dismissed):
LGTM :green_circle:

yenienserrano (approved):
LGTM :green_circle:

Ripdiegozz (approved):

PR #9175, opened 2026-09-15
Fix empty view when switching tabs with sub tabs

### LINKED ISSUE #9174: IT Hygiene tabs with sub-tabs do not render a default sub-tab
| 5.0.0 | 09   | Chrome                       |

**Description**

In IT Hygiene, moving from a main tab that has sub-tabs to another main tab that also has sub-tabs leaves the view empty: the sub-tab bar is rendered, but no sub-tab is selected and no content is loaded below it. The view stays blank until the user clicks a sub-tab manually.

Main tabs with sub-tabs:

- System: OS, Hardware
- Software: Packages, Windows KBs, Browser extensions
- Network: Addresses, Interfaces, Protocols, Listeners, Traffic
- Identity: Users, Groups

Dashboard, Processes and Services have no sub-tabs and are not affected.

Coming from a tab without sub-tabs (for example Dashboard -> System) works correctly: the first sub-tab is selected and rendered.

**Preconditions**

1. A Wazuh dashboard with the IT Hygiene app available and inventory data indexed.

**Steps to reproduce**

1. Navigate to IT Hygiene (`#/overview/?tab=it-hygiene&tabView=dashboard`).
2. Click on the `System` tab. The `OS` sub-tab is selected and its content is rendered (URL: `...&tabView=system&tabSubView=os`).
3. Click on the `Software` tab.

**Expected Result**

1. `Software` selects its first sub-tab (`Packages`) by default and renders its content, the same way `System` does when opened from `Dashboard`.

**Actual Result**

1. The `Software` sub-tab bar is rendered with no sub-tab selected and no content below it. The same happens for any other jump between tabs with sub-tabs (System -> Network, Network -> Identity, etc.). The URL keeps the previous tab's `tabSubView` value (for example `...&tabView=software&tabSubView=os`).

**Screenshots**

Screen recording: navigating `Dashboard` -> `System` (works, `OS` is selected) and then to `Software` / `Network` / `Identity` (blank, no sub-tab selected).

https://github.com/user-attachments/assets/3f2b61e5-effd-46af-a3b4-37189eeee90e

**Additional context**

Sub-tab selection is driven by the `tabSubView` URL search parameter:

- `plugins/main/public/components/common/tabs/module-sub-tabs.tsx` renders `TabsManagedBySearchParam` with `searchParamNavigation='tabSubView'`.
- `plugins/main/public/components/navigation/tabs-managed-by-search-params.tsx` matches `tabSubView` against the sub-tab ids and falls back to a `Redirect` to the first sub-tab.
- `NavigationService.switchSubTab` (`plugins/main/public/react-services/navigation-service.tsx`) only updates the `tabView` parameter when the main tab changes, so a `tabSubView` value belonging to the previous tab (for example `os` while standing on `Software`) stays in the URL and matches none of the current sub-tabs.

That explains why the first navigation works (there is no `tabSubView` yet, so the fallback applies) and the following ones do not (a stale, non-matching `tabSubView` is present).

The same problem was already worked around locally for Regulatory Compliance in `plugins/main/public/components/overview/regulatory-compliance/shared/compliance-module.tsx`, which validates `tabSubView` against its own tab list and rewrites it when it does not match. IT Hygiene has no equivalent handling.

The fix should be applied once in the shared navigation layer (`TabsManagedBySearchParam` / `switchSubTab`) instead of per module, so every module with sub-tabs behaves the same and the Regulatory Compliance workaround can be removed.

yenienserrano:
## Description

Modules whose tabs have sub tabs (IT Hygiene System, Software, Network and Identity, FIM Inventory, and every Regulatory Compliance module) rendered an empty view when moving from one tab with sub tabs to another. The sub tab bar was drawn with nothing selected and no content below it, and the URL kept the previous tab's `tabSubView` value (for example `...&tabView=software&tabSubView=os`).

Sub tab selection is driven by the `tabSubView` search parameter. `TabsManagedBySearchParam` matched it against the sub tab ids and, when nothing matched, relied on a `Redirect` component that navigates from a mount-only `useEffect`. That navigation happens before `useRouterSearch` subscribes to the history, so the pushed URL never reached the component state: the tabs stayed unselected and no route matched, leaving the view blank. Coming from a tab without sub tabs worked only by accident, because `tabSubView` was absent.

The fix moves the fallback out of navigation and into rendering: the component now resolves the selected tab from the search parameter with the first tab as fallback, so a missing or foreign `tabSubView` can never produce an empty view, and the URL is synced afterwards with a `replace` so the redirect does not add a history entry.

`NavigationService.switchSubTab` was left untouched: it owns the `tabView` parameter for the main module tabs and is not what selects a sub tab.

Closes #9174

## Proposed Changes

- `plugins/main/public/components/navigation/tabs-managed-by-search-params.tsx`
  - Resolve the selected tab as `tabs.find(tab => tab.id === search[searchParamNavigation]) ?? tabs[0]` and render its component directly, instead of routing through `Switch`/`Route`/`Redirect` from `components/router-search`.
  - Drive `isSelected` from the resolved tab, so the first tab is highlighted even before the URL is normalized.
  - Sync the search parameter with the rendered tab in a `useEffect` using `NavigationService.updateAndNavigateSearchParams(..., { replace: true })`, which reads the live URL parameters instead of a possibly stale snapshot and does not push a new history entry for a fallback.
  - Type `tabs[].component` as `React.ComponentType` and `tabsProps` as `React.ComponentProps<typeof EuiTabs>`, replacing the `any`s the lint rules reject.
- `plugins/main/public/components/overview/regulatory-compliance/shared/compliance-module.tsx`
  - Removed the per-module workaround that validated `tabSubView` and re-navigated, now that the shared layer handles it. The `moduleId` prop it needed is gone, and its callers were updated.

### Results and Evidence

https://github.com/user-attachments/assets/457db474-ab9e-4743-a376-51f4c84c64bb

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

Added `plugins/main/public/components/navigation/tabs-managed-by-search-params.test.tsx`, covering:

- The tab matching the search parameter is rendered.
- The first tab is rendered and the URL is completed when the search parameter is missing.
- The first tab is rendered and the URL is rewritten when the search parameter belongs to another module (the reported bug).
- The fallback replaces the current history entry instead of pushing a new one.
- Clicking a tab navigates and renders it.

## Review Checklist

- [x] Code changes reviewed


rodrigofez (approved):
🟢 Test

Before

https://github.com/user-attachments/assets/5fa5f837-5d29-4124-a92a-7886e6c45bd4

After

https://github.com/user-attachments/assets/83691602-c221-48f1-b580-6feac929a700

asteriscos (approved):
🟢 

https://github.com/user-attachments/assets/7af57824-53bd-4080-a6f8-f6fbdd5ac79a

PR #9178, opened 2026-09-16
Improvements in token enrollment

### LINKED ISSUE #9145: Dashboard: wizard mints tokens via the manager API, and defaults to a verifying manager-API link
## Description

`wazuh/wazuh` is adding, in a PR that also adds the manager's own certificate work, a REST mint
endpoint for enrollment tokens (`POST /agents/enrollment-tokens`, RBAC action
`enrollment_token:create`, `request_type='local_master'`) that mirrors
`wazuh-manager-authd --create-enrollment-token --address <host>`: it validates the host against the
listener certificate's SAN, reads port and `global_prefix` from the running listener, and returns the
packed `ver`/`adr`/`pin`/`key` token. `wazuh-dashboard-plugins` is a separate repository this plan
cannot read directly, but the Deploy-new-agent wizard is the natural caller of that endpoint, and
today it does not call anything like it — the wizard builds its own enrollment command locally,
without asking the manager whether the typed host is even covered by the certificate.

## Expected behaviour

The Deploy-new-agent wizard passes the operator-typed host straight to the manager's mint endpoint
and shows whatever the manager returns — including a refusal when the host is not in the listener's
SAN — instead of assembling `adr`/port/prefix itself and hoping the resulting command works. 

## Proposed fix

Wire the wizard's "generate enrollment command" step to `POST /agents/enrollment-tokens` (or
whichever manager REST surface the mint PR lands as), passing the host the operator typed and the
desired TTL; render the manager's own error message when the host is refused (not a second,
locally-computed SAN check that can drift from what the manager actually enforces).

## References

- `api/api/configuration.py:44,206-240` (current self-signed `apid.pem`, `CN`/SAN `localhost`)
- `src/remoted/remoted_module/src/http_server/RestinioHttpServer.cpp:522-585` (today's startup-only expiry check; no metric yet)
- `docs/ref/modules/remoted/metrics.md:392,471` (existing `GET /cluster/{node_id}/daemons/stats` shape the new metric will use)
- wazuh-dashboard-plugins — Deploy-new-agent wizard, `wazuh.yml` `hosts:`/`ca:` handling — not verifiable from this tree
- Design document `https-agent-tls-verification-solution.md` §7, items 12 (wizard part), 18, 20, and the interface half of item 8/D33
- Endpoint specification: https://github.com/wazuh/wazuh/pull/39040/changes#diff-16ecaa11b00bf069585c4276e961223a8256b87800556f61c667021e2dd95915

asteriscos:
## Description

Follow-up UI work on the enrollment token feature added in #9145. Three rough
edges surfaced while using the minted-token screens:

- The **Create enrollment token** flyout discarded a partially filled form
  without asking — a stray click outside it, or on the close button, threw the
  work away silently.
- In the **Enrollment token details** flyout, the description was clipped to a
  single line with an ellipsis and only readable through a tooltip, even though
  it is prose with spaces to wrap on.
- In the **Deploy new agent** wizard, the Enrollment token step showed both ways
  of obtaining a token at once — the paste field and the full mint form — with
  whichever one was not in use greyed out. The two exclude each other, so the
  step was roughly twice as long as it needed to be.

No server, API or configuration changes. Presentation and step-local state only.

Related to #9145.

## Proposed Changes

**Confirmation before discarding an unsubmitted token form**

The **Create enrollment token** flyout now goes through the app's existing
`UnsavedChangesGuardedFlyout` (already used by the Discover document-details
flyout and the Case Management tab) instead of `WzFlyout`. All three exits — the
header X, a click outside, and the footer Cancel button — ask for confirmation
while any field has moved off the value it was offered with, or either switch is
on. Once the token is minted the prompt stops: the form is gone and there is
nothing left to lose.

The flyout is split into a wrapper holding `mintedToken` and a content component
below it, because only a descendant of the guarded flyout can report unsaved
changes to it.

**Readable description in the details flyout**

The Description value no longer uses `TruncatedValueTooltip`. That control exists
for the id and the address, which are opaque strings with no spaces to break on;
a description wraps over as many lines as it needs and is justified.
`overflow-wrap: break-word` only breaks inside a word when that one word is wider
than the flyout, so line breaks stay on word boundaries otherwise.

**Token source sub-tabs in the wizard step**

The Enrollment token step now presents **Generate a new token** / **Use an
existing token** as sub-tabs, following the pattern used in the public Wazuh
documentation. Only the open tab renders, which removes the paste field, its
help text, the horizontal rule and the "Or generate a new token:" caption from
the generate path, and everything else from the reuse path. The default state is
the subtitle, the tab strip, the advanced-options link and the button.

The cross-disabling machinery this replaces (`isReusingToken`,
`tokenRequestWasStarted`, and a `disabled` prop on each of the six controls) is
deleted. Three behaviours are worth a reviewer's attention:

- **The open tab owns the token** the deployment command is built from. A token
  committed on the other tab is handed back up as `null` when leaving it, so the
  command is never built from something no longer on screen.
- **A minted token survives a round trip.** It is held in step state and
  re-committed when the generate tab comes back — the server returns the token
  text once, so switching tabs must not cost a second mint.
- **Typed values are kept across a tab change, except ones that fail
  validation, which are cleared.** `fieldIsNotApplicable` feeds a banner that
  names invalid fields and withholds the deployment commands until they are
  clean. With both inputs always rendered this could not strand anyone; with
  tabs, an invalid value left on the hidden tab would block the wizard with no
  visible field to correct. Clearing only the broken field closes that hole
  locally instead of plumbing the tab state down into the status service.

The step subtitle was updated: it no longer says the reuse path lives behind the
advanced options.

### Results and Evidence

image

image

image

### Artifacts Affected

`plugins/main` only:

- `public/components/enrollment-tokens/components/create-enrollment-token-flyout.tsx`
- `public/components/enrollment-tokens/components/enrollment-token-details-flyout.tsx`
- `public/components/enrollment-tokens/components/enrollment-token-details-flyout.scss` (new)
- `public/components/endpoints-summary/register-agent/components/enrollment-token/enrollment-token.tsx`
- `public/components/endpoints-summary/register-agent/utils/register-agent-data.tsx`

### Configuration Changes

None.

### Documentation Updates

None required. No new settings, endpoints or CLI surface.

### Tests Introduced

All colocated with the source. 24 tests added or rewritten:

- `create-enrollment-token-flyout.test.tsx` — 8 new: pristine close goes
  straight through; a dirty field and a switch-only change each prompt;
  declining keeps the form; accepting discards; the footer button is guarded
  too; no prompt after minting; the minted values render stacked.
- `enrollment-token-details-flyout.test.tsx` (new file) — 3: the whole
  description renders unwrapped by the truncating control, the id and address
  keep the one-line treatment, and a missing description still falls back to a
  dash.
- `enrollment-token.test.tsx` — rewritten, 9 new about the tabs: which tab opens
  and when, what each tab renders, values kept vs. invalid values cleared on
  switch, the token dropped on the way out and restored on the way back. These
  now render through a small stateful harness that holds the token the way the
  wizard does, which is what makes the drop/restore observable.

Suite results: `register-agent` 188 passed / 17 suites; `endpoints-summary` and
`enrollment-tokens` together 349 passed / 50 suites. Lint clean.

## Review Checklist

- [ ] Code changes reviewed
- [ ] Tests cover the new functionality
- [x] Correct labels applied — `no-changelog`


rodrigofez (approved):
🟢 Test
Screenshot 2026-09-16 at 10 56 20 AM
Screenshot 2026-09-16 at 10 56 34 AM

https://github.com/user-attachments/assets/4f0a906e-8ce5-4fa9-a998-06350c725db1

Ripdiegozz (approved):
### 🟢 Tests

**Confirmation before discarding an unsubmitted token form**

https://github.com/user-attachments/assets/45631d5b-8434-43f0-8d4b-eb40dec2cd42

**Token source sub-tabs in the Enrollment token step**

https://github.com/user-attachments/assets/32ad7756-9de6-467a-a2bf-45589ab5592d

PR #9179, opened 2026-09-16
Fix agent Summary crash when pinning Incident Response

### LINKED ISSUE #9177: Pinning Incident Response in the agent menu breaks the Summary page
**Description**

The agent Summary page shows a row of application shortcuts followed by a
`More...` menu with every application in it. Hovering an entry in that menu
reveals a pin button that adds the application to the shortcut row.

Pinning **Incident Response** replaces the whole page with the "Something went
wrong." error screen:

```
TypeError: Cannot read properties of undefined (reading 'appId')
    at AgentsWelcome (.../bundles/plugin/wazuh/wazuh.chunk.1.js)
```

The new pinned list is written to `localStorage` under
`wz-menu-agent-apps-pinned` before the render fails, so the error survives a
reload. Since the error screen replaces the menu, there is no way to unpin from
the UI — the key has to be removed from browser storage by hand.

Every other application in the menu pins fine. Incident Response is the only
one that fails.

**Preconditions**

1. At least one registered agent.
2. Fewer than 6 applications already pinned — the pin button is hidden once the
   limit is reached.

**Steps to reproduce**

1. Navigate to `/app/endpoints-summary#/agents?tab=welcome&agent=001`.
2. Click `More...`.
3. Hover **Incident Response** under *Security operations*.
4. Click the pin button at the right of the row.

**Expected Result**

1. Incident Response is added to the shortcut row next to the `More...` menu,
   the same as any other application.

**Actual Result**

1. The page is replaced by "Something went wrong." with
   `TypeError: Cannot read properties of undefined (reading 'appId')`.
2. Reloading the page shows the same error, because the pin was already saved.
   Recovering means deleting `wz-menu-agent-apps-pinned` from browser storage.

**Screenshots**

https://github.com/user-attachments/assets/0284ce00-1250-4b6d-8b3a-f5855c5565fd

**Additional context**

`renderModules()` in
`plugins/main/public/components/common/welcome/agents-welcome.js` looks each
pinned application id up in `WAZUH_MODULES`:

```js
const moduleID = Object.keys(WAZUH_MODULES).find(
  key => WAZUH_MODULES[key]?.appId === applicationId,
).appId;
```

`incident-response-dashboard` is the only id exposed in the agent menu that has
no matching `appId` in `plugins/main/common/wazuh-modules.ts`, so `.find()`
returns `undefined` and reading `.appId` from it throws.

Two things are wrong in those three lines:

- An application id with no entry in `WAZUH_MODULES` is not handled at all.
- `Object.keys(...).find(...)` returns a key string, not the module object, so
  `.appId` is `undefined` even when a match is found. `moduleID` is therefore
  always `undefined`, and the `hasAgentSupportModule(agent, moduleID)` check on
  the next line always passes. The operating-system support filter on the
  shortcut row has never actually filtered anything.

rodrigofez:
## Description

Pinning **Incident Response** in the agent Summary applications menu replaced the whole
page with the "Something went wrong." error screen:

```
TypeError: Cannot read properties of undefined (reading 'appId')
```

The pinned list was persisted to `localStorage` before the render failed, so the error
survived a reload and the only recovery was deleting `wz-menu-agent-apps-pinned` by hand.

Closes #9177

## Proposed Changes

`renderModules()` in `agents-welcome.js` translated every pinned application id into a
legacy `WAZUH_MODULES` module id:

```js
const moduleID = Object.keys(WAZUH_MODULES).find(
  key => WAZUH_MODULES[key]?.appId === applicationId,
).appId;
```

That was wrong twice. `Object.keys(...).find(...)` yields a key string, so `.appId` was
always `undefined` and the `hasAgentSupportModule(agent, moduleID)` check below never
filtered anything. And an application with no `WAZUH_MODULES` entry made `.find()` return
`undefined`, so reading `.appId` threw — `incident-response-dashboard` is the only id in
the agent menu in that situation.

- Added `getAgentPinnedApplications()` in
  `public/components/common/welcome/utils/pinned-applications.ts`: it resolves pinned ids
  against `Applications`, discards ids with no matching application, discards applications
  unsupported by the agent operating system, and limits the result to the shortcuts that
  fit in the header.
- `renderModules()` now uses that helper and the `WAZUH_MODULES` indirection is gone. The
  shortcut row filters by **application id**, the same identifier the applications menu
  (`menu-agent.js`) already passes to `hasAgentSupportModule`, so both stay consistent.
- The operating system support filter on the shortcut row now actually applies.
- Shortcut items are keyed by application id instead of the array index.

A browser that already stored the failing pinned list recovers on its own: the id is valid,
it simply renders now.

### Results and Evidence

https://github.com/user-attachments/assets/369584a4-ec30-4afd-a89f-2ae357cd5564

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

Colocated unit tests in
`public/components/common/welcome/utils/pinned-applications.test.ts` covering:

- pinned applications returned in the pinned order;
- Incident Response, the application with no `WAZUH_MODULES` entry, resolving instead of
  throwing;
- identifiers with no matching application being discarded;
- applications unsupported by the agent operating system (Docker on Windows, Vulnerability
  Detection on SunOS) being discarded;
- the maximum shortcut limit;
- empty pinned list and undefined agent.

### How to Test

1. Navigate to `/app/endpoints-summary#/agents?tab=welcome&agent=001`.
2. Click `More...`.
3. Hover **Incident Response** under *Security operations* and click the pin button.
4. Incident Response is added to the shortcut row and the page keeps rendering.
5. Reload the page: the shortcut is still there and the page loads normally.

## Review Checklist

- [x] Code changes reviewed


asteriscos (approved):
CR: 🟢 
Test: 🟢 

https://github.com/user-attachments/assets/4c7b7232-1ae6-4d4e-9665-cee8bc36c33b

yenienserrano (approved):
LGTM :green_circle:

https://github.com/user-attachments/assets/5e1e85df-3fc9-4ccc-a484-c1859706ee72

PR #9185, opened 2026-09-17
Improve agent view menu pins

### LINKED ISSUE #9183: Improve agent view menu pins
# Description

Currently, the agent view allows to pin modules to the top menu, but the UX can be improved. It's not clear at first sight when a module is pinned, and moreover, there's a limit on pinned modules, and it's also hard to understand its state and when the limit has been reached.

https://github.com/user-attachments/assets/7b0c78df-9929-4199-b869-793bab3aa500

## Tasks

- [ ] Improve the UX of the pin module feature
  - The user should know which modules are pinned without hovering
  - Improve the way to display when the pinned elements limit has been reached

rodrigofez:
## Description

In the agent view you could not tell which applications were pinned without hovering each row, and once the pin limit was reached the pin control just disappeared with no explanation. The header also hid pinned shortcuts even when there was room for them.

Closes #9183

## Proposed Changes

- The pin toggle is always visible, so pinned applications stand out at a glance.
- At the pin limits the toggle is disabled and its tooltip says why. This replaces a toast that could never appear, because the icon was hidden in exactly the case that would have triggered it.
- The header now shows as many pinned shortcuts as fit. It used to pick the amount from a table of window widths, which was too conservative: at 1280px it showed 2 where 4 fit.
- The pin limit goes from 6 to 5, which is all the header could ever display.
- `menu-agent.js` is now `menu-agent.tsx`, a function component.
- The `resize` listener is removed on unmount; it was never cleaned up.

### Results and Evidence

https://github.com/user-attachments/assets/57ca063d-4a11-4f68-884c-009488540b65

### Artifacts Affected

None.

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `menu-agent.test.tsx`: pin state visible without hovering, pin and unpin, toggles disabled at the maximum and at the minimum.
- `pinned-applications.test.ts`: the limits, the stored list sanitizer and the pin toggle.

## How to Test

1. Open an agent from Endpoints Summary and click **More...**.
2. Check the pinned applications show a filled pin without hovering them.
3. Pin applications until you reach 5. The remaining pins are disabled and their tooltip explains the limit.
4. Unpin down to one. The last pin is disabled and its tooltip explains why.
5. Resize the window and check the header shows as many pinned shortcuts as fit.

## Review Checklist

- [ ] Code changes reviewed


asteriscos (approved):
CR: 🟢 
Test: 🟢 

https://github.com/user-attachments/assets/15fd4fb5-c8a9-4d7a-a82c-a7d33eab82cb

yenienserrano (approved):
LGTM :green_circle:

https://github.com/user-attachments/assets/ae47c0a1-553a-4032-98fd-c0e01accaf79

PR #9187, opened 2026-09-18
fix: EuiBasicTable/EuiButtonGroup prop-type warnings

### LINKED ISSUE #9184: EuiBasicTable/EuiButtonGroup prop-type warnings on Endpoints Summary, Endpoint Groups and Vulnerability Detection
| Wazuh   | Rev     | Browser |
|---------|---------|---------|
| 5.0.0   | 5.0.0-09 | Chrome |

**Description**
While driving the running dev dashboard end-to-end with an automated browser (Playwright) and capturing the DevTools console per page — not just from a static code read — two `main` plugin pages throw React prop-type warnings from EUI components. Both are confirmed live in the browser console.

**Preconditions**
1. Dev environment up (`docker/osd-dev`), logged in as `admin`.

**Steps to reproduce**

*Issue A — `itemIdToExpandedRowMap` boolean instead of object*
1. Navigate to Endpoints Summary (`/app/endpoints-summary`).
2. Open the browser console.
3. Observe: `Warning: Failed prop type: Invalid prop `itemIdToExpandedRowMap` of type `boolean` supplied to `EuiBasicTable`, expected `object`.`
4. Repeat on Endpoint Groups (`/app/endpoint-groups`) — same warning appears.

*Issue B — missing required `legend` on `EuiButtonGroup`*
1. Navigate to Vulnerability Detection (`/app/vulnerability-detection`).
2. Open the browser console.
3. Observe: `Warning: Failed prop type: The prop `legend` is marked as required in `EuiButtonGroup`, but its value is `undefined`.` — thrown by the "Evaluated / Under evaluation" toggle button group.

**Expected Result**
1. No React prop-type warnings in the console for either page.
2. `EuiBasicTable`'s expandable-row state (`itemIdToExpandedRowMap`) is a proper `{ [id]: ReactNode }` map, so per-row expand/collapse works reliably for every agent/group row.
3. `EuiButtonGroup` receives an accessible `legend` string.

**Actual Result**
1. Endpoints Summary (`plugins/main/public/components/endpoints-summary/`) and Endpoint Groups (`plugins/main/public/controllers/management/components/management/groups/`) pass a `boolean` where `EuiBasicTable` expects an id→row map for `itemIdToExpandedRowMap`. Likely a state variable that should track which row ids are expanded got collapsed into a single boolean toggle — this can silently break expand/collapse for specific rows rather than failing loudly.
2. Vulnerability Detection's "Evaluated / Under evaluation" toggle (`plugins/main/public/components/overview/vulnerabilities/`) renders an `EuiButtonGroup` without the required `legend` prop (screen-reader accessible label) — a one-line fix.

**Screenshots**
N/A — console-only warnings, no visual regression observed.

**Additional context**
- Both confirmed live via browser console during a Playwright-driven pass through every registered app in `plugins/main`, `wazuh-check-updates` and `wazuh-ai-assistant`, not inferred from source alone.
- Related task for this same session:
  - [x] Do a broader pass reviewing browser console output across the various registered apps/modules in `plugins/main` (and the other plugins) — several recurring React/OSD-framework deprecation warnings showed up across many pages in this session (legacy `childContextTypes`/`contextTypes` API usage, `ReactDOM.render` deprecation, `defaultProps` deprecation on function components, `findDOMNode` deprecation, `componentWillReceiveProps` rename warning, missing `key` prop on list children). Most look like framework-wide noise rather than page-specific bugs, but worth triaging which are cheap wins to silence vs. which need a version bump or a different EUI usage pattern.

    **Triage result** (static codebase audit across all four plugins — `plugins/main`, `wazuh-core`, `wazuh-check-updates`, `wazuh-ai-assistant`):

    | Warning category | Result |
    |---|---|
    | Legacy Context API (`childContextTypes`/`contextTypes`/`getChildContext`) | None found in first-party code. No action needed. |
    | `findDOMNode` | None found in first-party code. No action needed. |
    | `componentWillReceiveProps` (unsafe form) | None found — the one lifecycle hit is already correctly renamed to `UNSAFE_componentWillReceiveProps`. No action needed. |
    | Missing `key` prop on list children | Scanned ~393 `.map()` call sites across all four plugins; zero genuine misses found via static analysis — **update:** continued live browser testing afterward did find one real instance the static pass missed (see below). |
    | `ReactDOM.render` (deprecated in React 18) | 3 first-party call sites in `plugins/main/public/` — fixed. |
    | `defaultProps` on function components (deprecated in React 18.3+) | 2 first-party call sites in `plugins/main/public/components/common/tables/` — fixed. |

    `wazuh-core`, `wazuh-check-updates`, and `wazuh-ai-assistant` had zero first-party hits for any category. The two real, fixable categories (both confined to `plugins/main/public/`) were resolved in PR #9187, following the `react-dom/client` `createRoot` pattern already used in `wazuh-ai-assistant`.

    **Further live-browser findings (post-audit):** continued interactive testing surfaced six more warnings — three are exactly the kind of runtime-only case the static `key`-prop audit above had already flagged as a blind spot, the other three are new categories outside the original six:
    - Missing `key` prop in `InventoryDashboardTable`'s dashboard-panel list (`plugins/main/public/components/common/dashboards/inventory.tsx`) — a two-branch `if`/`return` in the `.map()` callback only keyed one branch; every current caller (IT Hygiene, SCA, FIM inventories, case management) took the unkeyed branch. Fixed in PR #9187.
    - Invalid `<div>` cannot appear as a descendant of `<p>` HTML nesting warning in home-overview `StatTile`s (`plugins/main/public/components/common/welcome/home-overview/components/common/stat-tile.tsx`) — `EuiStat`'s title defaults to a `<p>` wrapper, but three callers build their value with `RedirectAppLinks`, which always renders a `<div>`. Fixed in PR #9187 by overriding `titleElement='div'`, matching a pattern already used elsewhere in this codebase.
    - React does not recognize the `setPagination` prop on a DOM element, thrown on Discover's `EuiDataGrid` (via its internal focus-trap wrapper) — `useDataGrid()` (`plugins/main/public/components/common/data-grid/use-data-grid.tsx`) returns hook-only extras (`setPagination`, `dataGridStatePersistenceManager`, `columnsAvailable`) mixed into the same object as real `EuiDataGridProps`, and most consumers spread it unfiltered onto `<EuiDataGrid>`. Investigation found this affects 4 call sites, not just the one reported (`wz-discover.tsx`, `table.tsx`, `wz-data-grid.tsx`, and the Vulnerability Detection inventory) — one of which had a partial, out-of-sync fix already. Fixed in PR #9187 with a single shared constant applied consistently across all four.
    - `` `NaN` is an invalid value for the `height` css style property `` on Discover's `EuiDataGrid`, thrown by its internal virtualized `Grid` (react-window) — the same `useDataGrid()` hook's `rowCount` fell back to `0` only when `results` itself was falsy, not when it lacked a `.hits` property; `wz-discover.tsx` initializes `results` as a truthy `{}` placeholder before its first fetch resolves, so `rowCount` resolved to `Math.min(undefined, ...)` = `NaN`, which `EuiDataGridBody` used to compute its grid's height. Only `wz-discover.tsx` hits this path — it keeps `EuiDataGrid` mounted at all times (hiding it via a CSS class instead of unmounting), unlike its three siblings, which guard/conditionally render the grid until `results.hits.total` is a real number. Fixed in PR #9187 with a proper nullish default.
    - A second missing `key` prop warning, this time in `WzRibbonBody` (cluster management's node configuration overview, `ConfigurationCards`) — a data problem, not a JSX one: the shared `WzRibbonBody`/`WzRibbonItem` component already keys each item from `item.key`, but `ConfigurationCards` built its 7-item list without a `key` field at all, even though the shared `IRibbonItem` type declares it required, so every item's key resolved to `undefined`. The other 3 consumers of this shared ribbon component already supplied a proper key. Fixed in PR #9187 with a stable, kebab-case key derived from each item's label.
    - A third missing `key` prop warning, in the Dev Tools header's `TopNavMenu` — the legacy `EuiTabs` branch keyed the inner `EuiTab`, not the value actually returned from `.map()`; for the one item defining `renderWrapper` (the API Reference link), the returned value is the plain `<a>` `renderWrapper` produces, with the keyed `EuiTab` buried inside it as `children`, so React never saw a key on that list item. Fixed in PR #9187 by moving the key to wrap whichever element is actually returned, via `React.cloneElement`, using the item's already-unique `id`.
- Loosely related repo precedent (different specific warnings, older Wazuh versions, not duplicates): #4092 "Multiple unexpected warning during Wazuh app session" (4.3), #5821 "Multiple unexpected Errors during Wazuh app session" (4.5.2) — both are "known issues" trackers for console noise audits.

Desvelao:
## Description

Closes #9184

A Playwright-driven pass through every registered app in `plugins/main`, capturing
the live browser DevTools console (not inferred from source alone), found two
React prop-type warnings:

- `EuiBasicTable` on Endpoints Summary and Endpoint Groups received
  `itemIdToExpandedRowMap={false}` (a boolean) instead of an object/`undefined`.
- `EuiButtonGroup` on Vulnerability Detection's "Evaluated / Under evaluation"
  toggle was missing the required, screen-reader-facing `legend` prop.

The issue also asked for a broader pass over the browser console across
`plugins/main` (and the other plugins) for recurring React/OSD-framework
deprecation warnings. That triage (posted on the issue) found two more real,
fixable categories, both confined to `plugins/main/public/`: deprecated
`ReactDOM.render` usage (3 call sites) and `defaultProps` on function
components (2 call sites). Legacy context API, `findDOMNode`, and
`componentWillReceiveProps` were all audited and found clean.

Continued live browser testing after that static audit surfaced two more
real, distinct warnings the static pass missed or that fall outside its
original scope:

- A missing `key` prop in `InventoryDashboardTable`'s dashboard-panel list —
  the static `.map()`-pattern audit's "zero genuine misses" conclusion held
  for what it could see, but this bug's two-branch `if`/`return` structure
  was exactly the kind of runtime-only case that audit had already flagged
  as a blind spot.
- An invalid `<div>` cannot appear as a descendant of `<p>` HTML nesting
  warning in home-overview `StatTile`s — a category outside the original
  six enumerated warning types.
- React does not recognize the `setPagination` prop on a DOM element,
  thrown by `EuiDataGrid`'s internal focus-trap wrapper on the Discover
  table — traced to `useDataGrid()` returning hook-only extras
  (`setPagination`, `dataGridStatePersistenceManager`, `columnsAvailable`)
  mixed into the same object as real `EuiDataGridProps`, spread unfiltered
  onto `<EuiDataGrid>`. Found to affect 4 call sites, not just the one
  reported.
- `` `NaN` is an invalid value for the `height` css style property `` on
  Discover's `EuiDataGrid`, thrown by its internal virtualized `Grid` — the
  same `useDataGrid()` hook's `rowCount` resolved to `Math.min(undefined,
  ...)` = `NaN` while `results` was still its initial truthy-but-empty
  placeholder, before the first search resolved.
- A missing `key` prop warning on the cluster management node configuration
  overview (`ConfigurationCards`) — a data problem, not a JSX one: the
  shared `WzRibbonBody` already keys each item from `item.key`, but this
  caller built its 7-item list without a `key` field at all, even though
  the shared type declares it required.
- A missing `key` prop warning on the Dev Tools header (`TopNavMenu`) — the
  legacy `EuiTabs` branch keyed the inner `EuiTab`, not the value actually
  returned from `.map()`, so the one item defining `renderWrapper` (its
  `<a>`) reached React with no key.

## Proposed Changes

- `plugins/main/public/components/common/tables/table.tsx`: the shared
  `TableBasicManageExpandedItems` wrapper (used by `TableWzAPI` /
  `TableWithSearchBar`, and therefore by Endpoints Summary, Endpoint Groups,
  and every other table built on this shared component) passed
  `itemIdToExpandedRowMap={isExpandable && itemIdToExpandedRowMap}` to
  `EuiBasicTable`. When a table doesn't opt into expandable rows (no `itemId` +
  `ExpandableRowContent`), `isExpandable` is `false`, and `false && ...`
  short-circuits to the boolean `false` instead of `undefined`. Changed to
  `isExpandable ? itemIdToExpandedRowMap : undefined` so the prop is either the
  real `{ [id]: ReactNode }` map or omitted — with no behavior change for
  tables that do use expandable rows (e.g. Security > Roles).
- `plugins/main/public/components/overview/vulnerabilities/common/components/vuls-evaluation-filter.tsx`:
  added the required `legend='Evaluated / Under evaluation'` prop to the
  `EuiButtonGroup`, matching the plain-string convention used by other
  `EuiButtonGroup` instances in `plugins/main`.
- `plugins/main/public/application.tsx`,
  `plugins/main/public/components/wz-header-nav-control/header-nav-control.tsx`:
  migrated the plugin's app-mount and header nav-control mount from the
  deprecated `ReactDOM.render`/`unmountComponentAtNode` to React 18's
  `react-dom/client` `createRoot`/`root.unmount()`, following the same
  pattern already used in `plugins/wazuh-ai-assistant`.
- `plugins/main/public/components/common/charts/charts/donut.tsx`: migrated
  the D3 tooltip's `ReactDOM.render` to `createRoot`, creating the root once
  per tooltip lifecycle (on `mouseenter`) and unmounting it on `mouseout`,
  instead of re-rendering into a plain DOM node on every `mousemove`.
- `plugins/main/public/components/common/tables/components/export-table-csv.tsx`,
  `plugins/main/public/components/common/tables/table-wz-api.tsx`: converted
  `defaultProps` on these function components to ES6 default parameters.
  `table-wz-api.tsx` also dropped a dead `searchBar: false` default that
  had no matching prop read anywhere in the file (the real toggle is
  `searchTable`), and widened the `title` prop type to `string | null` to
  match its actual resolved value.
- `plugins/main/public/components/common/dashboards/inventory.tsx`: the
  `getDashboardPanels.map()` callback's default branch (`className`
  omitted — true for every current caller: IT Hygiene, SCA, and FIM
  inventories, case management) returned the bare `<DashboardRenderer>`
  element with no `key`, unlike the `className`-truthy branch which already
  keyed its wrapping `<div>`. Added the same
  `key={dashboardId || agentDashboardId}` directly on `<DashboardRenderer>`
  so it's present regardless of which branch runs.
- `plugins/main/public/components/common/welcome/home-overview/components/common/stat-tile.tsx`:
  `EuiStat` wraps its title in a `<p>` by default (`titleElement`), but
  three `StatTile` callers (`finding-severity-tiles.tsx`, `sca-tiles.tsx`,
  `threat-hunting-section.tsx`) build their `value` with `RedirectAppLinks`,
  which always renders a `<div>` root with no way to change that. Added
  `titleElement='div'` to `StatTile`'s `EuiStat` call — matching the pattern
  already used in `vuls_severity_stat.tsx` — fixing all three call sites
  (and any future one) from this one shared component, with no visual
  change (`EuiTitle`'s styling comes from CSS classes, not the wrapping tag).
- `plugins/main/public/components/common/data-grid/use-data-grid.tsx`,
  `plugins/main/public/components/common/wazuh-discover/wz-discover.tsx`,
  `plugins/main/public/components/common/wazuh-discover/table.tsx`,
  `plugins/main/public/components/common/wazuh-data-grid/wz-data-grid.tsx`,
  `plugins/main/public/components/overview/vulnerabilities/dashboards/inventory/inventory.tsx`:
  added an exported `DATA_GRID_NON_EUI_PROP_KEYS` constant next to the
  `useDataGrid` hook — the single source of truth for which of its return
  keys aren't part of `EuiDataGridProps` — and applied
  `omit(dataGridProps, DATA_GRID_NON_EUI_PROP_KEYS)` consistently before
  every `<EuiDataGrid>` spread. `table.tsx` already filtered two of the
  three keys with a hardcoded list that had drifted out of sync (missing
  `dataGridStatePersistenceManager`); the other three call sites filtered
  nothing at all.
- `plugins/main/public/components/common/data-grid/use-data-grid.tsx`:
  `rowCount` fell back to `0` only when `results` itself was falsy, not when
  it lacked a `.hits` property. `wz-discover.tsx` initializes `results` as
  a truthy `{} as SearchResponse` placeholder before its first fetch
  resolves, so `rowCount` resolved to `Math.min(undefined,
  MAX_ENTRIES_PER_QUERY)` = `NaN`, which `EuiDataGridBody` used to compute
  its virtualized `Grid`'s height. Replaced the truthy check with
  `(results?.hits?.total as number) ?? 0`. Only `wz-discover.tsx` hits this
  path — it keeps `EuiDataGrid` mounted at all times (hiding it via a CSS
  class instead of unmounting), unlike its three siblings, which
  guard/conditionally render the grid until `results.hits.total` is a real
  number.
- `plugins/main/public/components/management/cluster/components/configuration_cards.tsx`:
  added a stable, kebab-case `key` (e.g. `'node-name'`, `'bind-address'`)
  to each of the 7 configuration items, derived from each item's label —
  matching the convention already used for `RibbonItemLabel` enum values
  elsewhere in this component family. The other 3 consumers of
  `WzRibbonBody`/`WzRibbon` already supplied a proper `key`, so no
  shared-component change was needed.
- `plugins/main/public/components/tools/devtools/application/components/top-nav/top-nav-menu.tsx`:
  moved the `key` to wrap whichever element the `.map()` callback actually
  returns (the `<a>` from `renderWrapper`, or the bare `EuiTab` fallback)
  via `React.cloneElement(..., { key: item.id })`, using the item's
  already-unique, stable `id` instead of array index. Removed the
  now-dead `key={idx}` on the inner `EuiTab`, which was never the
  top-level list element once wrapped.

### Artifacts Affected

None (frontend-only change, no build/config/package artifacts affected).

### Configuration Changes

None.

### Documentation Updates

None.

### Tests Introduced

- `plugins/main/public/components/common/tables/table-with-search-bar.test.tsx`:
  added a regression test asserting `EuiBasicTable` never receives a boolean
  `itemIdToExpandedRowMap` when expandable rows are not configured. Updated the
  existing snapshot (`itemIdToExpandedRowMap={false}` no longer appears — the
  only line that changed).
- `plugins/main/public/components/overview/vulnerabilities/common/components/vuls-evaluation-filter.test.tsx`
  (new file): asserts the button group renders with an accessible
  `legend`/group name.
- `plugins/main/public/application.test.tsx` (new file): asserts `renderApp`
  mounts with `createRoot` on the given element and that the returned
  cleanup calls `root.unmount()`.
- `plugins/main/public/components/wz-header-nav-control/header-nav-control.test.tsx`
  (new file): asserts the nav control registers in the right slot
  (`registerRight`/`registerLeftBottom`) per the home-page setting, mounts
  with `createRoot`, and unmounts on cleanup.
- `table-wz-api.test.tsx` and `export-table-csv.test.tsx` snapshots were
  regenerated — the only diffs are the removed dead `searchBar` prop and the
  top-level element no longer echoing resolved `defaultProps` values (now
  plain JS parameter defaults, not a public React prop); behavior is
  unchanged.
- `donut.tsx`'s D3-driven tooltip has no dedicated unit test (jsdom + d3
  mouse-event simulation is high-effort for low signal); covered instead by
  the existing `donut-card.test.tsx` (still passing) plus manual
  verification below.
- `stat-tile.test.tsx`: added a regression test asserting no
  `validateDOMNesting` console warning fires when `value` contains
  block-level content (simulating `RedirectAppLinks`'s `<div>`) — confirmed
  it fails without the fix and passes with it.
- `inventory.tsx` has no existing test file, and `InventoryDashboardTable`
  is wrapped in two data-source HOCs plus a data-fetching hook and several
  heavy child components with no existing test mocks anywhere in this
  codebase to reuse — building full isolated-render coverage from scratch
  was judged high setup cost for a one-line JSX fix. Verified manually
  instead (below).
- `use-data-grid.test.tsx`: added a regression test asserting
  `DATA_GRID_NON_EUI_PROP_KEYS` covers every hook-only key actually present
  on `useDataGrid()`'s return value, guarding against this exact class of
  drift (a filter list falling out of sync with the hook's return shape)
  recurring.
- `use-data-grid.test.tsx`: added a regression test reproducing the exact
  `{}`-placeholder scenario `wz-discover.tsx` starts with, asserting
  `rowCount` resolves to `0`, not `NaN` — confirmed it fails without the
  fix and passes with it.
- `configuration_cards.test.tsx` (new file, none existed before): renders
  `ConfigurationCards`, spies on `console.error`, and asserts no unique-key
  warning fires, plus that all 7 configuration fields render — confirmed it
  fails without the fix and passes with it.
- `top-nav-menu.test.tsx`: added a test case with an item defining
  `renderWrapper` (mirroring `apiReferenceItem`'s shape), asserting no
  unique-key warning fires — confirmed it fails without the fix and passes
  with it. Existing snapshots unchanged (this is a pure React-internals
  fix, no DOM/behavior change).

## Review Checklist

Manually verified in the dev dashboard (Docker `osd-dev`):
- Endpoints Summary (`/app/endpoints-summary`) — no `itemIdToExpandedRowMap`
  warning in console.
- Endpoint Groups (`/app/endpoint-groups`, Agents and Files tabs) — no
  `itemIdToExpandedRowMap` warning in console.
- Vulnerability Detection (`/app/vulnerability-detection`) — no `legend`
  warning for the Evaluated/Under evaluation toggle; toggle still filters
  correctly.
- Security > Roles (uses `itemId`/`ExpandableRowContent`) — expand/collapse
  still works per row, confirming the shared-component fix doesn't affect
  tables that do use expandable rows.
- Any `plugins/main` page — no `ReactDOM.render` deprecation warning on
  initial app mount; the header nav control still renders/unmounts
  correctly.
- A `ChartDonut` tooltip — still appears/updates/disappears correctly on
  hover with no console warning, including rapid mouseenter/mouseout
  cycling.
- A `TableWzAPI`-based table with `downloadCsv` and `title` set — still
  renders and exports correctly with no `defaultProps` warning.
- Any inventory-style module tab (e.g. IT Hygiene System > Hardware, SCA >
  Inventory) — no "Each child in a list should have a unique key prop"
  warning; the dashboard panel still renders/refreshes correctly.
- The Threat Hunting home-overview section and Endpoint Security section
  (where the affected `StatTile`s render) — no `validateDOMNesting` warning;
  severity-tile numbers/links still render, tooltip shows, and click still
  redirects via `RedirectAppLinks`.
- Discover, any `TableDataGridBasic`-based table, the standalone
  `WazuhDataGrid`, and the Vulnerability Detection inventory — no "React
  does not recognize the `setPagination` prop" (or any other unrecognized
  DOM attribute) warning; pagination, sorting, and column visibility still
  work correctly in all four.
- Discover — reload or perform a fresh search and watch the console during
  the brief window before results load; no `` `NaN` is an invalid value for
  the `height` css style property `` warning, and the grid still renders
  correctly once data arrives.
- Cluster management's node configuration overview (`ConfigurationCards`) —
  no unique-key warning; all 7 fields (Hidden, Name, Node name, Node type,
  Bind address, IP, Port) still display correctly.
- Dev Tools (Console) header, on an installation with `home:useNewHomePage`
  off — no unique-key warning; the Export button and the API Reference link
  (opening the docs in a new tab) still work correctly.

- [x] Code changes reviewed
- [ ] ...

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yenienserrano:
## Review — parallel pass (code review + issue coverage + live browser test)

Reviewed this PR in three independent tracks: a manual code review, an issue-coverage check against #9184, and a live browser test against a dev dashboard (osd-dev 3.6.0) running this PR's head commit with React in development mode, so prop-type and deprecation warnings are actually emitted.

The cleanup is solid: the two originally reported warnings are fixed at the shared component rather than per page, the `EuiDataGrid` hook-only prop leak is closed at all four call sites, `rowCount ?? 0` is strictly better than the previous ternary, and the affected Jest suites pass locally (13 suites / 73 tests / 5 snapshots). Note that CI never ran them — all four "Run unit tests" jobs failed on `yarn install` with a `502 Bad Gateway` from `registry.yarnpkg.com`, which is infrastructure, not code.

Three things should be addressed before merge. All three were reproduced live in the browser with the console captured in-frame; screenshots attached.

### 1. New warning introduced on every app exit — `application.tsx`

```
Warning: Attempted to synchronously unmount a root while React was already rendering.
React cannot finish unmounting the root until the current render has completed,
which may lead to a race condition.
```

`renderApp()` returns `() => root.unmount()`, and OSD core invokes that cleanup from inside React's commit phase. Captured JS call site:

```
ReactDOMRoot.unmount (osd-ui-shared-deps.js:184921)
  at wazuh.chunk.1.js:14845          <- plugins/main/public/application.tsx, `return () => root.unmount();`
  at Object.current (wazuh.plugin.js:9095)
  at unmount (core.entry.js:28058)
  at safelyCallDestroy (osd-ui-shared-deps.js:178521)
```

The old `ReactDOM.unmountComponentAtNode` did not emit this, so one deprecation warning has been traded for a different warning on every navigation out of a Wazuh app. Deferring the unmount resolves it:

```ts
return () => {
  queueMicrotask(() => root.unmount());
};
```

For reference, OSD's own `dataExplorer` plugin has the same pattern, and `header-nav-control.tsx` is byte-for-byte what OSD core does in `src/core/public/utils/mount.tsx` — so the header nav control change is defensible as-is, but the app mount is worth fixing since it fires on every app exit.

### 2. Same new warning from the donut tooltip cleanup — `donut.tsx`

The `useEffect` returns `onOut` as its cleanup, and `onOut` now calls `tooltipRootRef.current?.unmount()`, so the unmount runs inside `commitHookEffectListUnmount`. Reproduced by hovering a donut arc on Endpoints Summary and then clicking **Refresh**:

```
ReactDOMRoot.unmount (osd-ui-shared-deps.js:184921)
  at onOut (wazuh.chunk.1.js:17120)
  at safelyCallDestroy (osd-ui-shared-deps.js:178521)
  at commitHookEffectListUnmount (osd-ui-shared-deps.js:178689)
  at commitPassiveUnmountOnFiber (osd-ui-shared-deps.js:180623)
```

Component stack starts at `ChartDonut`. Note these two compound: leaving a Wazuh app while a donut tooltip is open fires both warnings on the same user action.

Two related points in the same file:

- `onEnter` overwrites `tooltipRootRef.current` without unmounting the previous root, and `onOut` only removes the *first* `body > div.wz-chart-tooltip`, so an enter without a matching out (or a second donut on the page) leaks a detached React root plus its tooltip div.
- Both would be resolved by creating one tooltip div and one root once in the effect, showing/hiding it on move/out, and unmounting only in the cleanup — which also removes the commit-phase unmount above.

Separately, `donut.tsx` is +58/−52 but only around 12 lines are the actual fix; the rest is Prettier reformatting of untouched code. Worth reverting the noise to keep an already 28-file PR reviewable.

### 3. The missing-`key` fix is incomplete — `dashboards/dashboard.tsx` was missed

Still reproducible on **Vulnerability Detection > Dashboard** and **Threat Hunting > Dashboard**:

```
Warning: Each child in a list should have a unique "key" prop.
Check the render method of `Dashboard`.
  at DashboardRenderer (wazuh.chunk.1.js:19001)
  at Dashboard (wazuh.chunk.1.js:19173)
```

`plugins/main/public/components/common/dashboards/dashboard.tsx` (~line 72) has the same `props.getDashboardPanels.map(...)` shape that was fixed in its sibling `inventory.tsx`, and its `<DashboardRenderer>` still has no `key` — only the `if (className)` wrapper branch below it has one. The PR added the key in `inventory.tsx` only. Same one-line fix applies:

```tsx
<DashboardRenderer
  key={dashboardId || agentDashboardId}
  dashboardId={dashboardId}
  ...
```

(React dedupes this warning per owner component, so it only shows once per page load — that is why it survived the original verification pass.)

### Smaller points

- `plugins/main/public/components/management/cluster/components/configuration_cards.test.tsx` — new file violates the repo's `filenames-simple/naming-convention` rule; eslint reports *"The filename must follow the rule: 'kebab-case'. Should rename to configuration-cards.test.tsx"*. The sibling source has the same violation, but it is pre-existing; a newly added file should not add one.
- `use-data-grid.test.tsx` — the drift guard asserts that every key in `DATA_GRID_NON_EUI_PROP_KEYS` exists on the hook's return, which is the harmless direction. It does not catch a *new* hook-only key being added to the return without being added to the list, which is exactly the drift that caused the original bug. Typing the constant as `readonly (keyof TDataGridReturn)[]` (after adding `setPagination` / `dataGridStatePersistenceManager` to `TDataGridReturn`) would make the compiler enforce it. A small exported `toEuiDataGridProps(props)` helper would also save the four call sites from each importing lodash `omit` plus the constant.
- `inventory.tsx:106` — `key={dashboardId || agentDashboardId}` is `undefined` when a panel defines neither, and duplicates if two panels share a `dashboardId`. Low risk, and consistent with the existing expression on the wrapping `div`, but the map index is a guaranteed-unique fallback.
- `table-wz-api.tsx` — `title` and `downloadCsv` are destructured out of `rest` and then re-injected into the spreads to `TableWithSearchBar` (:386) and `TableDefault` (:396), but neither component reads them. Also, `title = null` is what forces the `?? undefined` coalesce at :321; dropping the default (`title?: string`) removes both that and the `string | null` widening.
- The three new "no console warning" tests call `consoleError.mockRestore()` inline rather than in `afterEach`/`finally`. If `render` throws, `console.error` stays mocked for the rest of the file and later failures go silent.
- CHANGELOG has seven separate rows all linking #9184; every other entry in the table is one row per issue. There is multi-row precedent, so this is a preference, not a blocker.
- The "Results and Evidence" section of the PR description is still the unfilled placeholder comment and the "Relevant evidence provided" checkbox is unchecked. The screenshots below can fill it.

### Verified working (no regressions found)

Endpoints Summary and Endpoint Groups (Agents + Files tabs) — no `itemIdToExpandedRowMap` warning. Security > Roles — expandable rows still expand and collapse, so the shared-table change does not regress the tables that do use them. Vulnerability Detection toggle — `<legend class="euiScreenReaderOnly">Evaluated / Under evaluation</legend>` is present and filtering is real (Evaluated → 1,341 hits, Under evaluation → no results). Donut tooltips — 8 rapid mouseenter/mouseout cycles and 5 cross-chart rounds left zero leftover nodes. `TableWzAPI` title renders and CSV export returns `POST /api/csv` 200. Home-overview StatTiles — all 12 `.euiStat__title` render as `DIV` and a full-page scan found zero `<div>` inside `<p>`; the click redirect still works. Data grid — no `setPagination` attribute and no `NaN` height in the pre-results window, and pagination, sorting and column visibility all work; a DOM scan of `.euiDataGrid` found none of the three hook-only keys. Dev Tools header — no key warning, and the API Reference link still opens the docs. The bundle was verified to contain the PR code (`createRoot` ×3, zero `ReactDOM.render`/`unmountComponentAtNode` in `wazuh.plugin.js`).

Two things could not be tested in this environment: the cluster `ConfigurationCards` fix (the cluster app is not registered in 5.0.0 — `utils/applications.ts:775` reads `// cluster,`, and `/app/cluster` returns "Application Not Found"), and the `wazuh-discover` grid backed by alert rows (no alert data available; the identical code path was exercised through Vulnerability Detection Inventory, IT Hygiene and SCA instead).

One residual `ReactDOM.render` deprecation warning still fires once per page load, but it traces to `wazuh-security-dashboards-plugin`, not to this repo — out of scope here.

### Screenshots

**1. `application.tsx` — synchronous unmount on every app exit (Endpoints Summary → Discover)**

err-A-app-unmount

**2. `donut.tsx` — same warning from the tooltip cleanup (hover a donut arc, then Refresh)**

err-B-donut-unmount

**3. `dashboards/dashboard.tsx` — missing `key` on Vulnerability Detection > Dashboard**

err-C-dashboard-missing-key

**3b. Same warning reproduced on Threat Hunting > Dashboard**

err-C2-threathunting-missing-key

yenienserrano (commented):
https://github.com/wazuh/wazuh-dashboard-plugins/pull/9187#issuecomment-5731041401

rodrigofez (commented):
## Review

warning ´Warning: ReactDOM.render is no longer supported in React...´ no longer appears, but a new warning shows up 
Screenshot 2026-09-18 at 8 16 32 AM

Desvelao:
# Fixed

Implemented all suggestions of https://github.com/wazuh/wazuh-dashboard-plugins/pull/9187#issuecomment-5731041401 expect point 1 and 2.

1: no pattern found in Wazuh dashboard appliactions
2: unable to replicate

yenienserrano (approved):
LGTM :green_circle:

rodrigofez (approved):

